PR #3, opened 2018-06-29
RFC: Add TensorForest classifier and regressor to canned estimators

yupbank:
Since tree algorithm is one of the most popular algorithm used in kaggle competition
and we already have a contrib project [tensor_forest](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/tensor_forest) and people like them. If would be beneficial to move them inside of canned estimators.

cc: @nataliaponomareva

### DIFF
--- a/rfcs/20180626-tensor-forest.md
+++ b/rfcs/20180626-tensor-forest.md
@@ -107,11 +107,11 @@ classifier = estimator.TensorForestClassifier(feature_columns=[feature_1, featur
 						model_dir=None,
 						n_classes=2,
 						label_vocabulary=None,
+						head=None,
 						n_trees=100,
 						max_nodes=1000,
 						num_splits_to_consider=10,
 						split_after_samples=250,
-						base_random_seed=0,
 						config=None)
 
 
@@ -120,6 +120,12 @@ def input_fn_train():
   return dataset
 
 classifier.train(input_fn=input_fn_train)
+
+def input_fn_predict():
+  ...
+  return dataset
+
+classifier.predict(input_fn=input_fn_predict)
 
 def input_fn_eval():
   ...
@@ -133,11 +139,14 @@ Here are some explained details for the classifier parameters:
 - **feature_columns**: An iterable containing all the feature columns used by the model. All items in the set should be instances of classes derived from FeatureColumn.
 - **n_classes**: Defaults to 2. The number of classes in a classification problem.
 - **model_dir**: Directory to save model parameters, graph and etc. This can also be used to load checkpoints from the directory into an estimator to continue training a previously saved model.
-- **label_vocabulary**: A list of strings representing all possible label values. If provided, labels must be of string type and their values must be present in label_vocabulary list. If label_vocabulary is omitted, it is assumed that the labels are already encoded as integer or float  values within [0, 1] for n_classes=2, or encoded as integer values in {0, 1,..., n_classes-1} for n_classes>2 . If vocabulary is not provided and labels are of string, an error will be generated.
+- **label_vocabulary**: A list of strings representing all possible label values. If provided, labels must be of string type and their values must be present in label_vocabulary list. If label_vocabulary is omitted, it is assumed that the labels are already encoded as integer values within {0, 1} for n_classes=2, or encoded as integer values in {0, 1,..., n_classes-1} for n_classes>2 . If vocabulary is not provided and labels are of string, an error will be generated.
+- **head**: .A `head_lib._Head` instance, the loss would be calculated for metrics purpose and not being used for training. If not provided, one will be automatically created based on params
 - **n_trees**: The number of trees to create. Defaults to 100. There usually isn't any accuracy gain from using higher values (assuming deep enough trees are built).
 - **max_nodes**: Defaults to 10k. No tree is allowed to grow beyond max_nodes nodes, and training stops when all trees in the forest are this large.
 - **num_splits_to_consider**: Defaults to sqrt(num_features). In the extremely randomized tree training algorithm, only this many potential splits are evaluated for each tree node.
 - **split_after_samples**: Defaults to 250. In our online version of extremely randomized tree training, we pick a split for a node after it has accumulated this many training samples.
+- **config**: RunConfig object to configure the runtime settings.
+
 
 
 ### TensorForestRegressor
@@ -149,11 +158,11 @@ feature_2 = numeric_column('feature_2')
 regressor = estimator.TensorForestRegressor(feature_columns=[feature_1, feature_2],
 						model_dir=None,
 						label_dimension=1,
+						head=None,
 						n_trees=100,
 						max_nodes=1000,
 						num_splits_to_consider=10,
 						split_after_samples=250,
-						base_random_seed=0,
 						config=None)
 
 
@@ -162,6 +171,12 @@ def input_fn_train():
   return dataset
 
 regressor.train(input_fn=input_fn_train)
+
+def input_fn_predict():
+  ...
+  return dataset
+
+regressor.predict(input_fn=input_fn_predict)
 
 def input_fn_eval():
   ...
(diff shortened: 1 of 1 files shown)


ewilderj:
Adding the overview information. **This review will remain open for comment until the end of Monday, July 16th** (allowing for public holidays).

# TensorForest Estimator

| **Author(s)** | Peng Yu (yupbank@gmail.com) |
| **Sponsor**   | Natalia P (Google)                 |
| **Updated**   | 2018-06-26                                           |

## Objective

In this doc, we discuss the TensorForest Estimator API, which enables a user to create
[Extremely Randomized Forest](http://www.montefiore.ulg.ac.be/~ernst/uploads/news/id63/extremely-randomized-trees.pdf) Classifier and Regressor. And by inheriting from the `Estimator` class, all the corresponding interfaces will be supported.

rfcs/20180626-tensor-forest.md
```diff
@@ -0,0 +1,126 @@
...
+*   **model_dir:** Directory to save model parameters, graph and etc. This can also be used to load checkpoints from the directory into a estimator to continue training a previously saved model.
+*   **label_vocabulary:**  A list of strings represents possible label values. If given, labels must be string type and have any value in `label_vocabulary`. If it is not given, that means labels are already encoded as integer or float within [0, 1] for `n_classes=2` and encoded as integer values in {0, 1,..., n_classes-1} for `n_classes`>2 .  Also there will be errors if vocabulary is not provided and labels are string.
+*   **n_trees:** The number of trees to create. Defaults to 100. There usually isn't any accuracy gain from using higher values.
+*   **max_nodes:**  Defaults to 10,000. No tree is allowed to grow beyond max_nodes nodes, and training stops when all trees in the forest are this large.
+*   **num_splits_to_consider:** Defaults to `sqrt(num_features)` capped to be between 10 and 1000. In the extremely randomized tree training algorithm, only this many potential splits are evaluated for each tree node.
```
martinwicke: Are the 10 and 1000 boundaries universally accepted?

Nit, I would say "clipped", to my ear, "capped" only works for the upper bound.

rfcs/20180626-tensor-forest.md
```diff
@@ -0,0 +1,126 @@
...
+*   **split_after_samples:** Defaults to 250. In our online version of extremely randomized tree training, we pick a split for a node after it has accumulated this many training samples.
+*   **bagging_fraction:** If less than 1.0, then each tree sees only a different, random sampled (without replacement), bagging_fraction sized subset of the training data. Defaults to 1.0 (no bagging) because it fails to give any accuracy improvement our experiments so far.
```
martinwicke: If this gives no improvement, can we remove this argument?

We can always add stuff back, but we can never take it away (except at major versions) so we should be conservative in what we add.

yupbank:
For benchmark, yeah we might use this https://www.openml.org/search?q=ExtraTrees&type=flow 

> the efficiencies to be had for whole batch training? 
In this case not really, since the trees in tensor forest we are using are[ Hoeffding Tree](https://wiki.pentaho.com/display/DATAMINING/HoeffdingTree), which is a incremental tree, so we don't require full batch training

rfcs/20180626-tensor-forest.md
yupbank: for now i just borrowed from the original contrib implementation, but thanks for your suggestion, i guess i can use the benchmark tool to find out whether the original claim is valid

rfcs/20180626-tensor-forest.md
yupbank: for now i just borrowed from the original contrib implementation. it is not universal though. not sure why origin author implemented this

yupbank: not really, sklearn ExtraTree is using  `sqrt(num_features)` as default.
https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/tree/tree.py#L1192-L1202

rfcs/20180626-tensor-forest.md
yupbank: since the origin paper did have some numbers suggestion bootstrapping is not helping, i'll remove it form the api for now.

rfcs/20180626-tensor-forest.md
yupbank: i maybe i should remove this clip heuristic  ?

ewilderj:
@nataliaponomareva @martinwicke I think we're good to merge this now. Waiting for your LGTM and I'll merge.

martinwicke:
Can we reflect the discussion notes somewhere here? Could be as a link to a doc, even in the comment thread. I just don't want them lost. @tanzhenyu

ewilderj:
Agreed, before we've linked them at the bottom of the RFC. Either that or including them in this PR thread would also work, and we'll link the PR discussion at the bottom of the RFC.

rfcs/20180626-tensor-forest.md
```diff
@@ -0,0 +1,252 @@
...
+
+## Distributed version
+
+Since the trees are independent, for the distributed version, we would distribute the number of trees required to train evenly among all the available workers. For every tree, they would have two tf.resources available for training.
+
+## Differences from the latest contrib version
+
+- Simplified code with only limited subset of features (obviously, excluding all the experimental ones)
+- New estimator interface, support for new feature columns and losses
+
```
nataliaponomareva: can you also copy over this from the doc
We will try to reuse as much code from canned boosted trees as possible (proto, inference etc)

rfcs/20180626-tensor-forest.md
```diff
@@ -0,0 +1,252 @@
...
+|HiGGS| 11M| 28| 2| 70.9| 71.7|
+
+With single machine training, TensorForest finishes much faster on big dataset like HIGGS, takes about one percent of the time scikit-lean required.
+
+## Design Proposal
+
+### Interface
+### TensorForestClassifier
+
+```
```
ewilderj: use ```python to get the syntax highlighting?

rfcs/20180626-tensor-forest.md
```diff
@@ -0,0 +1,252 @@
...
+- **max_nodes**: Defaults to 10k. No tree is allowed to grow beyond max_nodes nodes, and training stops when all trees in the forest are this large.
+- **num_splits_to_consider**: Defaults to sqrt(num_features). In the extremely randomized tree training algorithm, only this many potential splits are evaluated for each tree node.
+- **split_after_samples**: Defaults to 250. In our online version of extremely randomized tree training, we pick a split for a node after it has accumulated this many training samples.
+- **config**: RunConfig object to configure the runtime settings.
+
+
+
+### TensorForestRegressor
+
+```
```
ewilderj: use ```python to get the syntax highlighting?

rfcs/20180626-tensor-forest.md
```diff
@@ -0,0 +1,255 @@
...
+
+At the start of training, the tree structure is initialized to a root node, and the leaf and growing statistics for it are both empty. Then, for each batch `{(x_i, y_i)}` of training data, the following steps are performed:
+
+1. Given the current tree structure, each instance `x_i` is used to find the leaf assignment `l_i` where this instance falls into.
+2. `Y_i` (the label of `x_i`) is used to update the leaf statistics of leaf `l_i`.
+3. If the growing statistics for the leaf `l_i` do not yet contain `num_splits_to_consider` splits, `x_i` is used to generate another split. Specifically, a random feature value is chosen, and `x_i`'s value at that feature is used for the split's threshold.
+4. Otherwise, `(x_i, y_i)` is used to update the statistics of every split in the growing statistics of leaf `l_i`. If leaf `l_i` has now seen `split_after_samples` data points since creating all of its potential splits, the split with the best score is chosen, and the tree structure is grown.
+
+
+## BenchMark
```
nataliaponomareva: Benchmark

rfcs/20180626-tensor-forest.md
```diff
@@ -0,0 +1,255 @@
...
+|Diabetes| 442| 10| Regression| 0.462| 0.461|
+|Boston| 506| 13| Regression| 0.793| 0.872|
+|Digits| 1797| 64| 10| 96.7| 97.6|
+|Sensit(Comb.)| 78k| 100| 3| 81.0| 83.1|
+|Aloi| 108K| 128| 1000| 89.8| 91.7|
+|rcv1| 518k| 47,236| 53| 78.7| 81.5|
+|Covertype| 581k| 54| 7| 83.0| 85.0|
+|HiGGS| 11M| 28| 2| 70.9| 71.7|
+
+With single machine training, TensorForest finishes much faster on big dataset like HIGGS, takes about one percent of the time scikit-lean required.
```
nataliaponomareva: hm where is time n this table? It is just performance metrics right?

yupbank: yeah, it's just performance metrics, i took it from the workshop paper.

nataliaponomareva: But what i am saying that this statement "much faster on big dataset" is not substantiated by this table. You either keep the table and say that it is from resource A, demonstrating that the quality is on par with scikit learn, and removing the statement that says that it trains faster. Or add a reference to the resource which states that it trains faster

yupbank: i see.. i'll add a source to the workshop paper, as the claim was also from the paper.

Karenou:
Hi, may I know whether the tensor forest package is still supported under tensorflow 2.0.x? Many thanks.

Issue #1, opened 2018-06-03
The Docker host machine only requires Nvidia Drivers


KinmanLam:
Currently, the Docker install GPU support instructions here:
https://www.tensorflow.org/install/install_linux#gpu_support 

Indicates it requires the full CUDA toolkit, cuDNN, and Nvidia drivers as outlined in:
https://www.tensorflow.org/install/install_linux#NVIDIARequirements

The document should be changed to indicate the host machine (running the container) requires only Nvidia drivers. See:
https://github.com/nvidia/nvidia-docker/wiki/CUDA#requirements

The container itself requires only the CUDA toolkit, cuDNN and Docker file can be found here:
https://hub.docker.com/r/nvidia/cuda/

ewilderj:
Please re-file this issue on tesnorflow/tensorflow. The community repo is for managing contributor processes.

PR #5, opened 2018-07-26
RFC: Generalizing tf.data batching using windowing and reducers

ewilderj:
**This RFC will be open for comment until Friday, August 10th, 2018.**

cc @jsimsa @mrry

# Generalizing tf.data batching using windowing and reducers 

| **Author(s)** | Jiri Simsa (Google)                                  |
| **Sponsor**   | Derek Murray (Google)                                |
| **Updated**   | 2018-07-26                                           |

## Objective 

This proposal addresses the known limitations of the current tf.data batching API:

*   it provides a mechanism for padded batching of sparse tensors
*   it facilitates customization of batching logic (users can now express batching logic as a pure Python function)
*   it enables application of different batching logic on different components

### DIFF
--- a/rfcs/20180726-tf-data-windowing-reducers.md
+++ b/rfcs/20180726-tf-data-windowing-reducers.md
@@ -1,0 +1,244 @@
+# Generalizing tf.data batching using windowing and reducers 
+
+| Status        | Proposed                                             |
+:---------------|:-----------------------------------------------------|
+| **Author(s)** | Jiri Simsa (Google)                                  |
+| **Sponsor**   | Derek Murray (Google)                                |
+| **Updated**   | 2018-07-26                                           |
+
+## Objective 
+
+This proposal addresses the known limitations of the current tf.data batching API:
+
+*   it provides a mechanism for padded batching of sparse tensors
+*   it facilitates customization of batching logic (users can now express batching logic as a pure Python function)
+*   it enables application of different batching logic on different components
+
+
+## Motivation
+
+The tf.data API is the de facto standard for creating TensorFlow input pipelines, whose purpose is to extract data from a storage system, transform it, and load it onto an accelerator.
+
+A common transformation performed by TensorFlow input pipelines is batching -- combining multiple tensors into a single tensor of higher dimension, most often to make a minibatch for training. Currently, the core tf.data API for batching consists of [batch](https://www.tensorflow.org/api_docs/python/tf/data/Dataset#batch) and [padded_batch](https://www.tensorflow.org/api_docs/python/tf/data/Dataset#padded_batch). The former assumes the inputs have the same shape and supports both dense and sparse inputs. The latter supports dynamically shaped inputs, such as you might find in sequential data: it assumes the inputs have the same rank but not necessarily the same shape and can pad differently shaped inputs to a common shape; only dense inputs are supported by padded_batch.
+
+The tf.data batching API has several limitations that has surfaced in various users requests:
+
+*   As already mentioned, the padded_batch transformation does not support sparse tensors inputs ([issue](https://github.com/tensorflow/tensorflow/issues/18302)).
+*   The current API is not flexible enough to accept user-provided batching logic (e.g. [issue](https://github.com/tensorflow/tensorflow/issues/20391)).
+*   The same batching logic needs to be applied to all components of the input dataset, which is not always desirable (e.g. [issue](https://github.com/tensorflow/tensorflow/issues/20391)). Users can work around this limitation by creating separate datasets to which different batching transformations are applied and them zipping the datasets; however, this can be inefficient, unergonomic, and error prone.
+
+
+## Proposal
+
+This document proposes leveraging the recently introduced support for _nested_ datasets as inputs to tf.data transformations to perform generalized batching as follows:
+
+1.  A <span style="text-decoration:underline;">window</span> transformation is used to combine consecutive elements of the input into a nested dataset (as opposed to a higher dimensional tensor).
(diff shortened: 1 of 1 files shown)


rfcs/20180726-tf-data-windowing-reducers.md
```diff
@@ -0,0 +1,244 @@
...
+
+*   Adding a `tf.data.Dataset.window` method, which provides the windowing functionality described in this proposal.
+*   Promoting the `tf.contrib.data.reduce_dataset()` method to `tf.data.Dataset.reduce()` and the `tf.contrib.data.Reducer` class to `tf.data.Reducer`.
+*   Allowing nested datasets as inputs of `map` and `filter`.
+*   Adding canned reducers for batching and padded batching of dense and sparse tensors to `tf.contrib.data`.
+
+
+## Open Questions
+
+*   Any interest in the window transformation supporting parameters for specifying the window shift and stride (similar to tf.contrib.data.sliding_window_batch)? Is there any other type of windowing that people are interested in?
```
bhack: What about the field of video analysis video inpainting? Have we some common shift a stride scenarios?

carlthome:
Super!

Having control over batching directly in the Python API would be great, and it would let me solve https://github.com/tensorflow/tensorflow/issues/20781 by making a `tf.data.Reducer` that sorts the tensor on the batch axis by some metadata like sequence id.

rfcs/20180726-tf-data-windowing-reducers.md
```diff
@@ -0,0 +1,244 @@
...
+#### Example 0: Count Dense Tensors 
+
+To introduce the concept of tf.data to readers unfamiliar with it, we illustrate how it can be used to count the elements of a dataset:
+
+
+```python
+def count(dataset):
+  """Counts the elements of a dataset."""
+
+  def init_fn(_):
```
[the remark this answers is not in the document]
jsimsa: It is the group key, which only makes sense if the reducer is used as part of the `tf.contrib.data.group_by_reducer`. We are still debating whether to have an unused `init_fn` argument in the non-keyed reducer or different interfaces for keyed and non-keyed reducers.

Mistobaan:
It feels like consuming the dataset in batches should be pushed downstream in the API. something along the line of `.get_next(batch_size=#)` The reduce API is something you would apply to the content of the dataset and not to its packaging/consumption.

carlthome:
Looking a little more closely at this I wonder if the return of `tf.data.Dataset.reduce()` shouldn't also be a `tf.data.Dataset`, because that seems like it's slowly becoming the expected object type in the high-level APIs such as tf.estimator:s and tf.keras.

E.g.
```py
estimator.train(lambda: dataset)

model.fit(dataset, steps)
```

It will probably be confusing if only some of the Dataset methods are for chaining. Thoughts?

ewilderj:
We're at the end of the official comment period, though this PR will remain open until after the design review committee has met. @jsimsa @mrry please provide replies to any outstanding questions and cover the Q&A from this PR as part of the design review meeting.

jsimsa:
@carlthome I would actually argue that because of the integration with `tf.estimator` or `tf.keras` you would want the result to be a tensor as opposed to a dataset.

As illustrated by the examples in the proposal, one common use of `tf.data.Dataset.reduce()` is to reduce windows of tensors of a dataset to a single tensor. If the reduced element would be a dataset instead, we would end up with a dataset of datasets which is something that `tf.estimator` or `tf.keras` would not necessarily know how to handle. If you wanted to pass a single tensor produce by `tf.data.Dataset.reduce()` to `tf.estimator` or `tf.keras`, you can always use `tf.data.Dataset.from_tensors` to wrap it in a dataset.

rfcs/20180726-tf-data-windowing-reducers.md
jsimsa: The stride/shift uses cases for video analysis I am aware of are covered by `tf.contrib.data.sliding_window_batch`.

rfcs/20180726-tf-data-windowing-reducers.md
```diff
@@ -0,0 +1,244 @@
...
+  """
+```
+
+
+### Reducers
+
+
+#### Example 0: Count Dense Tensors 
+
+To introduce the concept of tf.data to readers unfamiliar with it, we illustrate how it can be used to count the elements of a dataset:
```
ageron: Did you mean "To introduce the concept of tf.data.Reducer"?

georgesterpu:
Hi @jsimsa 
Is it possible to use the improved API to pad sequences differently than with constant values ?
The [tf.pad](https://www.tensorflow.org/api_docs/python/tf/pad) transformation supports the "reflect" and "symmetric" padding modes. Is a `PaddedBatchDataset` currently limited to the "constant" padding mode ?

jsimsa:
As Example #1 illustrates, the API allows you to write your custom padding logic, applying arbitrary `map` transformation (e.g. using `tf.pad`) to the elements to be padded.

In contrast, `padded_batch` is limited to either constant (or maximum) padding.

georgesterpu:
This looks really great, @jsimsa !
Would you please help me with an example of a `pad_fn` that when applied on a `WindowDataset` produces symmetrically right padded sequences up to the length of longest one ? There are a few differences between the examples in the RFC and the current codebase, and can't get the snippet below to work:

```python
import tensorflow as tf
import numpy as np

data = []
for _ in range(50):
    random_length = np.random.randint(low=10, high=100)
    data.append(tf.ones([random_length, 128]))

dataset = tf.data.Dataset.from_generator(lambda: data, output_types=tf.float32)


def pad_fn(sequences):
    return tf.pad(sequences, paddings=None, mode='SYMMETRIC')


dataset = dataset.window(4)
dataset = dataset.map(pad_fn)

for elem in dataset:
    print(elem)  # expected shape: [4, max_ts_batch, 128]
```

PR #8, opened 2018-07-31
RFC: TensorFlow Dockerfile Assembler

angersson:
**Review is now closed for comments**

# TensorFlow Dockerfile Assembler
 | Status        | Accepted       |
| **Author(s)** | Austin Anderson (angerson@google.com)
| **Sponsor**   | Gunhan Gulsoy (gunan@google.com)                     |
| **Updated**   | 2018-07-31                                           |

# Summary

This document describes a new way to manage TensorFlow's dockerfiles. Instead of handling complexity via an on-demand build script, Dockerfile maintainers manage re-usable chunks called partials which are assembled into documented, standard, committed-to-repo Dockerfiles that don't need extra scripts to build. It is also decoupled from the system that builds and uploads the Docker images, which can be safely handled by separate CI scripts.

**Important:** This document is slim. The real meat of the design has already
been implemented in [this PR to tensorflow/tensorflow](https://github.com/tensorflow/tensorflow/pull/21291).

### DIFF
--- a/rfcs/20180731-dockerfile-assembler.md
+++ b/rfcs/20180731-dockerfile-assembler.md
@@ -3,15 +3,23 @@
 | Status        | Proposed       |
 :-------------- |:---------------------------------------------------- |
 | **ID**        | <this will be allocated on approval>                 |
-| **Author(s)** | Austin Anderson (angerson@google.com)
+| **Author(s)** | Austin Anderson (angerson@google.com) |
 | **Sponsor**   | Gunhan Gulsoy (gunan@google.com)                     |
 | **Updated**   | 2018-07-31                                           |
 
 
 # Summary
 
-This PR describes a new way to manage TensorFlow's Dockerfiles that is detached
-from the build-and-delpoy-images system.
+This document describes a new way to manage TensorFlow's dockerfiles. Instead
+of handling complexity via an on-demand build script, Dockerfile maintainers
+manage re-usable chunks called partials which are assembled into documented,
+standard, committed-to-repo Dockerfiles that don't need extra scripts to build.
+It is also decoupled from the system that builds and uploads the Docker images,
+which can be safely handled by separate CI scripts.
+
+**Important:** This document is slim. The real meat of the design has already
+been implemented in [this PR to
+tensorflow/tensorflow](https://github.com/tensorflow/tensorflow/pull/21291).
 
 # Background
 
@@ -37,6 +45,10 @@ be finicky, and is not easily understood compared to vanilla Dockerfiles. Some
 Dockerfiles are duplicated, some are unused, and some users have made their own
 instead. None of the Dockerfiles use the ARG directive.
 
+Furthermore, `parameterized_docker_build.sh` is tightly coupled with the
+deploy-to-image-hub process we use, which is confusing because users who build
+the images locally don't need that information at all.
+
 This document proposes a new way for the TF team to maintain this complex set
 of similar Dockerfiles.
 
@@ -48,8 +60,9 @@ documented and support argument customization. Unlike the parameterized image
 builder script, this system excludes the image deployment steps, which should
 be handled by a totally different system anyway.
 
-This section lightly describes the design, which is fully implemented in this
-very PR.
+This section lightly describes the design, which is fully implemented in [this
+pull request to the main TensorFlow
+repo](https://github.com/tensorflow/tensorflow/pull/21291).
 
 Partial files are syntactically valid but incomplete files with Dockerfile
 syntax.
--- a/rfcs/20180731-dockerfile-assembler.md
+++ b/rfcs/20180731-dockerfile-assembler.md
@@ -1,0 +1,152 @@
+# TensorFlow Dockerfile Assembler
+
+| Status        | Proposed       |
+:-------------- |:---------------------------------------------------- |
+| **ID**        | <this will be allocated on approval>                 |
+| **Author(s)** | Austin Anderson (angerson@google.com)
+| **Sponsor**   | Gunhan Gulsoy (gunan@google.com)                     |
+| **Updated**   | 2018-07-31                                           |
+
+
+# Summary
+
+This PR describes a new way to manage TensorFlow's Dockerfiles that is detached
+from the build-and-delpoy-images system.
+
+# Background
+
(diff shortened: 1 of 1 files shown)


rfcs/20180731-dockerfile-assembler.md
```diff
@@ -0,0 +1,165 @@
+# TensorFlow Dockerfile Assembler
+
+| Status        | Proposed       |
+:-------------- |:---------------------------------------------------- |
+| **ID**        | <this will be allocated on approval>                 |
```
ewilderj: This ID line can be deleted, it's left in the template in error. The ID is actually the filename of the RFC. I've fixed the template now.

angersson: Fixed, thanks!

ewilderj:
cc @gunan

flx42:
It would be nice to mention:
- The current list of tags, with image sizes, and current limitations (e.g. the pip package making its way inside the image).
- The new list of tags, with image sizes

rfcs/20180731-dockerfile-assembler.md
```diff
@@ -0,0 +1,164 @@
...
+ARG _HELPER=${FROM_FOO:+foo}
+ARG BASE_IMAGE=${_HELPER:-bar}
+FROM ${BASE_IMAGE}
+…
+```
+
+...which means that you can dynamically set multiple FROM images. My first
+draft used ARGs and FROMs in a single Dockerfile to manipulate build stages.
+[The resulting
+Dockerfile](https://gist.github.com/angersson/3d2b5ae6a01de4064b1c3fe7a56e3821)
```
flx42: The Dockerfile in this gist is truncated.
Also, I'm not sure why ARGs and multi-stage builds need to work together. 

ARGs is for templating, allowing you to have a generic Dockerfile.
Multi-stage build is to avoid shipping with your build dependencies.

So I'm confused, for me there are totally different concepts.

Actually, when I took a quick look at the TF Dockerfiles, I thought multi-stage builds were not a good fit because:
- devel images need to maintain those build dependencies
- non-devel images already install from pip packages and don't have build dependencies (python deps excluded)
- many dependencies are installed through the package manager (`apt-get`) which doesn't work with multi-stage builds anyway

rfcs/20180731-dockerfile-assembler.md
```diff
@@ -0,0 +1,164 @@
...
+
+# Caveats and Rejected Alternatives
+
+I considered two alternatives while working on this.
+
+## Modern Multi-Stage Dockerfile
+
+"Multi-stage Building" is a powerful new Dockerfile feature that supports
+multiple FROM statements in one Dockerfile. It is meant to be used for creating
+artifacts with one image before using those artifacts in another image, but you
```
flx42: I don't understand why you say "but" here. Is the intended meaning "in addition"?

flx42:
Is it a requirement of this proposal that the end-user sees full Dockerfiles and can do a single `docker build`? Is it a common scenario? Do many users rebuild the docker images?

angersson:
@flx42 Thanks for mentioning the tags. This design is meant to be a foundational change that lets us better approach discussions about what-tags-should-we-have; the images I've started with mirror the existing Dockerfiles and aren't trying to make great strides yet.

I've updated the doc to better explain this. I've also revised some sections that you commented on. Thanks!

flx42:
Thanks! Why I'm asking about the "single Dockerfile requirement" in https://github.com/tensorflow/community/pull/8#commitcomment-29929864 is because you can achieve the same kind of composition that you have with multiple small Dockerfiles.


For instance, if you have a generic `FROM` (templated with an `ARG`), the Jupyter Dockerfile becomes:
```
ARG from
FROM ${from}

ARG PIP
RUN ${PIP} install jupyter
```

This is similar to your approach, you have a generic component that adds Jupyter to an existing image. But the composition of these Dockerfiles is expressed through `docker build` calls instead of a custom yaml:
```
docker build -t nvidia-devel -f Dockerfile.nvidia-devel .
docker build -t nvidia-devel-jupyter-py3 --build-arg from=nvidia-devel --build-arg pip=pip3 -f Dockerfile.jupyter .
```
At the CI level, this can be templated too, for the python version for instance.

This doesn't introduce any other format/scripting on top of Docker tools, but you don't have this single generated Dockerfile.

You can still achieve layer sharing by doing the builds in the right order, with a "tree" of Dockerfiles were only the leaves are pushed tags.
```
base
├── dev
│   ├── cpu-devel
│   └── nvidia-devel
└── nodev
    ├── cpu
    └── nvidia
```

rfcs/20180731-dockerfile-assembler.md
angersson: I think I may have been in the middle of updating the file when I decided to drop that approach. I fixed the line.

I've revised this entire section since your comment (which illustrated that this portion of the doc was poorly explained, thanks) to clarify why I included this explanation. Regular multi-stage builds wouldn't be very useful at all, but the support for multiple FROMs that they added would let us do very evil things that don't work very well in the long run. I included the gist to show how bad such a Dockerfile can get (it's not meant to be useful aside from that).

rfcs/20180731-dockerfile-assembler.md
angersson: It's a mixture of both -- I revised this whole section, so it should be better now. Thanks!

angersson:
Ah, right, now I understand. I didn't really think about using multiple files with multiple FROMs. I'll add a new section to the doc that describes that method and some of the tradeoffs for it.

One of the obvious downsides I can see with any process that offloads build complexity to the `docker build` call is that, by definition, it complicates the process of assembling all of those images. For a developer, that means reading a README, which is fine. For us, internally, it means we'd need to design a build script that pieces everything together for whatever images we want to build, or else have a large list of very similar build chains, which may grow into a helper script anyway. The design I have here front-loads that process and isolates the "what stuff is in this image" work, which I really like.

The two end results would be quite similar, only different in where complexity lives (and it also looks like the multi-stage builds you've described might build a bit faster because of definitive re-use of stages compared to the implied re-use here). I've also already written all of the logic for this design, so only maintenance costs factor in to the extra effort.

Thanks for suggesting this!

ewilderj:
Thanks everyone for your review and contribution!

PR #12, opened 2018-08-23
Post-hoc cond_v2 design doc

skye:


rfcs/20180507-cond-v2.md
```diff
@@ -0,0 +1,170 @@
+# **"Functional"** **cond design doc**
+
+| Status        | In-progress       |
```
ewilderj: This is an accepted design doc we're retrospectively publishing, so should the status be Approved?

seanpmorgan:
Adding post-hoc design docs into community is much appreciated.

rfcs/20180507-cond-v2.md
skye: Sure. I put in-progress since we're still working on the implementation, but it's true that the design itself is approved. I'll change it.

skye: Should be fixed now.

PR #13, opened 2018-08-23
RFC: Functional while_loop

saxenasaurabh:
**This RFC is open for comment until 2018-09-07**

cc @skye 

# Functional while_loop
| **Author(s)** | Saurabh Saxena (Google) |
| **Sponsor**   | Skye Wanderman-Milne (Google)                 |
| **Updated**   | 2018-08-23                                           |

## Objective

This proposal talks about an implementation of [while_loop](https://www.tensorflow.org/api_docs/python/tf/while_loop) which adds a single While op to the GraphDef as opposed to the current implementation that uses [lower level primitives](https://arxiv.org/abs/1805.01772). The goal is to simplify debugging and other analysis and to make it easier for compiler backends like XLA to [recognize](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/tf2xla/functionalize_while.cc) the while loop in the GraphDef. At runtime, a C++ optimization pass will lower this op to the primitive dataflow ops for feature parity with the current implementation similar to how we do for the [If op](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/lower_if_op.cc).


oxinabox:
@malmaud: 
It is what we needed for https://github.com/malmaud/TensorFlow.jl/pull/338

eaplatanios:
@ewilderj This is a general comment (not specific to this PR), but it would be nice to link to working implementations of proposed RFCs whenever such implementations exist.

PR #15, opened 2018-08-27
RFC: Unify RNN interface

qlzh727:
**Review period open until 2018-09-10**

| ------------- |:---------------------------------------------------- |
| **Author(s)** | Scott Zhu (scottzhu@google.com)                       |
| **Sponsor**   | Eugene Brevdo (ebrevdo@google.com), Patrick Nguyen (drpng@google.com)|
| **Updated**   | 2018-08-24                                           |

## Objective

Unify RNN (LSTM/GRU/other recurrent models) interfaces between TF RNN and Keras RNN at the release 
of TensorFlow 2.0.

### DIFF
--- a/rfcs/20180920-unify-rnn-interface.md
+++ b/rfcs/20180920-unify-rnn-interface.md
@@ -1,6 +1,6 @@
 # Unify RNN Inteface
 
-| Status        | Proposed                                             |
+| Status        | Accepted                                             |
 | ------------- |:---------------------------------------------------- |
 | **Author(s)** | @qlzh727 (scottzhu@google.com)                       |
 | **Sponsor**   | @ebrevdo (ebrevdo@google.com), @drpng (drpng@google.com)|
--- a/rfcs/20180920-unify-rnn-interface.md
+++ b/rfcs/20180920-unify-rnn-interface.md
@@ -143,33 +143,12 @@ parameter. By default it only returns the output for the last timestep, and user
 return additional output for all timesteps, as well as the final states for the last timestep.
 
 #### Changes to make
-1. Aggressive Approach (<b>NOT TAKEN, due to high usage of TF RNN layer API</b>).
-
-    Based on the difference above, it is hard to do an automatic regex replacement in user's code, 
-    maybe the migration strategy is to:
-    1. Provide clear example about how to convert from TF RNN to Keras RNN. 
-    1. Update the Keras RNN to cover all the use cases that are already covered by TF RNN, eg nested 
-       input/output/state.
-    1. Make TF RNN as deprecated, but still available in tf.v1_compatible, eventually delete it 
-       in 2.X.
-
-1. Less Aggressive Approach (Adapted)
-    
-    1. Keep both TF RNN and Keras API interface in 2.0.
-    1. Provide clear tutorial and documentation about both API and their differences, state our 
-       preference about the Keras one.
-    1. Could also writing a conversion/compatibility layer under tf.nn that forward the call to 
-       tf.keras.RNN, eg:
-       ```python
-       tf.nn.static_rnn(cell, input, ...):
-         keras_input = tf.transpose(tf.convert_to_tensor(input), perm=[1, 0, 2])
-         Keras_output, keras_state = tf.keras.layers.RNN(
-             cell, return_sequence=True, return_state=True, unroll=False)(keras_input)
-         # convert keras_output/state to tf.nn tensor format
-         return output, states
-       ```
-       This will get rid of the duplicated logic for maintenance purpose, but still keep TF RNN API 
-       interface around. This layer need some prototype and performance tests.
+1. Add support nested input/output/state for Keras.RNN layer, which is one of the feature gap between
+   Keras RNN and TF RNN.
+1. Add support for time-major input tensor (time, batch, feature).
+1. Provide clear example about how to convert from existing TF RNN to Keras RNN interface.
+1. Make TF RNN as deprecated, but still available in tf.v1_compatible, eventually delete it 
+   in 2.X.
        
 ### RNN Cell
 Following table lists out the current status for RNN cells between 2 APIs:
@@ -235,7 +214,7 @@ enable all the cells to be used by both layer interfaces.
 1. Deprecate the duplicated RNN cells in TF RNN and suggest user to use the Keras equivalent, 
 namely:
     1. nn.BasicRNNCell -> keras.SimpleRNNCell
-    1. (nn.BasicLSTMCell,  keras.LSTMCell) -> nn.LSTMCell
+    1. (nn.BasicLSTMCell, nn.LSTMCell) -> keras.LSTMCell
     1. nn.GRUCell -> keras.GRUCell
     1. nn.MultiRNNCell -> keras.StackedRNNCells
     1. (Optional) Remove the individual dropout param from Keras cells and preferred the DropWrapper 
@@ -250,7 +229,7 @@ The side effects of deleting cell is that previously saved checkpoint will not w
 
 There are extra implementations for various RNN cells and wrappers in tf.contrib.rnn. Since 
 tf.contrib is going away in TF 2.0. It is a good time to do the housekeeping and decide whether they
-should be moved to core, or leave them in contrib, or just delete.
+should be moved to core, or move to newly introduce SIG-Addon repository.
 
 |Name|Usage|Comment|
 |:---|:----|:------|
@@ -269,18 +248,18 @@ should be moved to core, or leave them in contrib, or just delete.
 |TimeFreqLSTMCell|low usage within Google|Leave in v1.compat due to low usage|
 |GridLSTMCell|low usage within Google|Leave in v1.compat due to low usage|
 |BidirectionalGridLSTMCell|low usage within Google|Leave in v1.compat due to low usage|
-|NASCell|low usage within Google|Leave in v1.compat due to low usage|
-|UGRNNCell|low usage within Google|Leave in v1.compat due to low usage|
-|IntersectionRNNCell|low usage within Google|Leave in v1.compat due to low usage|
+|NASCell|low usage within Google|Move to new Addon repository|
+|UGRNNCell|low usage within Google|Move to new Addon repository|
+|IntersectionRNNCell|low usage within Google|Move to new Addon repository|
 |PhasedLSTMCell|low usage within Google|Leave in v1.compat due to low usage|
 |ConvLSTMCell<br>Conv1DLSTMCell<br>Conv2DLSTMCell<br>Conv3DLSTMCell|low usage within Google|Keras has a Conv2DLSTM implementation and this is more generic. Probably port this core|
-|GLSTMCell|low usage within Google|Leave in v1.compat due to low usage|
-|SRUCell|low usage within Google|Leave in v1.compat due to low usage|
-|IndRNNCell|low usage within Google|Leave in v1.compat due to low usage|
-|IndyGRUCell|low usage within Google|Leave in v1.compat due to low usage|
-|IndyLSTMCell|low usage within Google|Leave in v1.compat due to low usage|
+|GLSTMCell|low usage within Google|Move to new Addon repository|
+|SRUCell|low usage within Google|Move to new Addon repository|
+|IndRNNCell|low usage within Google|Move to new Addon repository|
+|IndyGRUCell|low usage within Google|Move to new Addon repository|
+|IndyLSTMCell|low usage within Google|Move to new Addon repository|
 |AttentionCellWrapper|Medium usage|Move to core|
-|HighwayWrapper|low usage within Google|Leave in v1.compat due to low usage|
+|HighwayWrapper|low usage within Google|Move to new Addon repository|
 
 
 ### CuDNN implementation vs Normal implementation
--- a/rfcs/20180920-unify-rnn-interface.md
+++ b/rfcs/20180920-unify-rnn-interface.md
@@ -1,0 +1,377 @@
+# Unify RNN Inteface
+
+| Status        | Proposed                                             |
+| ------------- |:---------------------------------------------------- |
+| **Author(s)** | @qlzh727 (scottzhu@google.com)                       |
+| **Sponsor**   | @ebrevdo (ebrevdo@google.com), @drpng (drpng@google.com)|
+| **Updated**   | 2018-08-24                                           |
+
+## Objective
+
+Unify RNN (LSTM/GRU/other recurrent models) interfaces between TF RNN and Keras RNN at the release 
+of TensorFlow 2.0.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


rfcs/20180920-unify-rnn-interface.md
```diff
@@ -0,0 +1,377 @@
...
+    1. nn.BasicRNNCell -> keras.SimpleRNNCell
+    1. (nn.BasicLSTMCell,  keras.LSTMCell) -> nn.LSTMCell
+    1. nn.GRUCell -> keras.GRUCell
+    1. nn.MultiRNNCell -> keras.StackedRNNCells
+    1. (Optional) Remove the individual dropout param from Keras cells and preferred the DropWrapper 
+        from TF RNN.
+
+1. All the RNN cells need to have a unified namespace. 
+
+The side effects of deleting cell is that previously saved checkpoint will not work, and maybe TF
```
musically-ut: I don't follow, which `cell` is being removed?

Perhaps it should read `deleting the dropout parameter` instead of `deleting cell`?

qlzh727: As listed above, eg tf.nn.BasicRNNCell will be removed and tf.keras.SimpleRNNCell will be preferred cell that achieve the same behavior.

musically-ut: Thank you for the clarification. I had misunderstood the point; I thought that the API would be deprecated _when_ 2.0 is released.

jheymann85:
One bigger difference between the TF and Keras interface is that the former can handle arbitrarily nested outputs while the latter assumes a single output tensor. Scanning through the proposal, I couldn't find anything explicit about this.
Will the Keras interface support nested outputs in the future when it becomes the preferred one?

qlzh727:
@jheymann85, thanks for bring this up. The short answer is yes. The nested input/output is a feature that is heavily used within google as well. The keras interface will support that in the new API.

rfcs/20180920-unify-rnn-interface.md
```diff
@@ -0,0 +1,377 @@
...
+    1. tf.nn.rnn_cell.BasicLSTMCell,  tf.keras.LSTMCell -> tf.nn.rnn_cell.LSTMCell
+    1. tf.nn.rnn_cell.GRUCell -> tf.keras.GRUCell
+    1. tf.nn.rnn_cell.MultiRNNCell -> tf.keras.StackedRNNCells
+1. Move existing tf.nn.rnn* code from tensorflow/python/ops/ to tensorflow/python/rnn. The RNN layer
+and cell are not tensorflow ops, and is a bit weird to stay under ops. Externally the tf.export will
+ensure the object/function stays the same.
+1. Migrate selected RNN cell from contrib to core (tf/python/rnn). See detailed section below.
+1. Still keep both TF and Keras RNN layer API in TF 2.0, since both APIs are heavily used. Update 
+documentation to state that Keras is the preferred API to use, and encourage user to use Keras. Add 
+warning message in TF RNN layer, point user to documentation for suggested high level RNN layer API.
```
mgouicem: Could you clarify how the gates ordering will be impacted? Currently Keras and TF weights have different orders for both LSTM and GRU (e.g. for LSTM, TF has ICFO ordering when Keras has IFCO). Will those be unified or will they remain unchanged?

rfcs/20180920-unify-rnn-interface.md
```diff
@@ -0,0 +1,377 @@
...
+
+
+## Design Proposal
+
+1. Update RNN cells in Keras and TF RNN, so that all of them will work with Keras RNN layer and TF 
+RNN layer.
+1. Remove the duplicated RNN cells between Keras and TF, update the documentation for the migration 
+path.
+    1. tf.nn.rnn_cell.BasicRNNCell -> tf.keras.SimpleRNNCell
+    1. tf.nn.rnn_cell.BasicLSTMCell,  tf.keras.LSTMCell -> tf.nn.rnn_cell.LSTMCell
```
mgouicem: If keras becomes the prefered API, why the remaining LSTM cell lives in tf.nn?

rfcs/20180920-unify-rnn-interface.md
```diff
@@ -0,0 +1,377 @@
...
+swapped to use CuDNN version if it meets certain criteria:
+
+```python
+class LSTM(RNN):
+  def __init__(self, .....):
+     if activation == 'tan' and dropout == 0 and use_bias == True:
+       self.could_use_cudnn = True
+     else:
+       self.could_use_cudnn = False
+     ....
```
mgouicem: Will this impact the default behavior of the Cells? And in particular, the GRUCell (cuDNN support only 1406.1078v1 when Keras default is 1406.1078v3 and TF only supports 1406.1078v3) ?

rfcs/20180920-unify-rnn-interface.md
qlzh727: I think those will remain the same since they are just implementation details. The weights between Keras and TF cells has other differences than that, so we don't expect the weights to work between them.

rfcs/20180920-unify-rnn-interface.md
qlzh727: Thanks for pointing this out. The reason is that tf.nn.LSTMCell supports more features like state clipping and peephole. We might eventually update the tf.keras.LSTMCell to support those features, so that all the keras cells will be preferred.

rfcs/20180920-unify-rnn-interface.md
qlzh727: I don't think it will change the behavior. I am not so sure about the difference between 1406.1078v1 and v3, but I think one of them are the dropout defaults. I think you can also translate the code into:

if parameter_match_v1:
  self.could_use_cudnn = True
else
  self.could_use_cudnn = False

rfcs/20180920-unify-rnn-interface.md
mgouicem: I agree, these are implementation details. However, if you keep the gates ordering different in the two, this can imply a performance difference between the two (in particular when an acceleration library is used underneath, e.g. MKL-DNN or CUDNN).
From a usability point of view, it can be odd to have 2 interface layers with similar functionality but different performance.

rfcs/20180920-unify-rnn-interface.md
mgouicem: ok thanks for the clarification.

rfcs/20180920-unify-rnn-interface.md
mgouicem: ok thanks for the clarification

rfcs/20180920-unify-rnn-interface.md
qlzh727: Sorry for the late reply. 

I see what you are trying to address here. I think there is a issue like circular dependency here: the underlying acceleration library could like a stable input format from TF, and TF also would like to have a unified format from all the underlying library. 

Also I am not sure whether there will be a performance implication for this reason. The weights of the individual gate should be saved separately, we could pass them to underlying library with different order, it does not necessary mean that we have to break them and restitch back. Can u clarify the reason for the performance part a bit more?

mgouicem: Thanks for the reply.

The performance concern is exactly because of the weights reorders as you mention.
In the cases where the number of iterations is small (think seq2seq decoder), you either will need shadow copies of the weights that are reused from iteration to iteration, or you will have to reorder for each call to the library, which can have a large overhead (especially for inference).

Saving them separately would not solve the issue as all libraries I know of require to have the gate weights packed into one big matrix (it enables to run one big GEMM instead of many smaller ones). So you would still have some copy overhead.

From a library point of view, we are fine with whatever format you pick. However, if you have 2 formats in the same framework, it becomes harder to accommodate efficiently.

qlzh727: Ack, we will have discussion internally about the gate format. I think most of the LSTM cell in TF are in ICFO order, except the cudnn LSTM cell, which is IFCO since the underlying cudnn lib require the input to be in that format. I think unifying into one format will be helpful. Will post the discussion outcome here within this week.

mgouicem: Thanks. As mentioned earlier, it seems that [tf.keras uses IFCO](https://github.com/keras-team/keras/blob/f534cb0c459ce2a8384e3bc0375eb1cbf80edb56/keras/layers/recurrent.py#L1899) too.

ewilderj:
@qlzh727 please could you post the notes from the review meeting into this thread.

also, are there any updates to this document?

once both those are resolved I can merge this.

qlzh727:
@ewilderj, the review meeting is scheduled tomorrow, will update this PR after that.

karmel:
## Notes from review



*   What has changed? Nothing major.
*   Goals: reduce duplication/divergence between Keras + TF RNN; change behavior of LSTM/RNN/GRUs to use cudnn by default
*   A theoretical option: make NVIDIA change their implementation to match the standard on CPU
*   Q1 from ispir: will all the features from current TF RNN be supported-- tf.{static|dynamic|bidirectional}_rnn , which are the basic building blocks that Keras does not expose. 
    *   Useful low-level primitives (analogous to, say matmul)
    *   We don't need to recommend users use these, but nice to have available
    *   The Keras layer can take an arbitrary cell
    *   End goal is to support all the current features: nested inputs, and other small things
    *   If that is the case, then Keras should be able to handle the current TF use cases.
    *   apassos: I do not see how to control sequence_length in the keras API. I know masking is done in a weird way in keras which doesn't work in eager.
    *   **Keras has a mask, but we need some way to convert from seq length to mask. If that is provided, okay to remove TF rnn fns.**
*   Q2: should be bring in the research cells currently in contrib?
    *   Useful reference implementations. But maintenance overhead.
    *   Current change frequency is low.
    *   **For the ones that are research oriented, fine to exclude from core.**
    *   These can be included in the planned repo for subclasses of established TF APIs.
    *   https://tf-contrib-analyzer.herokuapp.com/ has stats. Many of these have few/no recorded OSS usages
*   Q3 from Intel: can we standardize the gate ordering (ICFO vs IFCO)?
    *   Implementation detail in one sense, but user wants it to be deterministic, and checkpoints will break.
    *   Unification is good.
    *   If unification, which one? Cudnn/Keras use IFCO, rest of TF uses ICFO.
    *   **Unify on IFCO, which maintains backwards compatibility with Keras and is Cudnn friendly.**
*   Prelude to defun questions: we have a functional interface that takes tensors in, outputs tensors. We can define these for CPU, GPU, etc, and then Grappler can check for these functions and swap in or out depending on the available hardware. So, for a dynamic RNN cell, you would define two fns, each of which is defunned and chosen at runtime. The API itself doesn't change, but the weights/checkpoints do.
*   Q4 from ebrevdo: how are we going to save and load weights?
    *   More prelude: in Cudnn, weights/variables are an opaque memory block, which improves performance. However, that variable requires Cudnn lib is available at time of graph creation. But, if you're building the graph on CPU for use in GPU, then the weights are initialized for CPU not GPU.
    *   Option 1: use CPU-compatible weights, and convert to Cudnn-friendly block with each batch. The c op handles this conversion under the hood.
    *   Option 2: make a Cudnn-friendly mem block without Cudnn.
    *   Option 3: we shouldn't need to initialize the unused variable on the CPU.
    *   **apassos and scottzhu to discuss offline.**
*   Q5: should be support time-major data?
    *   TF RNN is time-major, Keras is batch-major. Cudnn is time-major. Non-Cudnn is batch-major.
    *   Cells themselves don't have batches. Time-major makes it easier to slice the data to send to the cells; more efficient.
    *   **Add an option to be time_major in the API. Not included in the keras-team/Keras API.**

alanhdu:
FWIW, I'd be very sad to see the `tf.nn` set of functions get deprecated or even officially discouraged.

For our use case, re rely heavily on feeding in initial states to our RNNs as placeholders and getting their output state every time we call into the TF graph. `tf.nn.dyanmic_rnn` and `tf.nn.static_rnn` make this really trivial to manage, and allow us to do state-saving in Python, which let's us hook it into the rest of our system.

We've found that the Keras API, which makes the state sharing part of the `LSTM` object , to be much harder to use and manage. Intuitively to me, at least, `tf.nn.rnn_cell.RNNCell`s are like `tf.layers.Dense` or any other layer object -- stateless functions at  inference time -- and  `tf.nn.dynamic_rnn` is "just" a higher-order function like `map`. This is a nice mental model which makes Keras's stateful RNN layers seem really weird in comparison.

(To be fair though, the last time we tried to use Keras's RNN APIs was more than a year ago, so things could've gotten a lot better. This could also have totally been a documentation issue, rather than any real issue with the APIs).

qlzh727:
Hi @alanhdu. Thanks for the input. 

For initial_state of the Keras RNN, it actually accept initial_state as one of the parameter when constructing the layer. See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/layers/recurrent.py#L701.

You can feed in placeholder as initial state, and it should properly do the back prop. In the case the initial state is not provided, a zero filled tensor is used in that case.

Could u provide a short code snippet to explain the issue if the current Keras RNN API still cannot address your issue?

Thanks.

qlzh727:
@alanhdu 

keras API actually support same use case for returning output_state:

```
inputs = tf.placeholder(tf.float32, [None, None, 40])   # [batch_size, seq_length, dim]
input_state = tf.placeholder(tf.float32, [None, 50])     # [batch_siez, gru_size]
gru = keras.layers.GRU(50, **return_state=True**)

output, output_state = gru(inputs, initial_state=input_state)
```

See https://keras.io/layers/recurrent/#rnn and search for **return_sequences** and **return_state** for more details.

alanhdu:
@qizh727 Thanks for the pointer! I could've sworn that we tried to use those options unsuccessfully in the past, but obviously we were wrong and it works fine.

 I think I still stand-by my claim that `tf.nn.dynamic_rnn` is significantly more intuitive and easier to use, but if Keras supports our use case then I don't feel super strongly (although I'd love for there to be a "translation" guide somewhere documenting that).

bhack:
There is also a quite strong activity for Attention API design in Keras. Check https://github.com/keras-team/keras/issues/11172

qlzh727:
@alanhdu, yes, once we finish the API implementation, we will post documentation about how to move from deprecated API to use new one.

qlzh727:
@bhack, thanks for the info, I will have a check for that.

rfcs/20180920-unify-rnn-interface.md
```diff
@@ -0,0 +1,356 @@
+# Unify RNN Inteface
+
+| Status        | Proposed                                             |
```
goldiegadde: Scott, 
if all the comments are addressed can you please update the status to "Accepted"

qlzh727: Done.

Stercator:
Why do both RNN cells and layers exist, instead of having just RNN layers (containing the cell functionality)?

PR #16, opened 2018-08-27
RFC: TensorFlow API symbols and namespaces

annarev:
**Review open for comments until 2018-09-11**

# TensorFlow Namespaces

| Status        | Accepted       |
| **Author(s)** | Anna Revinskaya (annarev@google.com), Andrew Selle (aselle@google.com) |
| **Updated**   | 2018-08-27                                           |

## Objective

This document proposes organizing TensorFlow API symbols in corresponding logical subnamespaces. As TensorFlow library grows, it is important to structure namespaces in a clear way for easier discoverability and usability.

We define **endpoint** as full name that can be used to access a TensorFlow
symbol. For e.g. [name_scope](https://www.tensorflow.org/api_docs/python/tf/name_scope) can be accessed using either `tf.name_scope` or  `tf.keras.backend.name_scope`. Therefore, `name_scope` has 2 endpoints: `tf.name_scope` and  `tf.keras.backend.name_scope`.

At a high level we have the following goals:

*   Add a few additional namespaces.
*   Add additional endpoints for TensorFlow symbols in relevant namespaces.
*   Remove some of the existing endpoints.

### DIFF
--- a/rfcs/20180827-api-names.md
+++ b/rfcs/20180827-api-names.md
@@ -54,8 +54,6 @@ Furthermore, TensorFlow API has many users. Therefore, we should avoid removing 
 
 We plan to add the following additional namespaces:
 
-**tf.probability** - replaces `tf.distributions` namespace and will contain other ops related to statistics.
-
 **tf.random** - will contain random sampling ops.
 
 Note that we already introduced some new namespaces earlier in June, specifically
@@ -88,7 +86,7 @@ We also want to remove some of the existing endpoints. Specifically we were look
 *   Remove all endpoints that have been moved to `tf.random` namespace.
 *   Keeps most symbols in `tf.manip` in root namespace.
 
-In total, we propose to remove 222 endpoints, including 168 endpoints in the root namespace.
+In total, we propose to remove 207 endpoints, including 170 endpoints in the root namespace.
 
 See full list of endpoints we want to remove in *Appendix 2: Deprecated
 Endpoints*.
@@ -184,24 +182,6 @@ Work will be done in multiple stages:
 | tf.decode_csv | tf.io.decode_csv |
 | tf.depth_to_space | tf.nn.depth_to_space, tf.manip.depth_to_space |
 | tf.deserialize_many_sparse | tf.io.deserialize_many_sparse |
-| tf.distributions.Bernoulli | tf.probability.Bernoulli |
-| tf.distributions.Beta | tf.probability.Beta |
-| tf.distributions.Categorical | tf.probability.Categorical |
-| tf.distributions.Dirichlet | tf.probability.Dirichlet |
-| tf.distributions.DirichletMultinomial | tf.probability.DirichletMultinomial |
-| tf.distributions.Distribution | tf.probability.Distribution |
-| tf.distributions.Exponential | tf.probability.Exponential |
-| tf.distributions.FULLY_REPARAMETERIZED | tf.probability.FULLY_REPARAMETERIZED |
-| tf.distributions.Gamma | tf.probability.Gamma |
-| tf.distributions.Laplace | tf.probability.Laplace |
-| tf.distributions.Multinomial | tf.probability.Multinomial |
-| tf.distributions.NOT_REPARAMETERIZED | tf.probability.NOT_REPARAMETERIZED |
-| tf.distributions.Normal | tf.probability.Normal |
-| tf.distributions.RegisterKL | tf.probability.RegisterKL |
-| tf.distributions.ReparameterizationType | tf.probability.ReparameterizationType |
-| tf.distributions.StudentT | tf.probability.StudentT |
-| tf.distributions.Uniform | tf.probability.Uniform |
-| tf.distributions.kl_divergence | tf.probability.kl_divergence |
 | tf.divide | tf.math.divide |
 | tf.double | tf.dtypes.double |
 | tf.erf | tf.math.erf |
@@ -449,20 +429,6 @@ Work will be done in multiple stages:
 | tf.deserialize_many_sparse | replace with tf.io.deserialize_many_sparse |
 | tf.diag_part | replace with tf.linalg.tensor_diag_part |
 | tf.digamma | replace with tf.math.digamma |
-| tf.distributions.Bernoulli | replace with tf.probability.Bernoulli |
-| tf.distributions.Beta | replace with tf.probability.Beta |
-| tf.distributions.Categorical | replace with tf.probability.Categorical |
-| tf.distributions.Dirichlet | replace with tf.probability.Dirichlet |
-| tf.distributions.DirichletMultinomial | replace with tf.probability.DirichletMultinomial |
-| tf.distributions.Exponential | replace with tf.probability.Exponential |
-| tf.distributions.Gamma | replace with tf.probability.Gamma |
-| tf.distributions.Laplace | replace with tf.probability.Laplace |
-| tf.distributions.Multinomial | replace with tf.probability.Multinomial |
-| tf.distributions.RegisterKL | replace with tf.probability.RegisterKL |
-| tf.distributions.ReparameterizationType | replace with tf.probability.FULLY_REPARAMETERIZED |
-| tf.distributions.StudentT | replace with tf.probability.StudentT |
-| tf.distributions.Uniform | replace with tf.probability.Uniform |
-| tf.distributions.kl_divergence | replace with tf.probability.kl_divergence |
 | tf.encode_base64 | replace with tf.io.encode_base64 |
 | tf.erf | replace with tf.math.erf |
 | tf.erfc | replace with tf.math.erfc |
--- a/rfcs/20180827-api-names.md
+++ b/rfcs/20180827-api-names.md
@@ -227,8 +227,6 @@ Work will be done in multiple stages:
 | tf.lbeta | tf.math.lbeta |
 | tf.log_sigmoid | tf.math.log_sigmoid |
 | tf.logical_xor | tf.math.logical_xor |
-| tf.make_ndarray | tf.io.make_ndarray |
-| tf.make_tensor_proto | tf.io.make_tensor_proto |
 | tf.matmul | tf.linalg.matmul |
 | tf.mod | tf.math.mod |
 | tf.multinomial | tf.random.multinomial |
@@ -499,7 +497,6 @@ Work will be done in multiple stages:
 | tf.lgamma | replace with tf.math.lgamma |
 | tf.log_sigmoid | replace with tf.math.log_sigmoid |
 | tf.logical_xor | replace with tf.math.logical_xor |
-| tf.make_ndarray | replace with tf.io.make_ndarray |
 | tf.matching_files | replace with tf.io.matching_files |
 | tf.matrix_band_part | replace with tf.linalg.matrix_band_part |
 | tf.matrix_determinant | replace with tf.linalg.det |
--- a/rfcs/20180827-api-names.md
+++ b/rfcs/20180827-api-names.md
@@ -1,0 +1,640 @@
+# TensorFlow Namespaces
+
+| Status        | Proposed       |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | Anna Revinskaya (annarev@google.com), Andrew Selle (aselle@google.com) |
+| **Sponsor**   | Martin Wicke (wicke@google.com)                      |
+| **Updated**   | 2018-08-27                                           |
+
+## Objective
+
+This document proposes organizing TensorFlow API symbols in corresponding logical subnamespaces. As TensorFlow library grows, it is important to structure namespaces in a clear way for easier discoverability and usability.
+
+We define **endpoint** as full name that can be used to access a TensorFlow
+symbol. For e.g.
+[name_scope](https://www.tensorflow.org/api_docs/python/tf/name_scope) can be
+accessed using either `tf.name_scope` or  `tf.keras.backend.name_scope`.
+Therefore, `name_scope` has 2 endpoints: `tf.name_scope` and
+`tf.keras.backend.name_scope`.
+
+At a high level we have the following goals:
+
+*   Add a few additional namespaces.
+*   Add additional endpoints for TensorFlow symbols in relevant namespaces.
+*   Remove some of the existing endpoints.
+
+## Motivation
+
+TensorFlow currently has over 2000 endpoints total including over 500 endpoints in the root namespace. As number of symbols grows, it is important to maintain a clear structure to aid discoverability.
+
+Certain API symbol placements could be improved:
+
+*   Some namespaces were created recently and might not contain all the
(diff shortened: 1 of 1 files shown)


rfcs/20180827-api-names.md
```diff
@@ -0,0 +1,621 @@
...
+
+Furthermore, TensorFlow API has many users. Therefore, we should avoid removing endpoints if they are frequently used.
+
+## Design Proposal
+
+## Additional namespaces
+
+We plan to add the following additional namespaces:
+
+**tf.probability** - replaces `tf.distributions` namespace and will contain other ops related to statistics.
```
shoyer: Can you share any examples of "other ops related to statistics" that would fit in here? So far, the list below only contains functionality currently in `tf.distributions`.

I think "distributions" is slightly more descriptive than "probability" as a namespace for distribution objects. I'm also concerned about potential confusion with the TensorFlow probability project (https://github.com/tensorflow/probability), which currently uses the `tensorflow_probability` namespace.

annarev: Keeping these symbols in tf.distributions sounds good to me. I updated the doc accordingly.

I don't remember seeing any that are not in tf.distributions. This definition just included any other future ops we might add.

rfcs/20180827-api-names.md
```diff
@@ -0,0 +1,587 @@
...
+| tf.logical_xor | tf.math.logical_xor |
+| tf.matmul | tf.linalg.matmul |
+| tf.mod | tf.math.mod |
+| tf.multinomial | tf.random.multinomial |
+| tf.multiply | tf.math.multiply |
+| tf.negative | tf.math.negative |
+| tf.newaxis | tf.manip.newaxis |
+| tf.nn.in_top_k | tf.math.in_top_k |
+| tf.nn.l2_normalize | tf.math.l2_normalize, tf.linalg.l2_normalize |
+| tf.nn.log_poisson_loss | tf.losses.log_poisson_loss |
```
shoyer: `log_poisson_losss` feels like a slightly strange fit for `tf.losses`, given that it's a low-level function that doesn't adhere to the tf.losses interface (e.g., with the `loss_collection` and `reduction` arguments).

perfinion:
Is this RFC only about python? I'd like TF 2.0 to version the libtensorflow.so and _cc.so etc properly. (eg they need to be named libtensorflow.so.2.0) and the linker scripts should set versions on the symbols so in future ABI compat is easier. Should that be discussed somewhere else?

rfcs/20180827-api-names.md
```diff
@@ -0,0 +1,587 @@
...
+| tf.unsorted_segment_sqrt_n | tf.math.unsorted_segment_sqrt_n |
+| tf.unstack | tf.manip.unstack |
+| tf.variant | tf.dtypes.variant |
+| tf.verify_tensor_all_finite | tf.debugging.verify_tensor_all_finite |
+
+
+
+
+
+# Appendix 2: Deprecated Endpoints
```
martinwicke: Should we deprecate the tf.logging module? 

We clearly need some facility that TF can use internally, but I'm not sure that other people need to use the same? Maybe we keep a single endpoint with which users can get the logger, and then use that using their own import of python's logging?

Just a suggestion, triggered by https://github.com/tensorflow/tensorflow/issues/21930, which made clear to me that we're just providing a wrapper around logging without adding any real value.

martinwicke: Just talked about this with @tensorflow/api-owners and we do want to remove tf.logging from the public API (we will continue to use it for TF itself, but it's unclear what value, if any, it provides to users over just using Python's logging module.

rfcs/20180827-api-names.md
```diff
@@ -0,0 +1,587 @@
...
+| tf.saved_model.signature_constants.REGRESS_OUTPUTS | tf.saved_model.REGRESS_OUTPUTS |
+| tf.saved_model.signature_def_utils.build_signature_def | tf.saved_model.build_signature_def |
+| tf.saved_model.signature_def_utils.classification_signature_def | tf.saved_model.classification_signature_def |
+| tf.saved_model.signature_def_utils.is_valid_signature | tf.saved_model.is_valid_signature |
+| tf.saved_model.signature_def_utils.predict_signature_def | tf.saved_model.predict_signature_def |
+| tf.saved_model.signature_def_utils.regression_signature_def | tf.saved_model.regression_signature_def |
+| tf.saved_model.tag_constants.GPU | tf.saved_model.GPU |
+| tf.saved_model.tag_constants.SERVING | tf.saved_model.SERVING |
+| tf.saved_model.tag_constants.TPU | tf.saved_model.TPU |
+| tf.saved_model.tag_constants.TRAINING | tf.saved_model.TRAINING |
```
karmel: The lack of tag_constants here makes the meaning of these less clear. Is the primary goal reducing verbosity? Maybe .tags., or GPU_TAG even? Not sure.

karmel: Pasting @martinwicke 's response here to preserve the coherence of the thread: The primary reason is to reduce incidental complexity. This is particular
important for `signature_def_utils` and `signature_constants`, which are
very redundant. For tag_constants, how about making it a class, not a
module? tf.saved_model.Tags.GPU (and yeah, shorter, as you suggest as well)?

karmel: To be perfectly honest, I am one of the world-experts in these sets of constants at this point, and I still forget what the difference between tag_constants and signature_constants is, and which is used where. Let me think about these a little, and see if I have a better suggestion.

martinwicke:
Note also that I will not be upset if we remove either (signature_constants
seems the most likely) or even both. Not 100% sure what we would lose.

rfcs/20180827-api-names.md
annarev: I added a note about deprecating tf.logging under "Deprecated namespaces"

rfcs/20180827-api-names.md
```diff
@@ -0,0 +1,587 @@
...
+| tf.negative | tf.math.negative |
+| tf.newaxis | tf.manip.newaxis |
+| tf.nn.in_top_k | tf.math.in_top_k |
+| tf.nn.l2_normalize | tf.math.l2_normalize, tf.linalg.l2_normalize |
+| tf.nn.log_poisson_loss | tf.losses.log_poisson_loss |
+| tf.nn.log_softmax | tf.math.log_softmax |
+| tf.nn.log_uniform_candidate_sampler | tf.random.log_uniform_candidate_sampler |
+| tf.nn.sigmoid | tf.math.sigmoid |
+| tf.nn.softmax | tf.math.softmax |
+| tf.nn.top_k | tf.math.top_k |
```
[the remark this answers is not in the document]
annarev: I removed deprecation of tf.argmin from the doc since tf.argmax is not deprecated. Makes sense to keep both of them in root if one is in root.

However, tf.nn.top_k was never added to root namespace before. We want to avoid growing the root namespace (unless we are adding a very common op). The canonical endpoints for these symbols are:  tf.math.top_k, tf.math.argmin, tf.math.argmax. (i.e. all under same module - tf.math)

annarev:
I added some big changes to the doc, specifically:
- All endpoints under `tf.losses` and `tf.metrics` will be available under `tf.keras.losses` and `tf.keras.metrics` as well.
- `tf.layers` will be deprecated and replaced by `tf.keras.layers`.
- `tf.logging` will be deprecated. Python `logging` module can be used instead.

omoindrot:
@annarev : is there any particular reason why `tf.layers` is replaced by `tf.keras.layers` (and why we can't keep both) ?

Is the rationale to make users switch to `tf.keras` whenever they want to build a model?

annarev:
@omoindrot I updated the doc just now to keep layers under tf.layers which would be aliases for tf.keras.layers. Right now we have layers under tf.layers and tf.keras.layers. We want to keep one consistent set of layers. The plan is to add them to tf.keras.layers, but we can keep aliases in tf.layers as well.

However, signatures will likely change since we want all layers under tf.keras.layers to have consistent interface.

annarev:
I added 2 more new endpoints:
tf.bitcast - proposed additional endpoint tf.dtypes.bitcast to match tf.dtypes.cast.
tf.global_norm - proposed additional endpoint tf.linalg.global_norm. We also want to deprecate tf.global_norm endpoint, so that tf.linalg.global_norm is the only available endpoint.

Given these recent changes, I am extending review by 2 more days (until Friday).

yongtang:
Not sure if it is related to api and namespace, though I am wondering if we have a plan for data format types? In the existing APIs, there are several ways of representing the data format: `channels_first`/`channels_last`/`NCHW`/`NHWC`/`NCW`/`NWC`. Sometimes they adds complexity  to both implementation and to documentation. Would data format be covered as well?

martinwicke:
@perfinion let's talk about that on sig-build

@yongtang that's a good point. This RFC is mainly about the top-level symbols but we should talk about this. If 'channels-*' is sufficient I would prefer that, but it's not clear to me that it is. What we can definitely do is align these wherever we can. If HW vs WH is actually used, we should likely give up on the simpler symbols and just use that. @fchollet FYI.

rfcs/20180827-api-names.md
```diff
@@ -0,0 +1,607 @@
...
+
+## Deprecated endpoints
+
+We also want to remove some of the existing endpoints. Specifically we were looking for endpoints to remove based on the following criteria:
+
+*   Remove endpoints if they have preferred replacement and if these endpoints are not frequently used.
+*   Remove all endpoints that have been moved to `tf.quantization` namespace.
+*   Remove all endpoints that have been moved to `tf.random` namespace.
+*   Remove all endpoints from `tf.logging`.
+*   Keeps most symbols in `tf.manip` in root namespace.
```
martinwicke: Should we bother making manip? It's a mixture of set_ops and array_ops, and since we're keeping all of them (most of them?) in the root anyway, how about this:

tf.sets.* (only) for the set ops currently in manip
tf.* (only) for the tensor manipulation ops currently in manip

?

annarev: There is also manip_ops:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/manip_ops.cc
It currently only has 1 op (tf.manip.roll).
We also have tf.manip already:
https://www.tensorflow.org/api_docs/python/tf/manip

That being said, tf.manip.roll is not that frequently used and I think most references to other ops in tf.manip are still referring to these ops in root. So, we can probably deprecate tf.manip. It is just a bit strange since we just create it recently.

shoyer: +1 for not bothering with `tf.manipt`

annarev:
I updated the document to deprecate tf.manip namespace.
I also added deprecation for tf.sparse_to_dense. It is similar to tf.sparse_tensor_to_dense except the latter takes SparseTensor as input. So, calls to tf.sparse_to_dense can be replaced by creating SparseTensor and passing it to tf.sparse_tensor_to_dense.

ewilderj:
@annarev do you intend to have a design review meeting for this document?

annarev:
@ewilderj sorry for a late reply. We are not planning to have a design review meeting.

dsmilkov:
Hi @annarev, is this proposal finalized, and do you know when is the final date? Asking since there is pending PRs adding ops in TensorFlow.js 
(e.g. confusion_matrix) and we want to make sure we are aligned with the proposed namespaces here.

rfcs/20180827-api-names.md
```diff
@@ -0,0 +1,593 @@
...
+| tf.assert_type | tf.debugging.assert_type |
+| tf.bfloat16 | tf.dtypes.bfloat16 |
+| tf.bitcast | tf.dtypes.bitcast |
+| tf.bincount | tf.math.bincount |
+| tf.bool | tf.dtypes.bool |
+| tf.cast | tf.dtypes.cast |
+| tf.complex | tf.dtypes.complex |
+| tf.complex128 | tf.dtypes.complex128 |
+| tf.complex64 | tf.dtypes.complex64 |
+| tf.confusion_matrix | tf.train.confusion_matrix |
```
caisq: Why move `confusion_matrix` under `tf.train.*`? Confusion matrix doesn't have to do with training intrinsically. One can evaluate a model's confusion matrix even without training the model. The `tf.train.*` namespace seems to be mainly about optimizers and checkpoints, etc, which are very different things from confusion matrix.

I think confusion matrix belong more under the `tf.math.*` namespace. Also, `tf.metrics.*` may seem appropriate at first glance, but keep in mind that confusion matrix is not a single scalar, unlike all the things we have under `tf.metrics.* right now.

bileschi: +1

annarev: Thank you for suggestions! I already made changes in this doc, but we can definitely add more endpoints and deprecate tf.train.confusion_matrix in TensorFlow 2.0.

tf.metrics.* seems to be a good match. Let me confirm if adding tf.metrics.confusion_matrix would be ok.

annarev:
@dsmilkov This document has been finalized, but I can still make changes if a different name makes sense.

rfcs/20180827-api-names.md
```diff
@@ -0,0 +1,593 @@
...
+Furthermore, TensorFlow API has many users. Therefore, we should avoid removing endpoints if they are frequently used.
+
+## Design Proposal
+
+## Additional namespaces
+
+We plan to add the following additional namespaces:
+
+**tf.random** - will contain random sampling ops.
+**tf.keras.layers** - will contain all symbols that are currently under `tf.layers`. Note that signatures of these symbols will likely change to match layers under tf.keras.layers better.
```
asimshankar: @martinwicke : Should this be rephrased (as are other references to `tf.layers`, since we do plan to remove the functional layer calls and also the variable-scope related wrapping of `tf.keras.Layers`?)

martinwicke: Yes. I do think though that in addition to this rename RFC, we'll need another to publish the more detailed changes we're discussing.

rfcs/20180827-api-names.md
```diff
@@ -0,0 +1,593 @@
...
+| tf.saved_model.tag_constants.GPU | tf.saved_model.GPU |
+| tf.saved_model.tag_constants.SERVING | tf.saved_model.SERVING |
+| tf.saved_model.tag_constants.TPU | tf.saved_model.TPU |
+| tf.saved_model.tag_constants.TRAINING | tf.saved_model.TRAINING |
+| tf.saved_model.utils.build_tensor_info | tf.saved_model.build_tensor_info |
+| tf.saved_model.utils.get_tensor_from_tensor_info | tf.saved_model.get_tensor_from_tensor_info |
+| tf.scalar_mul | tf.math.scalar_mul |
+| tf.serialize_many_sparse | tf.io.serialize_many_sparse |
+| tf.serialize_sparse | tf.io.serialize_sparse |
+| tf.set_random_seed | tf.random.set_random_seed |
```
ageron: Shouldn't this be `tf.random.set_seed`?

PR #17, opened 2018-09-05
RFC: Deprecate Collections

rohan100jain:
The feedback phase will be open for two weeks until 2018-09-20

# Deprecate Collections
 
| **Author(s)** | Rohan Jain (Google) |
| **Sponsor**   | Asim Shankar (Google)                 |
| **Updated**   | 2018-09-05                                           |

## Summary

 TF 2.0 gives us a great opportunity to clean up APIs that are not as desirable. The collections API, although it permits a compact coding style and is widely used, poses a fair amount of technical difficulties and challenges and so we'd like to get rid of it.

1.  **Eager execution:** There isn't a concept of a graph when executing eagerly and therefore no support for collections.

2.  **Graph mode:** Collections are tied to a graph and assume that we build only one model per graph. As use cases get more and more complex, we have situations where we might build multiple models in a graph. Functions cause further issues because functions are graphs of their own. Any collections API usage inside a function would create state that is purely local to that function and the default graph would lose all that state.

### DIFF
--- a/rfcs/20180905-deprecate-collections.md
+++ b/rfcs/20180905-deprecate-collections.md
@@ -610,3 +610,19 @@ Tentatively, we have the following plan for deprecating collections
 
 
 1.  Figure out a concrete plan for removing CONCATENABLE_VARIABLES collection.
+
+# Design review notes
+* Going through dependencies of this effort.
+    * Object based checkpointing slated for 2.0. 
+    * V2 metrics are in 2.0 (so old metrics deprecated).
+    * Functional control flow slated for 2.0
+* Question: What is special about tables?
+    * Answer: Tables are immutable, they have to be initialized. We currently handle variables specially in defun, we could generalize that to tables, but haven’t looked into that yet. Tables are not variables today because they have non-Tensor representations (e.g., a std::hash_map)
+* Question: Can immutable tables be variables with a variant in them?
+    * Answer: This is a worthy direction to explore but we might not have enough time to do this.
+* Question: How do we create the “init_op” for serving? Do we need to track tables in tf.Checkpointable like we do variables? 
+    * Answer: Making tables be variables with variants does this automatically. Serialized format (for SavedModel) doesn’t need to change.
+* Question: How do we track all the tables?
+    * Answer: These tables are created by feature columns, which are used to create a Layer object, that Layer object can track all the tables/assets of interest. Make tables Checkpointable, and use the infrastructure for tracking checkpointables. Or a parallel mechanism? For V1, have a table_creator_scope() that can track all created tables
+* Other collections
+    * UPDATE_OPS: defun the layer and it is no longer a problem? This may be problematic for parameter server training if we have to wait for the updates before next iteration. Can be addressed by playing with strictness of function execution.
--- a/rfcs/20180905-deprecate-collections.md
+++ b/rfcs/20180905-deprecate-collections.md
@@ -1,9 +1,9 @@
 # Deprecating collections 
 
-| Status        | Proposed                                             |
+| Status        | Accepted                                             |
 :-------------- |:---------------------------------------------------- |
-| **Author(s)** | Rohan Jain (Google) |
-| **Sponsor**   | Asim Shankar (Google)                 |
+| **Author(s)** | Rohan Jain (Google)                                  |
+| **Sponsor**   | Asim Shankar (Google)                                |
 | **Updated**   | 2018-09-05                                           |
 
 
--- a/rfcs/20180905-deprecate-collections.md
+++ b/rfcs/20180905-deprecate-collections.md
@@ -444,7 +444,7 @@ class VariableTracker(object):
   def __init__(self):
     self.variables = []
 
-  def variable_tracker(next_creator, **kwargs):
+  def variable_tracker(self, next_creator, **kwargs):
     v = next_creator(**kwargs)
     self.variables.append(v)
     return v
--- a/rfcs/20180905-deprecate-collections.md
+++ b/rfcs/20180905-deprecate-collections.md
@@ -1,0 +1,575 @@
+# Deprecating collections 
+
+| Status        | Proposed                                             |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | Rohan Jain (Google) |
+| **Sponsor**   | Asim Shankar (Google)                 |
+| **Updated**   | 2018-09-05                                           |
+
+
+## Motivation
+
+TF 2.0 gives us a great opportunity to clean up API's that are not as desirable. The collections API, although it permits a compact coding style and is widely used, poses a fair amount of technical difficulties and challenges and so we'd like to get rid of it. The reasons for why we would want to do this are as follows
+
+
+
+1.  **Eager execution:** There isn't a concept of a graph when executing eagerly and therefore no support for collections.
+
+    Firstly, this has led to a fair amount of code been written as
+
+
+    ```python
+    if not context.executing_eagerly():
+      ops.add_to_collections(...)
+    ```
+
+    And developers need to be careful that in the eager execution code path no collections stuff is handled.
+
+
+    Furthermore, there are cases where collections have been used to keep graph based state (e.g. ensuring that only one variable for shared embedding columns is created). Such use cases need to be completely redesigned for eager.
+
+
+    With the assumption that eager execution would become more dominant in 2.0, this places undue burden on developers, makes code more error prone and results in divergence between writing code that executes eagerly and code that builds and executes graphs.
+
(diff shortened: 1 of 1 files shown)


rohan100jain:
Tried changing author

rfcs/20180905-deprecate-collections.md
```diff
@@ -0,0 +1,575 @@
...
+
+Initialize tables on first use. Similar to the [OneShotIterator](https://github.com/tensorflow/tensorflow/blob/r1.10/tensorflow/core/ops/dataset_ops.cc#L684), we'll pass in a function argument that can initialize the table. We can even explore the idea of the function to return a dataset and then initialize the table from that dataset. Having a function would avoid overheads of passing in large tensors that might be required to initialize on each lookup (index_table_from_tensor for instance might run into this problem). The Table API would remain the same since we hide initialization from the user and we wouldn't have to run an initializer explicitly. 
+
+The concern with this solution is that in serving, the first request would take more time than the rest of the calls and that sort of inconsistent performance is not desirable. 
+
+
+### **Asset file paths**
+
+Asset files currently only include vocabulary files that are used to initialize tables with. The tracking of tables should be sufficient to track vocabulary files. 
+
```
karmel: These also have to be replicated in the SavedModel, although passing these in as a python list would be way clearer than the strange thing we do now with collections, presumably. But it will require updating client code and SavedModel code as well.

rfcs/20180905-deprecate-collections.md
```diff
@@ -0,0 +1,575 @@
...
+### **Conditionals**
+
+The current Cond implementation uses collections but the new CondV2 implementation doesn't rely on it. Although moving to CondV2 is not part of TF 2.0 plans since its not a breaking change but it's quite likely that the timelines would line up. This effort (deprecating collections) would introduce a dependency on that effort.
+
+
+### **Updates**
+
+Batch normalization is the only use case for updates. For keras models, all updates are tracked via Model.updates. For the graph construction case, the updates are accessible via the updates attribute on the BatchNormalization layer itself and then they could be added to the train op. We will deprecate the [functional batch_normalization API](https://github.com/tensorflow/tensorflow/blob/r1.10/tensorflow/python/layers/normalization.py#L158) so that users deal with objects only.
+
```
karmel: Can you provide an example of what this would look like for user code?

rfcs/20180905-deprecate-collections.md
```diff
@@ -0,0 +1,575 @@
...
+## **Dependencies**
+
+This effort has dependencies on the following efforts
+
+
+
+1.  Queue runners removal and replacement with tf.data
+1.  [Variables 2.0](https://github.com/tensorflow/community/pull/11)
+1.  Object based checkpointing
```
karmel: My understanding was that this was not slated to be ready for 2.0? Has that changed?

rfcs/20180905-deprecate-collections.md
```diff
@@ -0,0 +1,575 @@
...
+
+1.  **Assets:** The ASSET_FILEPATHS collection tracks all external files ([lookup_ops](https://github.com/tensorflow/tensorflow/blob/r1.10/tensorflow/python/ops/lookup_ops.py#L550), [tf_transform](https://github.com/tensorflow/transform/blob/master/tensorflow_transform/analyzers.py#L830), [estimator](https://github.com/tensorflow/tensorflow/blob/r1.10/tensorflow/python/estimator/estimator.py#L941)) that are needed by the model for instance vocabulary files etc. These are used in a few places
+    1.  SavedModel: [builder_impl.py](https://github.com/tensorflow/tensorflow/blob/r1.10/tensorflow/python/saved_model/builder_impl.py#L354)
+    1.  TF Transform:  [analyzers.py](https://github.com/tensorflow/transform/blob/master/tensorflow_transform/analyzers.py#L98), [beam/impl.py](https://github.com/tensorflow/transform/blob/master/tensorflow_transform/beam/impl.py#L560), [impl_helper.py](https://github.com/tensorflow/transform/blob/master/tensorflow_transform/impl_helper.py#L405), [saved_io_transform.py](https://github.com/tensorflow/transform/blob/master/tensorflow_transform/saved/saved_transform_io.py#L213)
+
```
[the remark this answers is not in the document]
rohan100jain: Added.

rfcs/20180905-deprecate-collections.md
```diff
@@ -0,0 +1,575 @@
...
+
+For some specific use cases, we'll elaborate how we will handle them
+
+
+### **Global variables, trainable variables**
+
+For the keras models use case, Model.variables already tracks all variables created in a Keras model. Similarly, for trainable variables, Model.trainable_variables tracks all trainable variables. 
+
+In case of custom graph code, variable_creator_scope can be used to collect up variables.
```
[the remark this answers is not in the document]
rohan100jain: Added some example code. Regarding estimators: I think we can create one of these variable trackers before the model_fn is called, record the variables created and then feed them to the Scaffold so that the initialization happens correctly and the init_op etc. is set correctly.

rfcs/20180905-deprecate-collections.md
rohan100jain: Done.

rfcs/20180905-deprecate-collections.md
rohan100jain: My understanding was that its already live. @allenlavoie for more comments on this.

allenlavoie: Yep, `tf.train.Checkpoint` is in core. Estimators don't have an API for using it yet, and I am working on saving SavedModels using the same dependency system, but for training checkpoints it's all there.

rfcs/20180905-deprecate-collections.md
```diff
@@ -0,0 +1,612 @@
...
+*   [MovingAverageOptimizer.swapping_saver](https://github.com/tensorflow/tensorflow/blob/r1.10/tensorflow/contrib/opt/python/training/moving_average_optimizer.py#L140)
+*   [ExponentialMovingAverage.variables_to_restore](https://github.com/tensorflow/tensorflow/blob/r1.10/tensorflow/python/training/moving_averages.py#L473). We'll have to expose the variables created by ExponentialMovingAverage as they would have to be passed in here for this method.
+
+
+### **Tables**
+
+
+#### **Keras models**
+
+We'll provide a `track_table` method in the backend (similar to `track_variable`) and we'll make sure that when model.fit is called, tables are initialized similar to the way we initialize variables. Changes we'll have to make here are the following
```
rjpower: `track_variable` and `track_table` seem suspiciously similar to the global behavior we're trying to remove.  

Are these okay because they are special purpose?  Is there a rule of thumb for when global == OK?

rfcs/20180905-deprecate-collections.md
```diff
@@ -0,0 +1,612 @@
...
+
+
+We can add some helper methods on TableCreator to collect up all `initializers, vocabulary_filenames`.
+
+
+#### **Alternative proposal for tables**
+
+Initialize tables on first use. Similar to the [OneShotIterator](https://github.com/tensorflow/tensorflow/blob/r1.10/tensorflow/core/ops/dataset_ops.cc#L684), we'll pass in a function argument that can initialize the table. We can even explore the idea of the function to return a dataset and then initialize the table from that dataset. Having a function would avoid overheads of passing in large tensors that might be required to initialize on each lookup (index_table_from_tensor for instance might run into this problem). The Table API would remain the same since we hide initialization from the user and we wouldn't have to run an initializer explicitly. 
+
+The concern with this solution is that in serving, the first request would take more time than the rest of the calls and that sort of inconsistent performance is not desirable.
```
rjpower: Hidden table initializers have been a persistent thorn for us in some projects.  Having them be explicitly visible at the construction site would be a nice win for readability.

rfcs/20180905-deprecate-collections.md
rohan100jain: yeah they are indeed. The thinking is to provide them as a backstop because forcing everyone to migrate to managing their state immediately would be tough. As a principle, we will try not to use them in the tensorflow code base at all.

karmel: Tables in particular seem too special case. If we are just capturing initializers, maybe this can be generalized for that? Variables and initializers as special cases; everything else we keep out of global scope? We can handle assets through the EstimatorSpec (and the same for similar entities, perhaps?)

rohan100jain:
I added design review notes. Can someone take a look and approve this request so that I can merge it?

goldiegadde:
> I added design review notes. Can someone take a look and approve this request so that I can merge it?

Is there anything else actionable ? if not and everyone is in agreement with this proposal, Rohan, could you please update the status of the doc to "Accepted" and then we can merge it.

PR #18, opened 2018-09-07
RFC: Sunset tf.contrib

martinwicke:
**The feedback phase will be open for four weeks until 2018-10-01.**

# Sunsetting `tf.contrib`

 | Status        | *Proposed*                                           |
| **Author(s)** | Martin Wicke (wicke@tensorflow.org)                  |
| **Sponsor**   | Edd Wilder-James (ewj@tensorflow.org)                |
| **Updated**   | 2018-09-07                                           |
 
## Summary
 The `tf.contrib` module plays several important roles in the TensorFlow
ecosystem: It has made it easy for members of the community to contribute to
TensorFlow, and have their contributions tested and maintained. It is also used
as a staging ground to test early-stage and experimental features in TensorFlow.

 However, as the community has grown, the lack of scalability of the current
approach for maintaining and supporting `tf.contrib` has become apparent. 

 This RFC is a proposal to sunset the present `tf.contrib`, and replace its
important functions with more maintainable alternatives. Note that it also
affects some non-contrib code which is not part of the `tensorflow` module.

### DIFF
--- a/rfcs/20180907-contrib-sunset.md
+++ b/rfcs/20180907-contrib-sunset.md
@@ -198,7 +198,7 @@ formed) will be deleted by default at 2.0.
 | hvx                | satok16          | delete (redundant with NNAPI)        |
 | image              |                  | partial move to tensorflow/contrib?  |
 | input_pipeline     | rohan100jain     | delete                               |
-| integrate          | shoyer *mcoram   | delete?                              |
+| integrate          | shoyer *mcoram   | move to tensorflow/scientific?       |
 | kafka              | yongtang (mrry)  | move to tensorflow/io?               |
 | keras              | fchollet         | delete (moved to tf.keras)           |
 | kernel_methods    | petrosmol rostami | move to tensorflow/estimator?        |
--- a/rfcs/20180907-contrib-sunset.md
+++ b/rfcs/20180907-contrib-sunset.md
@@ -261,7 +261,7 @@ formed) will be deleted by default at 2.0.
 | testing            | ispirmustafa     | move to core, make private           |
 | text               |                  | move to tensorflow/contrib?          |
 | tfprof             |                  | delete (replaced by tf.profiler)     |
-| timeseries         | gsundeep karmel terrytangyuan | move to tensorflow/estimator         |
+| timeseries         | bananabowl karmel terrytangyuan | move to tensorflow/estimator         |
 | tpu                | saeta            | move to core                         |
 | training           | ebrevdo sguada joel-shor |                              |
 | util               |                  | delete (no owner), or move to tools  |
--- a/rfcs/20180907-contrib-sunset.md
+++ b/rfcs/20180907-contrib-sunset.md
@@ -261,7 +261,7 @@ formed) will be deleted by default at 2.0.
 | testing            | ispirmustafa     | move to core, make private           |
 | text               |                  | move to tensorflow/contrib?          |
 | tfprof             |                  | delete (replaced by tf.profiler)     |
-| timeseries         | gsundeep karmel  | move to tensorflow/estimator         |
+| timeseries         | gsundeep karmel terrytangyuan | move to tensorflow/estimator         |
 | tpu                | saeta            | move to core                         |
 | training           | ebrevdo sguada joel-shor |                              |
 | util               |                  | delete (no owner), or move to tools  |
--- a/rfcs/20180907-contrib-sunset.md
+++ b/rfcs/20180907-contrib-sunset.md
@@ -196,6 +196,8 @@ formed) will be deleted by default at 2.0.
 | hadoop             | yongtang (mrry)  | move to tensorflow/io?               |
 | hooks              | ispirmustafa     | delete (already in core)             |
 | hvx                | satok16          | delete (redundant with NNAPI)        |
+| igfs (#22194)      | dmitrievanthony  | move to tensorflow/io                |
+| ignite (#22210)    | dmitrievanthony  | move to tensorflow/io                |
 | image              |                  | partial move to tensorflow/contrib?  |
 | input_pipeline     | rohan100jain     | delete                               |
 | integrate          | shoyer *mcoram   | move to tensorflow/scientific?       |
--- a/rfcs/20180907-contrib-sunset.md
+++ b/rfcs/20180907-contrib-sunset.md
@@ -1,0 +1,280 @@
+# Sunsetting `tf.contrib`
+
+![Tranquil Sunset](https://www.publicdomainpictures.net/pictures/30000/velka/sunset-15.jpg)
+
+| Status        | *Proposed*                                           |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | Martin Wicke (wicke@tensorflow.org)                  |
+| **Sponsor**   | Edd Wilder-James (ewj@tensorflow.org)                |
+| **Updated**   | 2018-08-20                                           |
+
+## Objective
+
+The `tf.contrib` module plays several important roles in the TensorFlow
+ecosystem: It has made it easy for members of the community to contribute to
+TensorFlow, and have their contributions tested and maintained. It is also used
+as a staging ground to test early-stage and experimental features in TensorFlow.
+
+However, as the community has grown, the lack of scalability of the current
+approach for maintaining and supporting `tf.contrib` has become apparent. 
+
+This RFC is a proposal to sunset the present `tf.contrib`, and replace its
+important functions with more maintainable alternatives. Note that it also
+affects some non-contrib code which is not part of the `tensorflow` module.
+
+## Motivation
+
+`tf.contrib` serves the following functions:
+
+* Members of the community can submit code which is then distributed with the
+standard TensorFlow package. Their code is reviewed by the TensorFlow team and
+tested as part of TensorFlow's tests.
+* Additions to TensorFlow can be tested without falling under the
+[API stability guarantees](https://www.tensorflow.org/guide/version_compat)
+imposed by semver. New features in TensorFlow is often staged in `tf.contrib`
+before being "moved to core", where it falls under API stability guarantees.
+
+`tf.contrib` is organized in "projects". Each project's code is located in a
+subdirectory of [`tensorflow/contrib`](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib).
+There are currently 107 of such projects in `tf.contrib`. Each project is
+nominally maintained and supported by its owners. However, adherence to and
+enforcement of this rule has been weak, and most projects are effectively
+maintained by the TensorFlow team. 
+
+This leads to some issues with `tf.contrib`:
+
+* The barrier to contribution is too high. Because all code contributed to
+`tf.contrib` becomes part of the TensorFlow distribution, and because the
+TensorFlow team effectively assumes maintenance responsibilities for it, it has
+to be thoroughly reviewed. Even reviewing code for addition to contrib can take
+a long time.
+* `tf.contrib` is not actively maintained and the quality of its content is
+highly variable. Many of the projects in `tf.contrib` are long deprecated,
+though rarely formally. Many more have been abandoned by their original owners,
+or have been left unsupported and unmaintained for long periods of time.
+* Even where projects in `tf.contrib` are vigilantly maintained and
+enthusiastically supported, issues with and PRs for these projects are all mixed
+together in the rather unwieldy TensorFlow GitHub issue and PR trackers. No
+matter how well supported, the TensorFlow team still carries the burden of issue
+triage and PR management for all projects in `tf.contrib`.
+
+
+## Design Proposal
+
+This RFC argues that the `tf.contrib` has outlived its utility and should be
+retired. Each project in contrib has one of three options for its future: move
+to core; move to a separate repository; or delete. This decision will be made
+with the project owners, a proposal for all projects is below, most of which
+have been reviewed with the respective project owners.
+
+### Moving to core
+
+Some projects in `tf.contrib` have incubated successfully, and their content
+should be moved to TensorFlow. Examples are `tf.contrib.lite`,
+`tf.contrib.eager` or `tf.contrib.tpu`. Such a move will require an RFC, and is
+reserved for projects which are mature, useful for a large fraction of users,
+and which can be maintained by the TensorFlow team.
+
+For parts of these projects for which the exact API remains unclear, symbols
+should be prefixed with `experimental`. We have created an exception in our 
+[API stability guarantees](https://www.tensorflow.org/guide/version_compat) 
+for all symbols (functions, classes, methods, modules, or arguments) which
+begin with `experimental`. We reserve the right to change or remove these at
+will, or change their behavior.
+
+### Moving to a separate repository
+
+To provide an alternative avenue for contributed code, one or several new
+repositories will be created which are maintained by the community. Some larger
+projects, especially those which are likely to be interesting to a distinct
+community, are best moved into their own repository, and maintained by a SIG
+([which needs to be formed](https://github.com/tensorflow/community/blob/master/governance/SIGS.md)).
+A successful example of this is `tf.contrib.probability`, which has moved to
+`tensorflow/probability`.
+
+At least one such SIG should be formed which will maintain a "generic"
+TensorFlow contrib repository. This can contain community-contributed code
+which doesn't belong to a larger project, but which is useful for the TensorFlow
+community. Examples of such contributions include specialized and possibly
+rarely used ops or layers. The scope of such a repository should be kept
+reasonably focused (e.g., only ops, Layers, Metrics, Optimizers, Initializers)
+as to allow its maintainers to actually maintain the code.
+
+While the exact operating modalities for this repository will be up to the SIG,
+we propose to create the repository `tensorflow/contrib`. The TensorFlow team
+will provide test infrastructure similar to what is running for TensorFlow
+(though potentially with reduced platform coverage, TBD). The code in
+`tensorflow/contrib` should be written against an installed version of
+TensorFlow (i.e., using `import tensorflow as tf`, not the direct file imports
+used currently), and using only public APIs.
+
+Importantly, the content of this repository will not be included in the
+TensorFlow pip package, but is instead independently distributed. If the SIG
+chooses, it can release a separate pip package. The TensorFlow team can assist
+in setting up a build and release workflow.
+
+Even with such a repository available, it is important that TensorFlow related
+code created by the community is easy to find and use. To this end, we will
+enhance the community repository to contain an index of community projects. We
+invite all who have created tensorflow-related code to add their project there.
+To simplify the process of publishing your code in a usable form, we will also
+publish a template repository which shows how to set up a TensorFlow-related
+repository including a proper build against TensorFlow, tests, and distribution
+via a pip package.
+
+### Deleting
+
+Projects that cannot be maintained elsewhere, or which nobody is willing to
+maintain, will be removed from distribution with the release of TensorFlow 2.0.
+The TensorFlow 2.0 pip package will not include `tf.contrib`. The code for
+`tf.contrib` will be removed from the master branch once 2.0.0-RC0 is released,
+but will of course remain available on 1.x release branches.
+
+Note that *a lot* of projects are currently marked for deletion. This is not a
+reflection of the viability or quality of these projects, but rather due to the
+fact that no owner willing to maintain this project separately has been
+identified *yet*. If you see a project that you would like to maintain going
+forward, feel free to either initiate the process of forming a SIG to maintain
+it in the TensorFlow community umbrella, or simply fork the code. TensorFlow
+is free software.
+
+## List of Projects
+
+This section lists the possible fate of all projects currently in `tf.contrib`.
+The decisions here are preliminary, not all owners have been contacted (those
+marked with a * have not). Named in parentheses are people with a potential
+interest in this, but who are not owners.
+
+If you have an interest in one of these projects and are not named as an owner
+(or as an interested party), please speak up. I've added some (tentative) move
+targets to as of yet nonexistent repositories: `tensorflow/io`,
+`tensorflow/contrib` (I would prefer a better name for this), and
+`tensorflow/scientific` will all require new SIGs to maintain them.
+
+Note that any project that has no other fate specified (including those moving
+to a repo tentatively maintained by a SIG, but for which no SIG is actually
+formed) will be deleted by default at 2.0.
+
+| Project                 | Owner(s)    | Fate                                 |
+|:----------------------- |:----------- |:------------------------------------ |
+| tools/graph_transforms  | petewarden  | delete (replaced by grappler)        |
+| tools/quantization | petewarden       | delete (part of tflite)              |
+| all_reduce         | poxvoculi   | delete (replaced by DistributionStrategy) |
+| android            | petewarden       | delete (redundant: tflite/ARM build) |
+| autograph          | alexbw           | move to core                         |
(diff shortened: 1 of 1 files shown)


seanpmorgan:
The removal of tech debt is palpable :smile: 
As far as new names for `tensorflow/contrib`, some ideas:
- `tensorflow/extensions`
- `tensorflow/supplemental`
- `tensorflow/complementary`

rfcs/20180907-contrib-sunset.md
```diff
@@ -0,0 +1,280 @@
...
+| gan                | joel-shor        | move to separate repo                |
+| gdr            |*byronyi (poxvoculi)  |                                      |
+| graph_editor       | purpledog        | delete                               |
+| grid_rnn           |(qlzh727 ebrevdo) | delete                               |
+| hadoop             | yongtang (mrry)  | move to tensorflow/io?               |
+| hooks              | ispirmustafa     | delete (already in core)             |
+| hvx                | satok16          | delete (redundant with NNAPI)        |
+| image              |                  | partial move to tensorflow/contrib?  |
+| input_pipeline     | rohan100jain     | delete                               |
+| integrate          | shoyer *mcoram   | delete?                              |
```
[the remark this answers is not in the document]
martinwicke: Oops, my bad. Yes.

martinwicke:
Well, given that contrib is going away, I'm not sure we want to do this for contrib. But we are thinking about something like this for different projects in the tensorflow org, @ewilderj FYI.

bhack:
Yeah I meant for contrib projects moving  in  the separate repository (so the ones not moving in the core or deleted).

martinwicke:
If all goes according to plan, there may be 3 repositories afterwards:
tensorflow/contrib (I also like Sean's suggestions), tensorflow/scientific,
and tensorflow/io. I don't really want to distinguish projects inside
those, but in the end, that should be up to the SIG which maintains the
repo in question.

rfcs/20180907-contrib-sunset.md
```diff
@@ -0,0 +1,280 @@
...
+| tensor_forest      | nataliaponomareva yupbank | delete (moving to core)     |
+| tensorboard        |  nickfelt  jart  | move to tensorflow/tensorboard       |
+| tensorrt           | tobyboyd *samikama *aaroey *jjsjann123| move essential parts to core |
+| testing            | ispirmustafa     | move to core, make private           |
+| text               |                  | move to tensorflow/contrib?          |
+| tfprof             |                  | delete (replaced by tf.profiler)     |
+| timeseries         | gsundeep karmel  | move to tensorflow/estimator         |
+| tpu                | saeta            | move to core                         |
+| training           | ebrevdo sguada joel-shor |                              |
+| util               |                  | delete (no owner), or move to tools  |
```
leandro-gracia-gil: There's one function in this project that proved to be particularly useful for static optimizations on libraries built on top on TensorFlow, and as far as I know there is no alternative: tf.contrib.util.constant_value.

Considering the future of this project is undecided yet, would it be possible to ensure that at least this particular function or an alternative to it are available in TensorFlow 2.0?

AndreasMadsen:
Hi, author of `sparsemax` here. While I understand some contrib moduels are hard to maintain, `sparsemax` shouldn't be as it is implemented in python. There was a C++/CUDA implementation before, but I was asked to replace that for the sake of maintainability. I would therefore hope `sparsemax` will make it into `tensorflow/tensorflow`.

It was actually the TensorFlow team who originally asked me to implement it. It does also appear to be used quite a bit, see for example the past interrest in making it nan & infinity safe https://github.com/tensorflow/tensorflow/issues/15564.

perfinion:
Very much support this! In the unbundling dependencies work I've been doing, there are quite a few deps that are only used by some random tf.contrib module but nowhere else in the entire build so it's been a lot of work to deal with for not much gain.

A tensorflow-contrib package could even be published to pypi eventually with whatever is left after deleting so for the people that do want it pip install tensorflow-contrib would make import tf.contrib the same as before. Are version numbers of the new repos going to match the main TF release or are they just going to be independent?

Another thing I'd ask is when doing this migration, it should be done with preserving the git history as much as possible. That makes things much easier to follow in the new repos. Something like clone the entire TF repo then delete everything and commit then push to the new github repo.

facaiy:
I'm a little worried that tensorflow/contrib might meet the same maintenance challenges later like those of tensorflow core today.

bhack:
@facaiy That's why I proposed to track/update maturity and owners of sub-api. But seems that it will be defined in the self-organization process of the new SIG.

seanpmorgan:
@perfinion I second the idea to preserve git history, but I'm torn on keeping the same module name. There will likely be no compatibility guarantee between the new "contrib" repo and the still living 1.x module so I worry it'll create confusion.

As far as versioning, I think it would be nice to match with 2.x versions, but may be unrealistic for the SIG to go through a release cycle at the same cadence as the TF team.

dmitrievanthony:
Hi, let me remind that currently I have two opened pull requests with two new modules into `contrib`:

- Apache Ignite Dataset ([#21853](https://github.com/tensorflow/tensorflow/pull/21853)) which introduces new data source.
- Apache Ignite File System ([#22194](https://github.com/tensorflow/tensorflow/pull/22194)) which introduces new file system.

I really hope they will be accepted and merged soon. At the same time, if we are talking about 2.0 I think it makes sense to consider these modules as well. So, @martinwicke,  let me ask you do not forget about them.

facaiy:
> @bhack That's why I proposed to track/update maturity and owners of sub-api. But seems that it will be defined in the self-organization process of the new SIG.

Sounds great to me. 

Does it make sense to create a new organization, say tensorflow-thirdparty? And let SIG manage how to migrate those module into the new organization: move them totally into a big repository, or split them into each separate small repository. 

I mean this is for all: tensorflow/contrib, tensorflow/scientific, and tensorflow/io. All SIGs share the new organization and namespace. Is it worthwhile ?

bhack:
Are you moving tf.slim in model just for models retro-compatibility and virtually deprecate it or do you plan to release new models with this api? Having models that rely on core repository API will help to enforce the quality of the API and its implicit testing. I.e. High level API was often not in shape also cause many reference models stressed more slim.

yongtang:
@dmitrievanthony I think Apache Ignite would be a good fit for tensorflow/io.

There are two pending PRs https://github.com/tensorflow/tensorflow/pull/18224 (Avro) and https://github.com/tensorflow/tensorflow/pull/19461 (Parquet) that may also fit tensorflow/io. I think the Avro support may takes additional time. The Parquet support is awaiting the upstream Bazel's http_archive issue (https://github.com/bazelbuild/bazel/issues/5932), which should be resolved in Bazel 0.17.1 (https://github.com/bazelbuild/bazel/issues/5059).

Bazel 0.17.1 is planned to be released tomorrow (https://github.com/bazelbuild/bazel/issues/5059#issuecomment-420975574).

/cc @fraudies @galv

rfcs/20180907-contrib-sunset.md
```diff
@@ -0,0 +1,280 @@
...
+| gdr            |*byronyi (poxvoculi)  |                                      |
+| graph_editor       | purpledog        | delete                               |
+| grid_rnn           |(qlzh727 ebrevdo) | delete                               |
+| hadoop             | yongtang (mrry)  | move to tensorflow/io?               |
+| hooks              | ispirmustafa     | delete (already in core)             |
+| hvx                | satok16          | delete (redundant with NNAPI)        |
+| image              |                  | partial move to tensorflow/contrib?  |
+| input_pipeline     | rohan100jain     | delete                               |
+| integrate          | shoyer *mcoram   | move to tensorflow/scientific?       |
+| kafka              | yongtang (mrry)  | move to tensorflow/io?               |
```
dmitrievanthony: Lets add `ignite` ([#22210](https://github.com/tensorflow/tensorflow/pull/22210)) and `igfs` ([#22194](https://github.com/tensorflow/tensorflow/pull/22194)) here?

seanpmorgan:
Do we foresee a path to core from "contrib"? 

Will there be any coordinated additions to core and subsequent removals from "contrib" or is it best to have the repos live independently? I worry that having the same functionality in two places will cause problems (though as it exists today this still happens).

martinwicke:
@bhack > Are you moving tf.slim in model just for models retro-compatibility and virtually deprecate it or do you plan to release new models with this api? Having models that rely on core repository API will help to enforce the quality of the API and its implicit testing. I.e. High level API was often not in shape also cause many reference models stressed more slim.

We do not plan to release more models with slim, but we don't control what research groups do, and it's quite possible more models are published using slim. 

I do consider slim deprecated (for one, it does not work well in eager mode).

martinwicke:
@seanpmorgan 
> Do we foresee a path to core from "contrib"?
> Will there be any coordinated additions to core and subsequent removals from "contrib" or is it best to have the repos live independently? I worry that having the same functionality in two places will cause problems (though as it exists today this still happens).

I think there will be path, for instance, I can imagine a specialized layer appearing in contrib when a paper is published, and being absorbed to core when over time that layer starts being widely used. I hope that through the SIG we stay in close enough contact to coordinate add/deprecate/delete in such cases.

rfcs/20180907-contrib-sunset.md
martinwicke: I am not sure how interesting this function still is with eager execution. @mrry FYI.

martinwicke:
@yongtang @dmitrievanthony I have added ignite and igfs -- I think the others @yongtang mentioned also make sense for tensorflow/io. Should I add them to the list, or should we wait for their issues to be resolved?

martinwicke:
@facaiy 
> Does it make sense to create a new organization, say tensorflow-thirdparty? And let SIG manage how to migrate those modules into the new organization: move them totally into a big repository, or split them into each separate small repository (I like this second way, however it might be an exhausting task).
> 
> I mean this is for all: tensorflow/contrib, tensorflow/scientific, and tensorflow/io. All SIGs share the new organization and namespace. Is it worthwhile ?

I think we'll use tensorflow for now. If we end up with many SIGs, we can reconsider, but for now, I think it's easiest to stay in the TensorFlow org.

rfcs/20180907-contrib-sunset.md
martinwicke: Eager execution in 2.0 still allows you to have graphs (though we'd prefer you think of them as functions), and those can still be exported, executed remotely, etc. `constant_value` is not particularly useful with eager execution since you can simply look at any tensor (its value will be there), or run any function (which will produce a value). 

Or is the capability you need the implicit estimate of whether the value can be computed cheaply?

@asimshankar

yongtang:
Thanks @martinwicke I think `parquet` could be added. The PR https://github.com/tensorflow/tensorflow/pull/19461 has already been approved. Just waiting for Bazel 0.17.2 release (https://github.com/tensorflow/tensorflow/pull/22281) to pass the test failures.

The `Avro` support (https://github.com/tensorflow/tensorflow/pull/18224) seems to have been slowed recently. So maybe we could wait until the PR is ready?

rfcs/20180907-contrib-sunset.md
leandro-gracia-gil: Glad to read about graphs in 2.0. If you happen to have any links where I can learn more about this new design, I'll be very happy to take a look so I can make my comments more useful and targeted in the future.

Answering your question in TF 1.x terms, we need the ability to quickly and statically access the value of constant tensors and statically known shapes without involving a session or a tensor evaluation. I'm not fully sure how this would translate to TF 2.0, but we need something faster than tf.Tensor.eval().

martinwicke:
I believe for Tensors which you know are constants, the value information
is simply always available, same for shapes, so I do think this function is
redundant.

rfcs/20180907-contrib-sunset.md
asimshankar: I'm hoping to publish a proposal for eager and graphs in TF 2.0 soon (I'm working on the text, PR hopefully this week). See keep and eye out for that.

leandro-gracia-gil: @asimshankar 
Thanks, I'll definitely keep an eye.

@martinwicke 
I think we're on the same page here, but let me confirm.

"I believe for Tensors which you know are constants, the value information
is simply always available, same for shapes, so I do think this function is
redundant."

Is that the case too if I don't know if a tensor is constant? For example, if I'm writing some function and I get a tensor, can I simply try to "statically" get its value? (which I expect would work if it's constant or somehow statically evaluable, but fail otherwise)

From what you say it looks like I probably can, and if that's the case then this function would indeed become redundant, but just making sure.

fraudies:
@yongtang @martinwicke PR https://github.com/tensorflow/tensorflow/pull/18224 (avro dataset) will undergo a bit of a redesign. We'll merge the dataset & parse function per discussion on the PR and offline. tensorflow/io is a good fit post TF 2.0.

facaiy:
@martinwicke 

> I think we'll use tensorflow for now. If we end up with many SIGs, we can reconsider, but for now, I think it's easiest to stay in the TensorFlow org.

Yes, I understand. I agree with you.

rfcs/20180907-contrib-sunset.md
```diff
@@ -0,0 +1,282 @@
...
+| stat_summarizer    |                  | delete (no owner)                    |
+| stateless          | (*ebrevdo *girving) | replace our random ops with this? |
+| summary            |  nickfelt        | move to core, replacing tf.summary   |
+| tensor_forest      | nataliaponomareva yupbank | delete (moving to core)     |
+| tensorboard        |  nickfelt  jart  | move to tensorflow/tensorboard       |
+| tensorrt           | tobyboyd *samikama *aaroey *jjsjann123| move essential parts to core |
+| testing            | ispirmustafa     | move to core, make private           |
+| text               |                  | move to tensorflow/contrib?          |
+| tfprof             |                  | delete (replaced by tf.profiler)     |
+| timeseries         | gsundeep karmel  | move to tensorflow/estimator         |
```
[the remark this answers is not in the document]
martinwicke: Done. @karmel @bananabowl FYI.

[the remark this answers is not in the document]
allenlavoie: @terrytangyuan which parts are you using/interested in maintaining? @bananabowl has been looking into migrating some subset of the non-linear autoregressive model to core. But the state space models still have a somewhat uncertain future AFAIK.

kingspp:
@martinwicke This might help in sunsetting contrib package - https://tf-contrib-analyzer.herokuapp.com

martinwicke:
This is awesome.

rfcs/20180907-contrib-sunset.md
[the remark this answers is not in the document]
terrytangyuan: We are mainly using non-linear autoregressive models as well. State space models have slight overlap with TF Probability, e.g. [linear gaussian state space model](https://github.com/tensorflow/probability/blob/master/tensorflow_probability/python/distributions/linear_gaussian_ssm.py) but I think the state space estimators should fit in tensorflow/estimators. Perhaps we can mark them as experimental and think about compatibility/integration with TF Probability later. I can certainly help look into migrating those to core if you agree with that direction. Feel free to start a new thread over email so we are not distracting this Github issue.

rfcs/20180907-contrib-sunset.md
```diff
@@ -0,0 +1,282 @@
...
+| meta_graph_transform |  petewarden    | delete                               |
+| metrics            | brainnoise       | delete (replaced with OO metrics)    |
+| mixed_precision    | protoget reedwm  | delete                               |
+| model_pruning      | suyoggupta       | move to core                         |
+| mpi                | (poxvoculi)      |                                      |
+| mpi_collectives    | *jthestness (poxvoculi) |                               |
+| nccl               | (tobyboyd)       | move essential parts to core         |
+| nearest_neighbor   |                  | delete                               |
+| nn                 |                  | partial move to tensorflow/contrib?  |
+| opt                | *joshburkart apassos | move to tensorflow/contrib?      |
```
facaiy: @martinwicke Hi, I'm interested in `opt` module. My colleague and I contributed codes for adamax optimizer, elastic average optimizer, and agn_optimizer. I would love to help if needed.

rfcs/20180907-contrib-sunset.md
```diff
@@ -0,0 +1,282 @@
...
+| learn              | wicke       | delete (replaced by tensorflow/estimator) |
+| legacy_seq2seq     | ebrevdo (qlzh727) | delete (replaced by seq2seq)        |
+| libsvm             |                  | delete (no owner)                    |
+| linalg             | rmlarsen langmore | delete (moved to core)              |
+| linear_optimizer   | petrosmol (karmel) | move to tensorflow/estimator       |
+| lite             | aselle  petewarden | move to core                         |
+| lookup          | ysuematsu (ebrevdo) | move to core                         |
+| losses             |                  | partial move to tensorflow/contrib   |
+| makefile           |  petewarden      | delete (RPI build now uses bazel)    |
+| memory_stats       | wujingyue        | delete                               |
```
ppwwyyxx: Is there an alternative to query memory usage (especially on GPUs) ?

martinwicke: You can ask `Session.run` to fill out [`StepStats`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/step_stats.proto), which includes information on memory usage.

ppwwyyxx: It does not support the functionality of the "MaxBytesInUseOp" in contrib, which returns me `AllocatorStats.max_bytes_in_use`. The closest thing in `StepStats` seems to be [AllocatorMemoryUsed.allocator_bytes_in_use](https://github.com/tensorflow/tensorflow/blob/9ab01c6732dae1143e22713375a9cc7758216787/tensorflow/core/framework/step_stats.proto#L32), which [corresponds to AllocatorStats.bytes_in_use](https://github.com/tensorflow/tensorflow/blob/d5c5f8ecc124ee9a866318f2bd7082df9e38ebf2/tensorflow/core/common_runtime/step_stats_collector.cc#L179-L181) instead.

rfcs/20180907-contrib-sunset.md
leandro-gracia-gil: @martinwicke 
Just to not leave this pending, will it be possible to check if a given tensor is constant in TF 2.x? That should also help solving the problem. In 1.x you could check the type of the op the tensor belongs to, but I ignore if that will be possible in TF 2.x.

rfcs/20180907-contrib-sunset.md
martinwicke: Is AllocatorMemoryUsed.peak_bytes equivalent to that?

byronyi:
For the new repos under TensorFlow org, will contributors be required to sign the same SLA as in the main repo? And license (Apache 2) remains unchanged or cannot be changed?

rfcs/20180907-contrib-sunset.md
ppwwyyxx: If I'm not mistaken, those are [__per-node__ statistics](https://github.com/tensorflow/tensorflow/blob/6419058dc6d065c34cccf55b11a7ba02c2ef7b06/tensorflow/core/framework/step_stats.proto#L22-L24). However the "MaxBytesInUse" op in contrib gives global peak memory usage.

martinwicke:
@byronyi Yes, we will keep Apache2 (if nothing else, for the practical reason that it allows us to move coder around between repos since they are all licensed the same), and we are required to have the CLA signed for all contributors to any repos under the TensorFlow org.

martinwicke:
That's correct, you would have to add them up manually (or is there
something I'm missing?). If I'm right about this, I don't think the
separate ops are worth maintaining to avoid that summation.

rfcs/20180907-contrib-sunset.md
asimshankar: (FYI, re https://github.com/tensorflow/community/pull/18#discussion_r218270083 see #20)

rfcs/20180907-contrib-sunset.md
ppwwyyxx: (Following the response in the main thread)
I tried and sum of `peak_bytes` for all node always return me a much larger results than the `MaxBytesInUse` op, sometimes close to 2x larger.
There is little docs on what these fields mean, but I guess sum of peak memory per node is expected to be much larger than the overall memory consumption (which I'm interested in) since nodes can have non-overlapped lifetime.
Conceptually, I think that the peak memory consumption of a whole system cannot be the sum of any per-node statistics.

martinwicke:
That makes sense. I would be fine with adding the equivalent of
MaxBytesInUse to StepStats. @mrry do you have a sense of why we never did
that / if there's a reason that may be hard?

mrry:
@martinwicke I think @yuefengz added the code for tracking memory usage via `StepStats`, so he'd probably know best.

(As an aside: in 2.0 without `Session.run()`, do we have a plan to expose `StepStats`?)

martinwicke:
I believe @alextp has a plan, yes. Not sure how exactly.

alextp:
@mrry Currently the TFE context exposes StepStats though the API is not super nice: https://github.com/tensorflow/tensorflow/blob/153578f3c90ca423501151adcbaf6b81e05e2440/tensorflow/python/eager/context.py#L587

Essentially you call `context.context().enable_run_metadata()` once and then call `context.context().export_run_metadata()` to get a proto with what happened since you last called export.

asimshankar:
@mrry @alextp @martinwicke : Also, as per #20, for any function call you can provide `RunOptions` and `RunMetadata`, so we can obtain `StepStats` for individual defuns as well.

rfcs/20180907-contrib-sunset.md
yuefengz: The (`AllocatorStats.max_bytes_in_use`)[https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/allocator.h#L54] should give you the peak memory per device. That op queries that field. You can also use another way to query that field: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/grappler/clusters/single_machine.cc#L237

martinwicke: I think @ppwwyyxx is interested in querying this from Python. We don't have a way to do that any more. Should the peak device memory usage be added to StepStats (or something in there?)

yuefengz: We can add a new field to StepStats to hold global peak memory per allocator if necessary.

rfcs/20180907-contrib-sunset.md
```diff
@@ -0,0 +1,282 @@
...
+| distribute         | josh11b priyag   | move to core                         |
+| distributions      | jvdillon     | delete (moved to tensorflow/probability) |
+| eager              | asimshankar      | move to core                         |
+| estimator          | ispirmustafa mikecase | move to tensorflow/estimator    |
+| factorization      | agarwal-ashish   | delete (rebuild in core pending)?    |
+| feature_column     | ispirmustafa     | move to core (experimental)          |
+| ffmpeg             | fredbertsch      | delete                               |
+| framework          |               | partially move to core, delete the rest |
+| fused_conv         |                  | delete                               |
+| gan                | joel-shor        | move to separate repo                |
```
facaiy: @galeone TFGAN will be moved to a separate repo. You can find more details here.

foxik:
Hi, we use `tf.contrib.crf` quite a lot and it would be great to have it in TensorFlow, but it does not seem to have an owner -- I would be willing to help with the merge (i.e., become an owner if there is none).

martinwicke:
Please reach out and join addons@tensorflow.org. I think crf is likely to
end up a part of tensorflow/addons.

>

drheatherwalker:
Is there any update on the fate of WALSMatrixFactorization and WALSModel in tensorflow 2.0?

PR #20, opened 2018-09-20
RFC: Functions, not sessions in 2.0

asimshankar:
**Review period closes 2018-10-04**

# Functions, not sessions in 2.0

| **Author(s)** | ashankar@google.com, joshl@google.com                |
| **Sponsor**   | apassos@google.com                                   |
| **Updated**   | 2018-09-18                                           |

Proposal to make TensorFlow be more "Pythonic" in 2.0. In four bullet points:

- Encourage the encapsulation of graph computation as Python functions 
(where the graph is executed when the function is invoked, instead of via Session)
- Align "state" in the TensorFlow runtime (e.g., resource tensors like those that back tf.Variable objects) with state in the Python program (e.g., Python objects corresponding to the runtime state with lifetimes attached to each other).
- Make it easy to export these encapsulations to a GraphDef+Checkpoint and/or SavedModel.
- Enable eager execution by default.

### DIFF
--- a/rfcs/20180918-functions-not-sessions-20.md
+++ b/rfcs/20180918-functions-not-sessions-20.md
@@ -21,6 +21,10 @@ This document elaborates on the proposal to make TensorFlow be more "Pythonic" i
 
 This document liberally employs the use of sample code to describe the end-user effect of proposed changes.
 
+(We say "encourage" instead of "require" since removing the Session API from the
+Python frontend within a year may be an unrealistic aspiration.  Particularly
+given the use in Estimators and the use of MonitoredSession and hooks. The
+`Session` API may have to stick around in `tf.compat.v1`).
 
 
 ## Design Proposal
@@ -30,7 +34,7 @@ This document liberally employs the use of sample code to describe the end-user 
 Today, the TensorFlow graph defines the union of all computation that the author of the graph may be interested in. The actual computation to execute is defined by the arguments to `tf.Session.run`. Once this subgraph is defined, the runtime can optimize and execute. For example, consider the following:
 
 
-```
+```python
 import tensorflow as tf
 
 x = tf.placeholder(tf.float32)
@@ -48,7 +52,7 @@ z1 = sess.run([z], feed_dict={x: 2., y: 2.}) # 4.0
 Though there is one `tf.Graph` object the user is interacting with (`tf.get_default_graph()`), the two `sess.run` calls are executing different programs (indeed the runtime ends up with two separate `Graph` objects in C++, one for each program), equivalent to:
 
 
-```
+```python
 def compute_z0(x):
   return tf.add(x, tf.square(x))
 
@@ -60,7 +64,7 @@ def compute_z1(x, y):
 The core proposal of this document is the alignment between computation expressed in Python and the computation executed by the runtime. Instead of defining a graph and then selecting the subgraph to execute at `sess.run()` time, the exact computation of interest is encapsulated in a Python callable. For example, the program above that uses `sess.run()` to compute `z0` and `z1` can be written as:
 
 
-```
+```python
 import tensorflow as tf
 
 @tf.defun
@@ -97,7 +101,7 @@ If one was not explicitly specified by the user (as an argument to `defun`), the
 Changes in input signature result in a new graph being traced. For example:
 
 
-```
+```python
 @tf.defun
 def f(x):
   one = tf.constant(1, dtype=x.dtype)
@@ -187,7 +191,7 @@ Worthy of note here - in TensorFlow 1.x, the memory underlying the variables `W`
 In TensorFlow graphs today, control dependencies are sometimes needed to ensure correct evaluation order. For example, consider the following:
 
 
-```
+```python
 v = tf.Variable(1.0)
 init_op = tf.global_variables_initializer()
 assign_op = v.assign(2.0)
@@ -205,7 +209,7 @@ with tf.Session() as sess:
 The output here is not deterministic, since `val` may evaluate to either 1.0 or 2.0 depending on whether the runtime happened to execute `assign_op` before `y` or not. `tf.control_dependencies` is a mechanism provided to add annotations at graph construction time to influence graph execution. The TensorFlow user, a Python programmer, is thus forced to think about two execution models - TensorFlow graphs and the Python interpreter. To eliminate this cognitive load, `defun` will automatically insert control dependencies to ensure that operations that produce or consume a given `DT_RESOURCE` tensor and operations that are marked stateful (`REGISTER_OP(...).SetIsStateful()`) follow graph construction order. Thus:
 
 
-```
+```python
 v = tf.Variable(1.0)
 @tf.defun
 def f():
@@ -226,7 +230,7 @@ In the above code, no `tf.Variable` objects are created inside a `tf.defun` deco
 Note that if the function naturally creates state only on the first trace, all is well:
 
 
-```
+```python
 v = None
 
 @tf.defun
@@ -267,7 +271,7 @@ We've introduced a single new symbol: `defun` that consumes a Python function an
 Since new graphs are traced when new input signatures are encountered, a `defun` can encapsulate multiple graphs. For example, consider the following snippet:
 
 
-```
+```python
 @tf.defun
 def f(x):
   return tf.square(x)
@@ -283,7 +287,7 @@ There are two graphs created here - one which corresponds to the `Square` operat
 The same holds for the case where arguments are not `Tensor`s, for example:
 
 
-```
+```python
 @tf.defun
 def f(x, use_multiply):
   return tf.multiply(x, x) if use_multiply else tf.square(x)
@@ -298,7 +302,7 @@ will result in 2 graphs being created.
 Note that the "type" of `Tensor` inputs to the function also incorporates the shape. For example:
 
 
-```
+```python
 @tf.defun
 def f(x): return tf.add(x, 1.)
 f([2.0])
@@ -321,7 +325,7 @@ will result in 3 graphs being created:
 Tracing the decorated function to create a new graph on each input shape is a conservative choice (allowing for `f` to create graphs dependent on the shape), which may be unnecessary. Users can explicitly specify an input signature to ensure that the same graph is used for multiple inputs. For example:
 
 
-```
+```python
 @tf.defun(input_signature=((tf.float32, [None]))
 def f(x): return tf.add(x, 1.)
 
@@ -342,7 +346,7 @@ f([[2.0]])    # Raises an error as the arguments don't match the input signature
 If a member function of a class does not create variables, it may be decorated with `@tf.defun` and it will work:
 
 
-```
+```python
 class ScalarModel(object):
   def __init__(self):
     self.v = tf.Variable(0)
@@ -366,7 +370,7 @@ assert int(model2.v) == 5
 This works since `increment()` has `self` as a non-tensor argument, and a new trace will be created for each value of `self`. However, if variables are created in a method, we want to allow a new set of variables for every instantiation of `self`. You get this behavior by using `@tf.method`:
 
 
-```
+```python
 class AnyShapeModel(object):
   def __init__(self):
     self.v = None
@@ -398,7 +402,7 @@ In addition, as long as all variable creation/initialization happens while we ar
 The definition of `tf.defun` above is careful to check that invoking a decorated Python function would have the same behavior as invoking an undecorated function. This is to guard against it being passed code from TensorFlow v1.x that expects to only be called once (and relies on things like graph collections to track which variables are created), for example:
 
 
-```
+```python
 def f(x, do_add):
   v = tf.Variable(5.0)
   if do_add:
@@ -412,7 +416,7 @@ def f(x, do_add):
 For this case, we use a different API, `tf.compat.v1.wrap_function`, that treats any created variables as static local state:
 
 
-```
+```python
 f_add = tf.compat.v1.wrap_function(f, tf.TensorSpec(tf.float32, ()), True)
 
 assert float(f_add(1.0)) == 6.0
@@ -446,7 +450,7 @@ We recognize that code written for TensorFlow 1.x commonly does not encapsulate 
 With the `tf.compat.v1.wrap_function` proposed above, most graph construction library functions written against TensorFlow 1.x can be incorporated into TensorFlow 2.x programs.
 
 
-```
+```python
 def f(x):
   W = tf.compat.v1.get_variable(name="weight", shape=[10, 10])
   b = tf.compat.v1.get_variable(name="bias", shape=[10],
@@ -458,7 +462,7 @@ def f(x):
 
 
 
-```
+```python
 f = tf.compat.v1.wrap_function(f, tf.placeholder(tf.float32, None))
 print(f(make_input_value()))
 assert len(f.variables) == 3
@@ -789,7 +793,7 @@ The author realizes that this section can do with more detail. However, to keep 
 `defun` decorates a graph construction function and transparently recreates graphs if needed. However, this does mean that if the function has data-dependent control flow then though the function will execute fine with eager execution enabled, `defun` decorating it will fail. For example:
 
 
-```
+```python
 def f(x, y):
   if tf.equal(y, 0.0):
     return y
@@ -809,7 +813,7 @@ df(x, y)  # Will raise an error complaining about the data-dependent control flo
 To fix this, one would have to use the graph construction APIs for control flow (`tf.cond`, `tf.while_loop`):
 
 
-```
+```python
 def f(x, y):
   return tf.cond(tf.equal(y, 0.0), lambda: y, lambda: x/y)
 
@@ -833,7 +837,7 @@ This situation can be improved with the help of  go/tf-autograph to allow expres
 The summary writing operations (<code>[tb.summary.scalar](https://www.tensorflow.org/api_docs/python/tf/contrib/summary/scalar)</code>, <code>[tb.summary.image](https://www.tensorflow.org/api_docs/python/tf/contrib/summary/image)</code> etc.) can be naturally placed in the graph by using them in a <code>defun</code>-decorated function. These operations require two "external" inputs - the summary writer resource and the condition, that will be picked up from the context (e.g., <code>[tb.summary.create_file_writer](https://www.tensorflow.org/api_docs/python/tf/contrib/summary/create_file_writer)</code> and <code>[tb.summary.record_summary_every_n_global_steps](https://www.tensorflow.org/api_docs/python/tf/contrib/summary/record_summary_every_n_global_steps)</code>).  When defining the graph, these inputs are converted to placeholders, which are then resolved at function invocation time. Thus, something like this:
 
 
-```
+```python
 writer = tf.contrib.summary.create_file_writer('/tmp/test')
 with writer.as_default(), tf.contrib.summary.always_record_summaries():
     f()
@@ -845,7 +849,7 @@ with writer.as_default(), tf.contrib.summary.never_record_summaries():
 Will write one summary to `writer` whether `f` is defined as:
 
 
-```
+```python
 def f(): 
     tb.summary.scalar("loss", compute_loss()) 
 ```
@@ -854,7 +858,7 @@ def f():
 Or
 
 
-```
+```python
 f = tf.contrib.eager.defun(f)
 ```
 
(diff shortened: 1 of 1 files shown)


rfcs/20180918-functions-not-sessions-20.md
```diff
@@ -0,0 +1,1002 @@
...
+```
+
+
+To support this `defun` imposes some requirements on the decorated function:
+
+
+
+1.  State (like `tf.Variable` objects) are only created the first time the function `f` is called. \
+How that is accomplished is left up to the implementation of `f`. \
+If any variables are created in the first execution of `f`, then `@tf.defun` will trace `f` a second time in order to record the behavior that will be used from then on. No variables may be created during that second trace, or any other trace after that (due to different dtypes, shapes, or non-tensor arguments).
```
ageron: I'm a bit confused by this sentence: what do you mean by "to record the behavior that will be used from then on"? Are you referring to the automatic addition of control dependencies for stateful ops?
Also, it seems a bit dangerous to me that `f` may be called once or twice depending on whether or not it creates variables. Can't you just "record the behavior" the first time round? Is there no way around this optional double-run? I'm sure tons of people won't expect this and their code will end up doing something twice when it should have been run once. Seems tricky to debug.

asimshankar: "Record the behavior that will be used then on" as in any operations run as part of the initialization should not be part of the graph (so nothing to do with control dependencies to ensure program order). For example, consider:

```python
v = None
@tf.defun
def f(x):
  global v
  if v is None:
    a = tf.random_uniform(...)
    b = tf.exp(a)
    v = tf.Variable(a)
  return x + v
```

We don't want the operations to initialize the variable (`tf.random_uniform()`, `tf.exp()`) to be included in the graph defined by the trace.  So the proposal is to run `f` twice. The first time it will create the variables and the second time it will generate the graph that will be used in subsequent calls to (the decorated) `f`.  Does that clarify the message? Though, I see your concern, and am very open to suggestions.

rfcs/20180918-functions-not-sessions-20.md
```diff
@@ -0,0 +1,1002 @@
+# TensorFlow 2.0: Functions, not Sessions.
+
+| Status        | (Proposed / Accepted / Implemented / Obsolete)       |
+:-------------- |:---------------------------------------------------- |
```
ewilderj: small nit. status is "Proposed"

asimshankar: Doh! Done!

rfcs/20180918-functions-not-sessions-20.md
ageron: Thanks for the clarification, @asimshankar . Wouldn't it be cleaner to just require variables to be created separately, as in Keras (where variables are created in `build()`, and computations are performed in `call()`)? This way you wouldn't have to handle the complexity of dealing with variables created within `f()`. Perhaps there's a great use case for this kind of code, but I don't see it, and it seems like a lot of extra complexity (not just in the implementation for the TensorFlow team, but also extra cognitive load for anyone using `defun()` or learning TF).

rfcs/20180918-functions-not-sessions-20.md
```diff
@@ -0,0 +1,1002 @@
...
+
+f(x, y) # Will be 1.0
+
+df = tf.defun(f)
+df(x, y)  # Will be 1.0
+```
+
+
+This situation can be improved with the help of  [autograph](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/contrib/autograph) to allow expression of control flow in Python. Whether autograph will be enabled by default or not is still under debate, but the option will be there as a flag on defun.
+
```
ageron: Are you saying that this will work?

```python
def f(x, y):
    if y == 0.0:  # or tf.equal(y, 0.0)?
        return y
    else:
        return x / y

df = tf.defun(f, autograph=True)
df(x, y)
```

rfcs/20180918-functions-not-sessions-20.md
```diff
@@ -0,0 +1,1002 @@
...
+*   Users of **<code>Estimator</code>** will see no change
+    *   Canned Estimators are black boxes that create and train models. Enabling eager execution will have no effect on their usage. This is true today.
+    *   The <code>model_fn</code> of a regular (non-canned) <code>Estimator</code> will remain as a graph construction function.
+*   <strong>[SavedModel](https://www.tensorflow.org/guide/saved_model#save_and_restore_models)</strong> will continue to be the format encouraged for exporting trained models
+    *   Crudely speaking, a SavedModel encapsulates a Graph, a checkpoint of variable values, and some metadata like signature information (names of input and output tensors).
+    *   A path will be provided to easily export models in this format (e.g., via <code>tf.keras.Model.save()</code>). There may be instances where converting the Python code to a graph is not trivial, in which case, exporting to a SavedModel (and thus a Graph) will fail.
+
```
ageron: Could you give an example of such a case where exporting to a `SavedModel` will fail?

rfcs/20180918-functions-not-sessions-20.md
```diff
@@ -0,0 +1,1002 @@
...
+### Creating state inside a `defun`
+
+How state (`DT_RESOURCE` tensors)  created inside a `defun` should be handled is actively being debated. Options include:
+
+
+
+1.  "Lifting" state out as a static local function variable
+1.  Mimic the undecorated code - creating and destroying variables on each call.
+
```
ageron: Did you consider just raising an exception? Why not just impose that functions decorated by `defun` should not create state? Can we list use cases that cannot be implemented in any other (simpler) ways?

ngc92:
For python 3 users, it would be nice if instead of specifying expected types and shapes as an argument to `defun`, pythons own type annotations would be parsed. So instead of
```
@tf.defun(input_signature=((tf.float32, [None, 5]))
def f(x):
  pass
```
we could write
```
@tf.defun
def f(x: tf.TensorSpec[tf.float32, (None, 5)):
  pass
```
While this could not be used in tensorflow itself (unless there was a build step that when targeting py2 would convert the latter into the former), but would give projects not concerned with backward compatibility a nicer syntax, and give linters the ability to detect some API misuses before the program is run.

rfcs/20180918-functions-not-sessions-20.md
asimshankar: Added a note. Thanks.

rfcs/20180918-functions-not-sessions-20.md
asimshankar: Yup, added. Thanks!

rfcs/20180918-functions-not-sessions-20.md
asimshankar: Great question. I gotta run now, but I'll respond to and let's continue discussion in the [thread above](https://github.com/tensorflow/community/pull/20#discussion_r219312959).

(Resolving this conversation to collapse with the thread above).

rfcs/20180918-functions-not-sessions-20.md
```diff
@@ -0,0 +1,1008 @@
...
+   </td>
+  </tr>
+</table>
+
+
+
+### Derived/Related Graphs
+
+One reservation expressed by TensorFlow graph/session enthusiasts today is that the ability to write generic analysis/inspection tooling on graphs, precluding the need to understand or modify the Python code that constructed the graph, is important to them. To put it differently, some find it easier to navigate the `GraphDef` program than navigating the Python program. \
```
ppwwyyxx: I'm heavily relying on the ability to execute custom subgraphs by using many SessionRunHooks, from what I see here the proposed solution seems unable to give me the flexibility TF 1.0 has.

Consider the following situation:
There are some intermediate tensors/summaries in the graph, which I want to evaluate (i.e. get the results as a numpy array) every once a while but at __different frequency__ (or at other different pre-conditions). Some may need to be evaluated per-step, some every 10k steps. And maybe none will need to be evaluated at all after I finished debugging. These combinations essentially create lot of subgraphs (or the new style "Functions") to evaluate.

With TF1.0 I can easily write SessionRunHooks that evaluate the tensors/summaries I want, at the frequency or other conditions I specify. I can easily modify the tensors/frequencies to evaluate, by editing the list of SessionRunHooks I use. And I can easily disable them all by removing the hooks. By doing that I do not have to touch the code where the tensors/summaries are originally defined (which are likely to be located in multiple files).

What's the alternative here if SessionRunHooks are abandoned?

carlthome:
Definitely prefer `tf.function` over `tf.defun`! Why would we de-fun TensorFlow programs? :upside_down_face:

omoindrot:
It also makes sense to have `@tf.function` if we have `@tf.method` (and not `@tf.defmethod`).

seanpmorgan:
Strongly agree with `@tf.function` or `@tf.func`. My first look at `tf.contrib.defun` was that it signaled some kind of defunct function.  Since the decorator will be so commonly used, it's worth making sure the name is intuitive at first glance and doesn't need to be looked up. 

Looking at common python decorators, I find the most clear decorators use either:
- an understood noun (`@classmethod`, `@propery`, `@app_endpoint()`) 
- written out action that wraps it (`@autograph.convert()`, `@login_required`, `@atexit.register `)

seanpmorgan:
Separately, will we need to handle `tf.defun` wrapping a function which uses `tf.py_func`? Seems like if nothing else it may need to throw an error since it wont be serialized in the graph.

Related to https://github.com/tensorflow/tensorflow/issues/10282

rfcs/20180918-functions-not-sessions-20.md
ageron: Here's a small [python program](https://gist.github.com/ageron/1a92faa5e72134f49e6bcbcbeecfe8af) that shows different ways to use `@tf.function` (I prefer that name rather than `@tf.defun`) and `@tf.method`.  The first one is nice and clean, and the others are gradually uglier. Only the ugliest solution requires creating some TF state within the function decorated by `@tf.function`.
I think we should just forbid creating any TF state within a function or method decorated by `tf.function` or `tf.method`. But of course there might be use cases I haven't thought of (`python -m this | grep Special`).

ngc92:
How does this interact with device selection? I.e. 
```
@tf.defun
def f(...):
    ...

with tf.device("/gpu:0"):
  f()
with tf.device("/gpu:1"):
  f()
```
the different f's have the same signature here, but to execute them on different devices we need to different graphs.

samjabrahams:
At a high level, I really like the idea to push towards a more functional/composable API. For me, having to figure out a clean way to reuse components of a graph while passing in different inputs (e.g. dataset inputs during training, then export a graph with placeholders for inference) has always been an annoyance. I will likely have more comments (currently traveling and have limited time to type things up), but I had a couple of concerns I want to bring up:

* Currently, the Session API takes a few runs to "warm up". That is, the first couple of runs will run slower. I suppose this is more of a question about Eager in general, but are there risks of things always running slowly by not using a session? Or will there actually be less overhead due to not needing to traverse a graph to find what operations to run?
* I'm concerned about this change being a half-measure. I haven't done user research myself, but I think that the difficulty of new user on-boarding is more than just the fact that the graph is run lazily. It's also that the API is _massive_ and thus prone to make new user's eyes glaze over. Not only that, but there is limited guidance on what idomatic TensorFlow code should use. This leads to confusion and a lot of people who have to learn the same lessons over and over. I believe that one of the largest opportunities of 2.0 is the ability to remove cruft and imagine what TensorFloe could be if it was built from the ground up. I understand that there are needs in terms of enabling backwards compatibility to prevent alienating the userbase, but right now I think this risks adding bloat in the form of "there's yet another way to run a graph!". 
    * I'm not saying that this API is exactly what we want; I need more time to digest it. What I would say is that either the scope should be much larger, or it should be reduced so that it doesn't require as many awkward compromises/workarounds in order to kind-of work with existing semantics.
* I think there should be more weight given to/conversation about diverging the Python vs non-Python APIs. I know that most of the people using C++/Go APIs may be considered "power users", and thus better equipped to deal with additional things to learn, but one of the strengths of TensorFlow is that the semantics between your Python/training code are similar to those in deployment. It's hard to argue against "Python code should be Pythonic", but I wonder if there is a way to keep the APIs unified at the same time.

asimshankar:
@samjabrahams , thanks for the comments. Some responses:

* "warm up": There should be no risk of running slowly by not using a session. Any warmup (e.g., creating the kernels etc.) may happen on the first time the function is executed, but there should be no overhead beyond that. As you guessed, there may even be less overhead (the ~10us or so sometimes spent in looking up the exact graph to execute based on feeds and fetches).

* This change is just one of many towards 2.0. You'll see talk about reducing the API surface in other RFCs like #15 #16 #18, and others coming. Is there something specific to this proposal that you think can be improved?

* Point taken about the other language APIs. I'll follow up on that.

asimshankar:
@ngc92 , re https://github.com/tensorflow/community/pull/20#issuecomment-423687366 - yes, different graphs will be created for those. This happens with `tf.contrib.eager.defun` today, as the context of execution is implicitly a part of the input signature. I'll clarify that in the document too.

rfcs/20180918-functions-not-sessions-20.md
ppwwyyxx: Thanks for the response. Actually I wasn't thinking about using names (strings). String manipulation can easily get wrong so I'm using the tensor objects. A more concrete example:
```python
class Model():
    def my_graph(self, x):
        self.maybe_evaluate = [f(x), g(x), h(x)]
        return output(x)

model = Model()
y = model.my_graph(x)
hooks = [ EvalHook(model.self.maybe_evaluate[0], frequency=10),
        EvalHook(model.self.maybe_evaluate[1], frequency=20), ... ]
while True:
  hooked_session.run(y)
```
Back to your comments: my Python graph construction code is already structured to return these tensors explicitly, because being explicit is less error-prone. 

To use the `NewGraphFunction` for the above flexibility, I assume I would need to make this wrapper (only implementing the outputs part, not the inputs part):
```python
def call_with_hook(f, fetches, hooks):  # f is a defun
    graph = f.graph_function((tf.float32, none)).graph
    fnew = tf.NewGraphFunction(
      graph,
      inputs=[],
      outputs=fetches + hooks.before_run())
    ret = fnew()
    hooks.after_run(ret[len(fetches):])
    return ret[:len(fetches)]
while True:
  call_with_hook(f, y, hooks)
```
and will use this `call_with_hook` wrapper in every iteration.

I can accept this solution, with the hope that `NewGraphFunction` can accept tensor objects 
 in addition to names (presumably that's easy), and it's fast enough (maybe I was just terrified by the name "new" which sounds slow).

asimshankar: Something along those lines.
You may still want to cache the updated graph function in each hook (parallel to how any graph construction code is typically in `SessionRunHook.after_create_session` and not `SessionRunHook.before_run`). So more like:

```python
class Hooked(object):
  def __init__(self, f, fetches, hooks):
    graph = f.graph_function((tf.float32, None)).graph
    self.fnew = tf.NewGraphFunction(graph, inputs=[], outputs=hooks.before_run())
    self.hooks = hooks
    self.fetches = fetches

  def __call__(self):
    ret = self.fnew()[len(fetches):]
    self.hooks.after_run(ret)
    return ret

hooked = Hooked(f, y, hooks)
while True:
  hooked()
```

(This is just a rough off-the-top-of-my-head draft demonstrating that such hooking is possible and hopefully intuitive. The finer details may still need some working out)

ppwwyyxx: But the point of my use case is that `before_run()` may return different fetches in each iteration and will therefore need to be called at each iteration -- this is also how the hooks work in 1.0. This makes it necessary to create many "GraphFunction" in 2.0, and I'm wondering whether it will have any negative performance issues compared to 1.0.

asimshankar: (Resolving as I take @ppwwyyxx 's thumbs-up emoji to mean that this concern has been addressed :)

rfcs/20180918-functions-not-sessions-20.md
```diff
@@ -0,0 +1,1123 @@
...
+  multiply = True
+
+p = Params()
+@tf.function
+def f(x, y):
+  return tf.multiply(x, 2.) if y.multiply else tf.add(x, 2.)
+
+f(3., p) # Returns 6.0
+y.multiply = False
+f(3., y)  # Mutations to `y` may not trigger a retrace, so might still return 6.0
```
facaiy: typo?

```python
p.multiply = False
f(3., p)  # Mutations to `p` may not trigger a retrace, so might still return 6.0 
```

asimshankar: Doh! Yup, thanks. Fixed.

rfcs/20180918-functions-not-sessions-20.md
```diff
@@ -0,0 +1,1123 @@
...
+```
+
+
+This situation can be improved with the help of  [autograph](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/python/autograph) to allow expression of control flow in Python. Whether autograph will be enabled by default or not is still under debate, but the option will be there as a flag on function. For example:
+
+
+```python
+df = tf.function(f, autograph=True)
+f(x, y) # Will be 1.0
+```
```
alanhdu: FWIW, I think that doing `tf.function(autograph=True)(f)` would be a lot more natural, since that enables something like:

```python
@tf.function(autograph=True)
def f(x, y):
     ...
```

rfcs/20180918-functions-not-sessions-20.md
asimshankar: Of course! We do intend to keep and improve various good things in TensorFlow, including the various ecosystem components. Just that this RFC is aimed at the lower level APIs so doesn't get into the details of all the higher level ecosystem. Thanks.

rfcs/20180918-functions-not-sessions-20.md
asimshankar: Yup, fixed. Thanks!

rfcs/20180918-functions-not-sessions-20.md
asimshankar: We're using `@tf.function` now, and there is no need for `@tf.method` as the one decorator suffices. (The RFC is updated with that). So thanks for nudging in that direction :)

Regarding forbidding creation of state - the nice and clean examples you provided will (and should be) encouraged. In fact, the documentation for `tf.function` does encourage state creation within the body of the decorated function.

From a pragmatic standpoint: This caveat about creating state has made it possible to wrap large blocks of existing third-party/user code/libraries in a `tf.function`. For example, the `__call__` method of Keras layers create state on the first invocation (since they invoke `build()`). So, allowing state creation allows things like:

```python
m = <some tf.keras.Model>
m.call = tf.function(m.call) # Use graph execution for the Keras model
m(<some inputs>)
```

without having to redesign the implementation of  Keras.

So for now I'd keep this (for a preview release) and reconsider as we get some more mileage before the final 2.0 release.

Once again, thanks a bunch for your comments.

asimshankar:
@goldiegadde : I updated the document in place based on the meeting notes.

goldiegadde:
> @goldiegadde : I updated the document in place based on the meeting notes.

Thanks @asimshankar . Can you push a version of the doc with Status as "Accepted". We can then proceed to merge it.

asimshankar:
@goldiegadde : Done. Thanks.

goldiegadde:
Thanks everyone for reviewing the RFC. I am going to merge this request now.

rfcs/20180918-functions-not-sessions-20.md
asimshankar: That doesn't exist yet, still a TODO but a bit lower down in the priority list at the moment.
(FYI @alextp )

alextp: tf.graph_function ended up being called tf.compat.v1.wrap_function and the object it returns has a prune method, which does what NewGraphFunction does AFAIU

ppwwyyxx: @alextp The name of `compat.v1` means this feature is not considered part of TF 2.0, right?
Are we still going to have this feature in TF 2.0? I would not like to see code full of `tf.compat.v1` because 2.0 lacks features in 1.0.

alextp: tf.compat.v1 symbols will be usable throughout the lifetime of tf 2.x but will not be kept on 3.x and beyond (and we have no idea when 3.x will come)

dstiers11 (approved):

PR #25, opened 2018-10-17
RFC: Distribution Strategy - Revised API

guptapriya:
Review period closes 2018-11-08

# Distribution Strategy - Revised API

| Status        | Proposed                                                                  |
| :------------ | :------------------------------------------------------------------------ |
| **Author(s)** | cjfj@google.com, dominikg@google.com, jhseu@google.com, joshl@google.com, |
|               | petebu@google.com, priyag@google.com, tomhennigan@google.com              |
| **Sponsor**   | wicke@google.com                                                          |
| **Updated**   | 2018-11-12                                                                |

## Objective

This document presents a proposal to seek feedback on a revised [Distribution Strategy](https://www.tensorflow.org/api_docs/python/tf/contrib/distribute/DistributionStrategy) API and illustrate its usage in various situations. Distribution Strategy aims to allow users to easily distribute their computation across 
different GPUs, TPUs and multiple machines. Until now, the recommended usage of Distribution Strategy has been through TensorFlow high level training frameworks such as [`tf.keras`](https://www.tensorflow.org/guide/keras) and [`Estimator`](https://www.tensorflow.org/guide/estimators). In this proposal, we want to show how one can use Distribution Strategy APIs directly for distributing custom training loops, and get feedback on those APIs. This use case is important for many users who want more control of their training loops, such as ML researchers. We have tested a similar version of this API internally with many researchers in Alphabet and the current proposal is based on their feedback.

This is also an opportune time to get public feedback as we are in the process of migrating Distribution Strategy APIs from `tf.contrib` to core TensorFlow (as part of TF 2.0). As part of the move, we want to improve and reorganize the APIs to make them more user friendly and understandable.

### DIFF
--- a/rfcs/20181016-replicator.md
+++ b/rfcs/20181016-replicator.md
@@ -347,6 +347,7 @@ class DistributionStrategy(object):
   def broadcast(self, value: tf.Tensor) -> PerReplica[tf.Tensor]:
     """Broadcasts the given value from the current device to all replicas."""
 
+  @property
   def extended(self) -> DistributionStrategyExtended:
     """Returns an instance of the class implementing additional DistributionStrategy functionality.
     
@@ -571,6 +572,13 @@ class InputContext(object):
     If `replication_mode` == `PER_WORKER`, this is the number of workers.
     If `replication_mode` == `PER_REPLICA`, this is total number of replicas.
     """
+
+  def per_replica_batch_size(self, global_batch_size: int) -> int:
+    """Returns the per replica batch size user should use in their input pipeline, given desired global batch size.
+
+    It would be computed as per_replica_batch_size = global_batch_size // num_replicas_in_sync. global_batch_size should be divisible by num_replicas_in_sync. This will throw an error if not.
+    """ 
+
 ```
 
 #### InputIterator class
@@ -708,8 +716,7 @@ with strategy.scope():
   optimizer = tf.train.MomentumOptimizer(learning_rate, 0.9)
 
 def input_fn(ctx):
-  assert effective_batch_size % ctx.num_replicas_in_sync == 0
-  return imagenet.ImageNet(effective_batch_size // ctx.num_replicas_in_sync)
+  return imagenet.ImageNet(ctx.per_replica_batch_size(effective_batch_size))
 
 def step_fn(inputs):
   image, label = inputs
@@ -791,8 +798,7 @@ def input_fn(ctx):
   d = d.interleave(tf.data.TFRecordDataset, cycle_length=num_readers)
   d = d.map(parser_fn)
 
-  assert effective_batch_size % ctx.num_replicas_in_sync == 0
-  return d.batch(effective_batch_size // ctx.num_replicas_in_sync)
+  return d.batch(ctx.per_replica_batch_size(effective_batch_size))
 
 ```
 
@@ -807,8 +813,7 @@ def sample_noise(batch_size):
       shape=(batch_size, num_latents), mean=0.0, stddev=1.0)
 
 def input_fn(ctx):
-  assert effective_batch_size % ctx.num_replicas_in_sync == 0
-  batch_size = effective_batch_size // ctx.num_replicas_in_sync
+  batch_size = ctx.per_replica_batch_size(effective_batch_size)
   ds = cifar.Cifar10(batch_size)
   return ds.map(lambda x: (x['image'], sample_noise(batch_size)))
 
--- a/rfcs/20181016-replicator.md
+++ b/rfcs/20181016-replicator.md
@@ -363,6 +363,33 @@ An instance of the `ReplicaContext` can be retrieved by calling `tf.distribute.g
 
 ```python
 class ReplicaContext(object):
+  “””A context within replicated computation.
+
+  IMPORTANT: The ordering of calls to cross-replica interactions must be identical in all replicas.
+  This includes `merge_call` and all communications methods (`all_sum`, etc.). Where possible, an exception
+  will be raised. However, some user errors are not detectable and may lead to silent numerical errors.
+  Particular care should be taken when iterating over Python dictionaries.
+
+  Example of incorrect usage:
+  ```
+  losses = {"loss1": ..., "loss2": ...}
+  total_losses = {k: ctx.all_sum(v) for k, v in losses.iteritems()}
+  ```
+
+  In Python versions earlier than 3.7, iteration order of dictionaries is not guaranteed.
+  Use an `OrderedDict`:
+  ```
+  losses = collections.OrderedDict([("loss1", ...), ("loss2", ...)])
+  total_losses = {k: ctx.all_sum(v) for k, v in losses.iteritems()}
+  ```
+
+  However, in this particular example, the dictionary keys are sortable. In such cases, this is best:
+  ```
+  losses = {"loss1": ..., "loss2": ...}
+  total_losses = ctx.all_sum(losses)
+  ```
+  “””
+
   @property
   def replica_id(self) -> tf.Tensor:
     "Returns the current replica ID."""
@@ -374,7 +401,7 @@ class ReplicaContext(object):
   def logical_device(self, logical_device_id: int) -> ContextManager:
     """Returns a context to place ops / variables on the given logical device.
 
-    The logical device ids are numbered 0 to N-1 where N is number of logical devices per replica. 
+    The logical device IDs are numbered 0 to N-1 where N is number of logical devices per replica. 
     """
 
   def all_sum(self, value: T) -> T:
@@ -383,7 +410,7 @@ class ReplicaContext(object):
     If `all_sum` is called in any replica, it must be called in all replicas.
     The nested structure and `Tensor` shapes must be identical in all replicas.
 
-    IMPORTANT: The ordering of communications must be identical in all replicas. Otherwise an error will be thrown.
+    IMPORTANT: The ordering of communications must be identical in all replicas. See class docstring.
 
     Example with two replicas:
       Replica 0:: `value`: {'a': 1, 'b': [40,  1]}
@@ -406,7 +433,7 @@ class ReplicaContext(object):
     If `all_min` is called in any replica, it must be called in all replicas.
     The nested structure and `Tensor` shapes must be identical in all replicas.
 
-    IMPORTANT: The ordering of communications must be identical in all replicas. Otherwise an error will be thrown.
+    IMPORTANT: The ordering of communications must be identical in all replicas. See class docstring.
 
     Example with two replicas:
       Replica 0:: `value`: {'a': 1, 'b': [40,  1]}
@@ -420,7 +447,7 @@ class ReplicaContext(object):
         The structure must be compatible with `tf.nest`.
 
     Returns:
-       A `Tensor` nest with the element-wise minimum `value`s from each replica. Otherwise an error will be thrown.
+       A `Tensor` nest with the element-wise minimum `value`s from each replica.
     """
 
   def all_max(self, value: T) -> T:
@@ -429,7 +456,7 @@ class ReplicaContext(object):
     If `all_max` is called in any replica, it must be called in all replicas.
     The nested structure and `Tensor` shapes must be identical in all replicas.
 
-    IMPORTANT: The ordering of communications must be identical in all replicas.
+    IMPORTANT: The ordering of communications must be identical in all replicas. See class docstring.
 
     Example with two replicas:
       Replica 0:: `value`: {'a': 1, 'b': [40,  1]}
@@ -443,7 +470,7 @@ class ReplicaContext(object):
         The structure must be compatible with `tf.nest`.
 
     Returns:
-       A `Tensor` nest with the element-wise maximum `value`s from each replica. Otherwise an error will be thrown.
+       A `Tensor` nest with the element-wise maximum `value`s from each replica.
     """
 
   def all_gather(self, value: T) -> T:
@@ -452,7 +479,7 @@ class ReplicaContext(object):
     If `all_gather` is called in any replica, it must be called in all replicas.
     The nested structure and `Tensor` shapes must be identical in all replicas.
 
-    IMPORTANT: The ordering of communications must be identical in all replicas. Otherwise an error will be thrown.
+    IMPORTANT: The ordering of communications must be identical in all replicas. See class docstring.
 
     Example with two replicas:
       Replica 0:: `value`: {'a': 1, 'b': [40,  1]}
@@ -478,7 +505,7 @@ class ReplicaContext(object):
     The nested structure, `Tensor` shapes, and `source_replica_id` must be
     identical in all replicas.
 
-    IMPORTANT: The ordering of communications must be identical in all replicas. Otherwise an error will be thrown.
+    IMPORTANT: The ordering of communications must be identical in all replicas. See class docstring.
 
     `value` will always be evaluated on all replicas.
 
@@ -499,7 +526,10 @@ class ReplicaContext(object):
     """
 
   def merge_call(merge_fn, *args, **kwargs):
-    """Merge args across replicas and run `merge_fn` in a cross-replica context."""
+    """Merge args across replicas and run `merge_fn` in a cross-replica context.
+
+    IMPORTANT: The ordering of `merge_call`s must be identical in all replicas. See class docstring.
+    """
 ```
 
 #### AggregationType enum
@@ -1021,7 +1051,3 @@ say they're copies of each other, or are sharded.
 *   How would we handle configuring the session options according to the strategy in eager execution?
 *   Can we keep type annotations in code for documentation purposes (but turn off type checking)? 
 *   ([From Karmel’s comment](https://github.com/tensorflow/community/pull/25#discussion_r228659379)) Keras supports DistStrat natively, but we have also discussed the fact that the inclusion of DistStrat in compile/fit limits our ability to handle, say, model parallelism or very large models. Is there an intersection of Keras and DS API that allows for model parallelism/more complicated use semantics without requiring too much from the Keras user?
-
-
-
-
--- a/rfcs/20181016-replicator.md
+++ b/rfcs/20181016-replicator.md
@@ -183,7 +183,7 @@ or more model replicas (on one or more workers). See [InputReplicationMode](#inp
         replica.
     *   `make_input_iterator`: This method is used to create the input iterator that is passed to run. It takes an `input_fn` which either returns a
         `Dataset` or a function which returns a nested structure of input tensors. It also takes an optional `InputReplicationMode`
-        `replication_mode` which specifies how, if at all, the input fn should
+        `input_replication_mode` which specifies how, if at all, the input fn should
         be replicated. The current options are (1) no replication, (2) replicate
         per worker (default option), and (3) replicate per compute-replica. See
         [InputReplicationMode](#inputreplicationmode) section for more details.
@@ -294,7 +294,7 @@ class DistributionStrategy(object):
   def make_input_iterator(
       self,
       input_source: InputFn[T],
-      replication_mode: InputReplicationMode = InputReplicationMode.PER_WORKER) -> InputIterator[T]:
+      input_replication_mode: InputReplicationMode = InputReplicationMode.PER_WORKER) -> InputIterator[T]:
     """Makes an iterator for input provided via an input function.
 
     Depending on the input replication mode, the input function may be called one time
@@ -309,7 +309,7 @@ class DistributionStrategy(object):
         The function must take an `InputContext` as the only parameter. It must return either
         an instance of `tf.data.Dataset` or a parameter-less Python callable that returns a
         `Tensor` nest.
-      replication_mode: The input replication mode. Defaults to per-worker replication.
+      input_replication_mode: The input replication mode. Defaults to per-worker replication.
     """
 
   @overload
@@ -589,21 +589,21 @@ class InputContext(object):
   def input_pipeline_id(self) -> int:
     """Returns the input pipeline ID.
 
-    If `replication_mode` == `SINGLE`, this is always `0`.
-    If `replication_mode` == `PER_WORKER`, this is the worker ID.
-    If `replication_mode` == `PER_REPLICA`, this is the compute replica ID.
+    If `input_replication_mode` == `SINGLE`, this is always `0`.
+    If `input_replication_mode` == `PER_WORKER`, this is the worker ID.
+    If `input_replication_mode` == `PER_REPLICA`, this is the compute replica ID.
     """
 
   @property
   def num_input_pipelines(self) -> int:
     """Returns the number of input pipelines.
 
-    If `replication_mode` == `SINGLE`, this is always `1`.
-    If `replication_mode` == `PER_WORKER`, this is the number of workers.
-    If `replication_mode` == `PER_REPLICA`, this is total number of replicas.
+    If `input_replication_mode` == `SINGLE`, this is always `1`.
+    If `input_replication_mode` == `PER_WORKER`, this is the number of workers.
+    If `input_replication_mode` == `PER_REPLICA`, this is total number of replicas.
     """
 
-  def per_replica_batch_size(self, global_batch_size: int) -> int:
+  def get_per_replica_batch_size(self, global_batch_size: int) -> int:
     """Returns the per replica batch size user should use in their input pipeline, given desired global batch size.
 
     It would be computed as per_replica_batch_size = global_batch_size // num_replicas_in_sync. global_batch_size should be divisible by num_replicas_in_sync. This will throw an error if not.
@@ -746,13 +746,14 @@ with strategy.scope():
   optimizer = tf.train.MomentumOptimizer(learning_rate, 0.9)
 
 def input_fn(ctx):
-  return imagenet.ImageNet(ctx.per_replica_batch_size(effective_batch_size))
+  return imagenet.ImageNet(ctx.get_per_replica_batch_size(effective_batch_size))
 
 def step_fn(inputs):
   image, label = inputs
+
   logits = model(images)
   cross_entropy = tf.nn.sparse_softmax_cross_entropy_with_logits(
-      logits=logits, labels=label)
+      logits=logits, labels=label)  
   loss = tf.reduce_mean(cross_entropy)
   train_op = optimizer.minimize(loss)
   with tf.control_dependencies([train_op]):
@@ -790,7 +791,7 @@ def eval_top1_accuracy(inputs):
   return top1_acc
 
 eval_input_iterator = strategy.make_input_iterator(
-    eval_input_fn, replication_mode=InputReplicationMode.SINGLE)
+    eval_input_fn, input_replication_mode=InputReplicationMode.SINGLE)
 per_replica_top1_accs = strategy.run(eval_top1_accuracy, eval_input_iterator)
 mean_top1_acc = strategy.reduce(per_replica_top1_accs)
 
@@ -828,7 +829,7 @@ def input_fn(ctx):
   d = d.interleave(tf.data.TFRecordDataset, cycle_length=num_readers)
   d = d.map(parser_fn)
 
-  return d.batch(ctx.per_replica_batch_size(effective_batch_size))
+  return d.batch(ctx.get_per_replica_batch_size(effective_batch_size))
 
 ```
 
@@ -843,7 +844,7 @@ def sample_noise(batch_size):
       shape=(batch_size, num_latents), mean=0.0, stddev=1.0)
 
 def input_fn(ctx):
-  batch_size = ctx.per_replica_batch_size(effective_batch_size)
+  batch_size = ctx.get_per_replica_batch_size(effective_batch_size)
   ds = cifar.Cifar10(batch_size)
   return ds.map(lambda x: (x['image'], sample_noise(batch_size)))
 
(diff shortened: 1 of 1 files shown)


rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+    If `broadcast` is called in any replica, it must be called in all replicas.
+    The nested structure, `Tensor` shapes, and `source_replica_id` must be
+    identical in all replicas.
+
+    IMPORTANT: The ordering of communications must be identical in all replicas.
+
+    `value` will always be evaluated on all replicas.
+
+    Example with two replicas:
+      Replica 0:: `value`: {'a': 1, 'b': [40,  1]}, `source_replica_id`: 1
```
alextp: source_replica_id: 0 here?

guptapriya: `source_replica_id` should be the same on all of them actually, it specifies the replica whose value we want to broadcast.

AakashKumarNain:
I don't understand why this thing was proposed in the first place. The whole idea of Tensorflow 2.0 was to have an easy to use interface, eager execution and **`no redundancy`**. However, this RFC contradicts it. It's like saying, we are having `tf.layers.Dense()` for high-level API but we also have `tf.nn.Dense()` if you want more flexibility. IMHO, this is going to create a mess again with two things each of which is able to achieve the same goal. Alternatively, the `hooks` should be provided within the `DistributedStrategy` API itself so that everyone is on the same page. If you are adding another API for researchers, then researchers would be using `Replicator` while applied ML engineers will be using `DistributedStrategy`.

seanpmorgan:
> Distribution Strategy’s low level APIs can be directly used to distribute
training programs written with low level TensorFlow APIs (i.e. without
Estimator/Keras). But it can be tedious and error prone as it exposes an
extensive and flexible API. There is a skeleton of a mid-level API within
Distribution Strategy where users can write their own main-loops more easily, but this has not yet
been developed fully or tested with real users.

IMO this is a pretty weak reason to create a second API. Why not just finish developing the mid-level API and make sure it works across real use cases?

terrykong:
I see that there was discussion of a `num_steps_per_run` parameter, but will the Replicator API support some kind of gradient aggregation? I know we usually like to just throw more accelerators at a problem and to increase the effective batch size, but for some of us who are more restricted in terms of compute power I wonder if we can support this kind of logic to have larger effective batch sizes. For example using N GPUs with batch size B, but aggregating gradients for R number of runs of the optimizer and then delay syncing to achieve an effective batch size of N*B*R.

kazemSafari:
@tfboyd Thank you so much. This will be extremely helpful in my current research projects. I greatly appreciate it.

terrykong:
> > I see that there was discussion of a `num_steps_per_run` parameter, but will the Replicator API support some kind of gradient aggregation? I know we usually like to just throw more accelerators at a problem to increase the effective batch size, but for some of us who are more restricted in terms of compute power I wonder if we can support this kind of logic to have larger effective batch sizes. For example using N GPUs with batch size B, but aggregating gradients for R number of runs of the optimizer and then delay syncing to achieve an effective batch size of N_B_R.
> 
> @terrykong I think that this is an orthogonal problem to the one that Replicator is trying to solve. I believe that it could be achieved using a relatively simple `Optimizer` wrapper, with or without Replicator.

@chr1sj0nes Gotcha, I see your point.

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+replicas operate asynchronously and update the variables independently. We can
+also support a hybrid approach, where the replicas are divided into "sync
+groups" - within each sync group, each replica is in sync, but the sync groups
+are asynchronous among themselves. To support this, we later define
+"num_replicas" as the total number of replicas (sync or async) and
+"num_replicas_in_sync" as the number of replicas in one sync group.
+
+#### Input Pipeline
+
+An instance of the input pipeline (e.g. `Dataset`, `FIFOQueue`) obtained by
```
karmel: Should Queues even be on the table? This would suggest they will be supported.

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+### Replicator API
+
+#### Detailed Design
+
+Here are the main ingredients of the API:
+
+*   `Replicator` class: This is the main class of the API and provides the main
+    methods to replicate the computation.
+    *   `__init__:`Creates a Replicator instance. It takes a
+        `DistributionStrategy` object which is used for (almost) all the
```
karmel: Why qualified with almost? What is not covered by DistStrat?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+    *   `run`: This method is used to replicate the computation. It takes a
+        function `fn` which specifies the computation, and an (optional)
+        `InputIterator` `inputs` to be fed into`fn`. It calls `fn` with one
+        element from the input iterator within distribution strategy `scope`.
+        Variables and optimizers for the model can be created inside or outside
+        `fn`, as long as they're created within `scope`. `run` can be called
+        multiple times with different `fn`. In graph mode, run simply creates
+        the replicated graph whereas in eager mode, it will run the actual
+        computation. It then returns the list of outputs of `fn` from each
+        replica.
```
karmel: A lot of these concepts seem to map to similar concepts in the Estimator world. Can you perhaps add a section explaining the difference between this API/distribution functionality and the existing Estimator API? I am concerned that we are introducing a new highish level API that looks very similar to Estimator and has many of the same usability cliffs, even though in the same breath, we are saying that Estimators are dispreferred versus Keras.

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+    *   `prepare_input`: This method is used to prepare the input for being
+        passed to run. It takes an `InputFn` `fn` which either returns a
+        `Dataset` or a function which returns
+        ([nests](https://www.tensorflow.org/api_docs/python/tf/contrib/framework/nest)
+        of) input tensors. It also takes an optional `InputReplicationMode`
```
karmel: Do we need to support both Dataset and nests of tensors? Can't a dataset of nested tensors suffice for the latter use case, such that this always returns a dataset?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+    `broadcast`, etc.
+*   Input: Replicating and distributing the input correctly across replicas is
+    an important component of replicating the computation. The API provides a
+    number of input related functionality to support different use cases.
+    *   `InputReplicationMode`: The user may want to use the same input pipeline
+        for all their replicas, or have a separate input pipeline per worker or
+        per replica. This mode allows the user to specify which mode they want.
+        The knowledge of this mode helps us create the appropriate input
+        iterator. See [InputReplicationMode](#inputreplicationmode) for more
+        details.
```
karmel: It seems like this is a splitting of the information that should in theory all be contained in DistributionStrategy. Why should we be passing information about how to replicate anywhere outside the DistStrat config?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+    more details.
+
+    If `fn` returns an instance of `tf.data.Dataset`, a corresponding `tf.data.Iterator` will be
+    created. Each replica taking its input from this pipeline will execute `Iterator.get_next()`.
+    The `tf.data.Iterator`(s) can be re-initialized by calling the `reinitialize` on the returned
+    `InputIterator` object.
+
+    If `fn` returns a Python callable (e.g. `FIFOQueue.dequeue`), that callable will be executed on
+    each replica taking its input from this pipeline. `InputIterator.reinitialize` will have no effect on
+    this pipeline.
```
karmel: Returning a callable/Queue seems strange in the 2.0 world. Do we need to support this?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+
+    Args:
+      fn: The input function.
+        The function must take an `InputContext` as the only parameter. It must return either
+        an instance of `tf.data.Dataset` or a parameter-less Python callable that returns a
+        `Tensor` nest.
+      replication_mode: The input replication mode. Defaults to per-worker replication.
+      prefetch_on_device: Indicates whether to prefetch input data to the devices.
+        For now we use a buffer size of 1 for this prefetch.
```
karmel: There are many theoretical dataset optimizations that one in theory would want to run. Why is prefetching an arg here, rather than something the user does in the input_fn with Dataset? Will this have to be expanded if new performance optimization options are added in the future?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+    If `replication_mode` == `PER_WORKER`, this is the number of workers.
+    If `replication_mode` == `PER_REPLICA`, this is `num_replicas`.
+    """
+```
+
+#### InputIterator
+
+An instance of `InputIterator` is returned by `prepare_input` and can be passed
+to `run`.
+
```
karmel: Why is it necessary to have this additional wrapping around the Dataset? Just to handle the non-Dataset case? Or does it serve another purpose?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+```python
+class InputIterator(Generic[T]):
+  """An input iterator, intended to be passed to `Replicator.run`."""
+
+  def get_next(self) -> List[T]:
+    """Returns the next inputs for each replica."""
+
+  def reinitialize(self) -> List[tf.Operation]:
```
karmel: Can you clarify what reinitialization is intended to accomplish? When is it necessary?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+
+### Eager mode support
+
+Eager mode support mostly drops out of building atop of `DistributionStrategy`.
+The entire API is designed to be eager compatible. We've called out the
+difference in behavior in eager and graph mode in the specific API
+documentations. In general, in graph mode, many of the APIs return ops which
+should be later executed with `session.run`, and in eager mode, they simply run
+the ops.
+
```
karmel: Keras supports DistStrat natively, but we have also discussed the fact that the inclusion of DistStrat in compile/fit limits our ability to handle, say, model parallelism or very large models. Is there an intersection of Keras and Replicator that allows for model parallelism/more complicated use semantics without requiring too much from the Keras user?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+#### Training
+
+```python
+with replicator.scope():
+  model = resnet.ResNetV1(resnet.BLOCKS_50)
+  optimizer = tf.train.MomentumOptimizer(learning_rate, 0.9)
+
+def input_fn(ctx):
+  assert effective_batch_size % ctx.num_replicas_in_sync == 0
```
karmel: In practice, if this assert fails, what happens? Or, more broadly, how does error propagation work? Would this stop all replicas, or if one fails, the rest continue?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+  loss = tf.reduce_mean(cross_entropy)
+  train_op = optimizer.minimize(loss)
+  with tf.control_dependencies([train_op]):
+    return tf.identity(loss)
+
+inputs = replicator.prepare_input(input_fn)
+per_replica_losses = replicator.run(step_fn, inputs)
+mean_loss = tf.reduce_mean(tf.stack(per_replica_losses))
+
+with tf.Session() as session:
```
karmel: How does this work in 2.0?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+  session.run(replicator.initialize())
+  for _ in xrange(num_train_steps):
```
karmel: nit: above, we are all py3, but xrange is not in py3.

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+    loss = session.run(mean_loss)
+  session.run(replicator.finalize())
+```
+
+#### Evaluation
+
```
karmel: How does checkpointing/restoring from checkpoint work in this world? Is failure handling left to the user?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+  optimizer = tf.train.RMSPropOptimizer(learning_rate)
+
+# Queues of trajectories from actors.
+queues = []
+def learner_input(ctx):
+  del ctx  # Unused.
+  queue = tf.FIFOQueue(
+      capacity=1, dtypes=trajectory_dtypes, shapes=trajectory_shapes)
+  queues.append(queue)
+
```
karmel: It feels strange to be showing deprecated-in-1.x code as a part of this review for a 2.0 interface.

rfcs/20181016-replicator.md
guptapriya: In the previous proposal there might have been some input handling initially that may have been done in the Replicator layer (likely until we added it to Dist strat). But with the upcoming update, we are incorporating the entire functionality into dist strat.

rfcs/20181016-replicator.md
guptapriya: This is part of the `prepare_input` method because one may want to handle different types of input differently, even when using the same distribution strategy. For instance, you may use the same distribution strategy for training and evaluation but use different input replication mode.
However, I do see the alternative that we require users to use a different strategy in that case. I will add an open question at the end to discuss this as I can see merits for both options.

rfcs/20181016-replicator.md
guptapriya: since this type of error will happen during graph construction, in the in-graph case, this will just mean that the graph will not be constructed and no replicas will start training. (for multi-GPU and TPU strategy, this would be the case)
For the between-graph case (which is typically used for multi worker), there are a couple different modes - standalone client and independent worker mode. We don't talk about these in detail in this design, but there will be an upcoming design talking about how to setup the cluster etc for multi worker cases easily. 
@yuefengz - can you comment on the current state of error propagation for between graph cases for multi worker training?

rfcs/20181016-replicator.md
guptapriya: The Reinforcement Learning example is written in eager style. Basically, calling `run` will run the computation instead of returning the op, so we just call `run` inside a loop.

rfcs/20181016-replicator.md
guptapriya: Will change all to `range` ( in the upcoming revision)

rfcs/20181016-replicator.md
guptapriya: Currently it does server another purpose. It allows us to call the function on different workers and create multiple datasets (for the PER_WORKER mode). So each worker will have their own copy of the dataset. If the user gave us an already created dataset, we have to send the data from one worker to all the other workers, and this is not efficient currently. So we support the PER_WORKER mode and ask this to be a function.
That being said, there is a proposal in the works for a distributed dataset which will alleviate this sort of requirement and allow us to efficiently distribute from a single dataset.

rfcs/20181016-replicator.md
guptapriya: Yes, it's actually the same as calling `initialize` again on an initializable iterator. In the upcoming udpate, we are changing this to be called just `initialize` so it's more clear and matches the regular tf.data iterator methods. An example of when it's useful to initialize the iterator again is during evaluation (See the evaluation example). Since the user is managing the training loop themselves, this is useful for them.

rfcs/20181016-replicator.md
guptapriya: The prefetching option here is `prefetch_on_device` which is directly related to the distribution aspect (as opposed to the regular `prefetch` which the user can still have in their input pipeline as usual). I believe the tf.data plan is to not expose this through tf.data optimizations but as an iterator (`MultiDeviceIterator`). Since in this case, DS is creating the iterator for the user, we need to know which type to create.
All that being said, I am leaning towards removing this option and just making that the default behavior. 
cc @rohan100jain thoughts?

rfcs/20181016-replicator.md
guptapriya: In the current proposal we are not providing any support with checkpointing/restoring and these would be left to the user if they're not using Keras/Estimator. 
There might be a proposal in the future to provide some help with that.

rfcs/20181016-replicator.md
guptapriya: Good point, will add a section. (I think this question still applies even with Replicator being absorbed in the core DS API). You're right that `fn` sounds similar to `model_fn` and `input_fn` sounds similar to `input_fn` in Estimator. The biggest difference is that this doesn't control the training loop whereas with Estimator, once you define, the model_fn, the training loop is run by Estimator and the user has no control over it. Examples such as the GAN and Reinforcement learning example here would be tough to do in Estimator. 
Will add a section.

rfcs/20181016-replicator.md
martinwicke: That is correct, the queues are not deprecated, just QueueRunners.

rfcs/20181016-replicator.md
guptapriya: Yeah this case will fail on all of them. But Karmel's question is also about broad error propagation - if graph creation fails on one of the workers but not on all?

yuefengz: There is not any existing mechanism to broadcast errors. That depends on the outer loop. E.g. Kubeflow can monitor these workers and kill the whole training cluster if any of the worker fails with non-retryable errors.

rfcs/20181016-replicator.md
karmel: I recall, perhaps once upon a time, that there was a restriction that the Estimator input_fn had to return datasets to work with DistStrat, and that returning a tuple of (feature, label) would not be supported. Or maybe that was with replicate_model_fn? Does that restriction still hold, or does the acceptance of nested tensors imply that other return types are also acceptable?

rfcs/20181016-replicator.md
guptapriya: I've removed it from this sentence as you're right, it might convey some special status for queues. I think we just meant that a queue might be used in an input function, but an example later shows this so I think we are good removing it from here.

guptapriya:
We have significantly updated the proposal based on the feedback - the major shift is to incorporate Replicator API into the Distribution Strategy API and re-organize it. Please give it a read and provide your feedback. We would especially welcome your thoughts on the "open questions and discussion topics" at the end of the document. 
The review period has been extended for a week until Nov. 8th.

rfcs/20181016-replicator.md
guptapriya: Great point. I think once we add support for fn returning tensors, we should be able to support `input_fn` returning (feature,label) tuples in Estimator as well.

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,962 @@
...
+| Status        | Proposed                                                |
+| :------------ | :------------------------------------------------------ |
+| **Author(s)** | cjfj@google.com, dominikg@google.com, jhseu@google.com, petebu@google.com, priyag@google.com |
+| **Sponsor**   | joshl@google.com                                        |
+| **Updated**   | 2018-10-17                                              |
+
+## Objective
+
+We propose to add a new API in TensorFlow for replicating computation across
```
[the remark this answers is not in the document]
guptapriya: Thanks! Fixed.

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,1053 @@
...
+The input function will be called for every replica, on the corresponding
+worker, thus creating as many input pipelines as number of replicas. Each
+replica will dequeue from its own `Dataset`.
+
+#### InputContext class
+
+An instance of `InputContext` is passed to the input function.
+
+```python
+class InputContext(object):
```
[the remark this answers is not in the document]
guptapriya: Thanks for the review @wangsiyu . 
For Keras, we have had some discussions around whether we should allow users passing an input function. From the feedback we've got so far, it seems that providing a dataset or numpy array directly, and letting DS figure out how to split/shard the input is the most user friendly approach for Keras APIs. So we are not planning to change Keras API to take an input function. We will instead work on efficient distribution of input from a single dataset. 

For Estimator, this is under discussion right now, since it already supports input function. As you mentioned, it doesn't currently take this input context, but it would be reasonable to modify that to allow it for users who do wish to modify their input pipelines. Most likely we will add this support. 
However, note that this would require users to modify their existing input code. Until now, our philosophy has been to allow keras/estimator users to be able to use their existing code as much as possible when adding DS. So we are trying to figure out what's the best user experience for users who don't wish to modify their existing input pipelines.

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,1053 @@
...
+
+```python
+inputs = tf.keras.layers.Input(shape=(1,))
+predictions = tf.keras.layers.Dense(1)(inputs)
+model = tf.keras.models.Model(inputs=inputs, outputs=predictions)
+dataset = tf.data.Dataset.zip((features, labels))
+strategy = tf.distribute.MirroredStrategy()
+model.compile(loss='mean_squared_error',
+              optimizer=tf.train.GradientDescentOptimizer(learning_rate=0.2),
+              distribute=strategy)
```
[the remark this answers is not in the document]
guptapriya: The plan for input is to rely on datasets to be able to (eventually) efficiently shard input. And this can happen at the compile/fit stage, with whatever dataset is provided at that time. Note that the initial implementation may not be efficient, but we want to design the APIs such that it makes the most sense for users, even if we have to sacrifice performance in the short term. 

AFAIK, there aren't any other limitations we've seen so far because of delaying strategy passing until compile that affect performance, but @anj-s maybe you can add more thoughts on this?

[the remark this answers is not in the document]
anj-s: I believe compile is the right stage to pass the strategy and set it as one of the model properties, similar to loss, optimizer etc.  We also only have access to the input at the fit stage (and possibly information about the workers as well). There isn't any performance issues so far with the current flow. As guptapriya@ mentioned we will be using an initial implementation for input sharding which may have performance issues but this is not dependent on when we pass the strategy object.

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,1053 @@
...
+  queue = queues[actor_id % len(queues)]
+  for _ in range(num_trajectories_per_actor):
+    observation = get_observation_from_environment()
+    action_taken, logits = agent(tf.expand_dims(observation, axis=0))
+    trajectory = Transition(observation, action_taken, logits)
+    queue.enqueue(tf.nest.flatten(trajectory))
+
+# Start the actors.
+for actor_id in range(num_actors):
+  threading.Thread(target=run_actor, args=(actor_id,)).start()
```
jperl: It would be nice to have a reinforcement learning story where the environments run in a `Process` instead of a `Thread` to support heavier computation.

The [agent model call](https://github.com/tensorflow/community/pull/25/files#diff-cb12a9548a2d42da420cb33feb2f7742R932) and [FIFOQueue.enqueue](https://github.com/tensorflow/community/pull/25/files#diff-cb12a9548a2d42da420cb33feb2f7742R934) would be problematic in a `Process`.

jperl: Thanks for the reply @petebu. I apologize if I am misunderstanding -- but can the Distribution Strategy API support mirrored variables on the same machine across processes, so that the end user does not need to manage the cross-process weight transfer manually?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,1053 @@
...
+        element from the input iterator within distribution strategy `scope`.
+        Variables and optimizers for the model can be created inside or outside
+        `fn`, as long as they're created within `scope`. `run` can be called
+        multiple times with different `fn`. In graph mode, run simply creates
+        the replicated graph whereas when eager execution is enabled, it will run the actual
+        computation. It then returns the list of outputs of `fn` from each
+        replica.
+    *   `make_input_iterator`: This method is used to create the input iterator that is passed to run. It takes an `input_fn` which either returns a
+        `Dataset` or a function which returns a nested structure of input tensors. It also takes an optional `InputReplicationMode`
+        `replication_mode` which specifies how, if at all, the input fn should
```
egonina: name this input_replication_mode?

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,1053 @@
...
+"model replica" in some places.
+
+When using model parallelism, a replica will span several physical devices (e.g.
+GPUs, TPU cores) and may also span multiple physical machines.
+
+`replica_id` will range from 0 to N-1 if N is the number of replicas.
+
+#### Worker
+
+The physical machine(s) containing the physical devices (e.g. GPUs, TPUs) on
```
egonina: this definition is a bit confusing to me - is a worker one physical machine or something that runs the replica code on one or more physical machines?

rfcs/20181016-replicator.md
egonina: Similar question to jperl@: When you say "To transfer the weights, you can serialize them out of graph and send them to the other process." does DS support weight transfer between two processes (running on one or multiple machines)? If not, can you outline how DS would be used in combination with other communication libraries to implement distributed RL with distributed data collection and one or more learners (with replay buffers/FIFO queue for experience trajectories)?

rfcs/20181016-replicator.md
guptapriya: A worker, as defined here, is the minimum of (one machine, one replica).
So in the most common case, it would correspond to one machine, and has one or multiple replicas.
But in the less common case where one replica needs to span > 1 machine, we are defining a worker to span the minimum number of machines needed for 1 replica.

I agreed this is a little confusing, and I am also worried this might be confusing with the "worker" task as used in TensorFlow in general (Although they correspond to each other in the common cases). 

Do you have suggestions for a different name, or a different concept we should define here?

rfcs/20181016-replicator.md
guptapriya: will change it, thanks

michaelstjules:
If I have a model that was not trained using a DistributionStrategy, can I just call tf.train.import_meta_graph or tf.import_graph_def in a DS scope to distribute the model? If not, what else would I need to do?

yuefengz:
@michaelstjules This is good question. The `tf.train.import_meta_graph` currently won't work with `run` or `call_for_each_replica` method because it directly adds nodes to a TensorFlow graph but distribution strategy need to capture the creation of variables and ops (to replicate them or place them on right devices). We'll think of some way to support it in some degree.

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,1014 @@
+# Distribution Strategy - Revised API
+
+| Status        | Proposed                                                                  |
```
ewilderj: @guptapriya please amend to "Accepted" and then we can merge this, thank you!

guptapriya: Great, just did that. Thanks @ewilderj !

rfcs/20181016-replicator.md
```diff
@@ -0,0 +1,1053 @@
...
+```python
+inputs = tf.keras.layers.Input(shape=(1,))
+predictions = tf.keras.layers.Dense(1)(inputs)
+model = tf.keras.models.Model(inputs=inputs, outputs=predictions)
+dataset = tf.data.Dataset.zip((features, labels))
+strategy = tf.distribute.MirroredStrategy()
+model.compile(loss='mean_squared_error',
+              optimizer=tf.train.GradientDescentOptimizer(learning_rate=0.2),
+              distribute=strategy)
+model.fit(dataset, epochs=5, steps_per_epoch=10)
```
sshrdp: The interface between the model.fit and Dataset API is not clear. Is the input dataset supposed to be a single-epoch or should it be repeated? If the dataset is just returning the input dataset and fit is controlling number of epochs to iterate over it, why is steps_per_epoch necessary?

guptapriya: That is good feedback @sshrdp and it applies to model.fit and Dataset API in general. In fact @sb2nov had been looking at cleaning this up to make it more intuitive and usable (along the lines of not needing `steps_per_epoch` for dataset case).
@sb2nov - can you shed more light?

rfcs/20181016-replicator.md
mjlbach: Is there any update on this? Currently trying to do this (have data gather loop in its own process) and have been using tf.train.replica_device_setter() building the graph in multiple processes and using the parameter server to share weights between agent collecting data and updater that updates the agent's policy model.

capilano:
Is this a work in progress or does the distribute strategy now support a clean way to efficiently implement custom training loops? I have seen examples of custom loops in tensorflow 1.14, but inside a session, is it possible to use tf.function in 1.14 (without any slowdown) or should we have to move to 2.0 to use the distribute strategy for custom training in eager mode as intended

guptapriya:
@capilano Most of the APIs mentioned in this RFC have been implemented. You can see tutorials here: 
2.0 style: https://www.tensorflow.org/beta/tutorials/distribute/training_loops
1.x style: https://www.tensorflow.org/tutorials/distribute/training_loops

You should be able to use tf.function in 1.x when using eager mode in the same way as the 2.0 tutorial above describes. You may need to use a more recent version of TF 1.x than 1.14 though, as some of the APIs have been added later. You can find them in TF 1.x nightly release.

capilano:
@guptapriya I have looked at them, and when I tried implementing my model using TPU distribute strategy about a month back in 1.14(latest stable release),there were errors while distributing the dataset (in eager mode) and also it was slower than using an estimator.
As a side note, the 1.x style example (TPU custom training part) does not mention anything about how the losses are reduced over multiple replicas, where as the GPU mirrored strategy manually sums over all replicas. Does the TPU strategy behave differently here? And how about batch_norm mean and variances(are they per replica or combined?)
Also, is there support for non Keras models,let's say if they were built using tf.nn.layers? I am trying to port a GAN model that I am currently training using estimators without much success.
I even tried directly converting the DCGAN Keras example https://www.tensorflow.org/beta/tutorials/generative/dcgan by putting everything that needed to be put under strategy scope,but it throws an error in the current stable tensorflow build as of today. I'll take your suggestion and try nightly or 2.0 beta. Thanks

PR #34, opened 2018-11-16
RFC: SavedModel Save/Load in 2.x

allenlavoie:
**Review period closes 2018-12-04**
## SavedModel Save/Load in 2.x

| Status        | Accepted       |
| **Author(s)** | Allen Lavoie (allenl@google.com), André Susano Pinto (andresp@google.com), Arno Eigenwillig (arnoegw@google.com), Rohan Jain (rohanj@google.com) |
| **Sponsor**   | Karmel Allison (karmel@google.com) |
| **Updated**   | 2019-02-28                                           |

Objective: provide an API for serialization/deserialization in TF-2.0 that supports both serving and reuse use-cases.

### DIFF
--- a/rfcs/20181116-saved-model.md
+++ b/rfcs/20181116-saved-model.md
@@ -1,6 +1,6 @@
 ﻿# SavedModel Save/Load in 2.x
 
-| Status        | Proposed       |
+| Status        | Accepted       |
 :-------------- |:---------------------------------------------------- |
 | **Author(s)** | Allen Lavoie (allenl@google.com), André Susano Pinto (andresp@google.com), Arno Eigenwillig (arnoegw@google.com), Rohan Jain (rohanj@google.com) |
 | **Sponsor**   | Karmel Allison (karmel@google.com) |
--- a/rfcs/20181116-saved-model.md
+++ b/rfcs/20181116-saved-model.md
@@ -1,0 +1,744 @@
+﻿# SavedModel Save/Load in 2.x
+
+| Status        | Proposed       |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | Allen Lavoie (allenl@google.com), André Susano Pinto (andresp@google.com), Arno Eigenwillig (arnoegw@google.com), Rohan Jain (rohanj@google.com) |
+| **Sponsor**   | Karmel Allison (karmel@google.com) |
+| **Updated**   | 2018-11-15                                           |
+
+
+## Objective
+
+Provide an API for serialization/deserialization in TF-2.0 that supports both serving and reuse use-cases.
+
+TensorFlow 2 will include neither `tf.Session` nor `tf.train.Saver` as public symbols. Current SavedModel export and import workflows rely heavily on these symbols. This document proposes adding `tf.saved_model.save` and `tf.saved_model.load` as 2.x-compatible replacements. These symbols are mentioned at a high level in [the Functions, not Sessions RFC](https://github.com/tensorflow/community/pull/20).
+
+## Motivation and introduction
+
+
+### Serialization use-cases
+
+There are several reasons to serialize state in programs which use TensorFlow, each with different requirements. This proposal addresses serving and sharing.
+
+
+#### Training
+
+When writing a new model, the first need users run into is a way to continue training after restarting a program. The original code still exists, although it may be modified slightly.
+
+[tf.train.Checkpoint](https://www.tensorflow.org/api_docs/python/tf/train/Checkpoint) handles this case. TensorFlow state referenced by Python objects is recorded and can be restored later in time. In order to be used the user needs to have access to the code that constructs the objects.
+
+
+#### Serving
+
+Once trained, users may want to serve requests using their model. Ideally for this use-case the representation should be a hermetic and stateless black box, usable through a stable interface, and with minimal dependencies.
+
+SavedModel as a format satisfies this use-case. Various APIs for creating SavedModels exist in TensorFlow 1.x (SavedModelBuilder, Estimator, others). Generally the only interface to the model needed for this use-case is a signature specifying what goes into the model and what comes out. This use-case is well supported in TF 1.x and will continue to be supported under this proposal.
+
+
+#### Sharing
+
+Beyond serving, users may want to reuse all of parts of a trained model. SavedModel allows saving the specific computation together with its pre-trained weights, without depending on the model definition in Python and its particular framework. This helps reproducibility (say, for results reported in a paper) and reuse (say, for importing a pre-trained embedding into a model that uses it on a new task). While serving and reproducibility call for a complete model, reuse typically concerns part of a trained model and its composition. That means loading the SavedModel must restore enough Python state to allow building on top of it.
+
+These use-cases are not well addressed in core TensorFlow 1.x APIs. Export APIs have been complicated by concerns relevant to the serving use-case, and not much time has been spent on usability for re-importing models into Python. TensorFlow Hub has solved this (esp. the import workflow) for sharing graphs, but this needs significant redesign for TF 2.x in light of [the Functions, not Sessions RFC](https://github.com/tensorflow/community/pull/20).
+
+
+### Goals
+
+
+
+*   Support use-cases covered by existing SavedModel export/import APIs and existing Hub module APIs: serving and sharing
+    *   Export a smaller part of a full model.
+    *   Import a part into a larger model.
+    *   Fine-tune an imported part (requires defining "trainable" set of variables, running update ops for batch normalization, exposing regularization losses).
+    *   Re-export an imported part (including its new fine-tuned state).
+    *   Programmatically inspect signatures/functions
+    *   Import state once but data flow using that state multiple times
+*   Export to SavedModel in a way that is compatible with existing serving infrastructure. Extensions to the format may be required for 2.x compatibility, but existing loading procedures will continue to work.
+*   Reimport TensorFlow functionality of exported objects back into Python
+    *   With minimal special casing for Keras types, support a SavedModel implementation of `tf.keras.Model.save` and `tf.keras.model.load_model`. Details of this special casing will be left to another document.
+    *   Exporting and importing should be idempotent (reimported representations are saveable)
+*   Importing existing SavedModels into Python in TensorFlow 2.x. This will share the `tf.saved_model.load` symbol, but will lack any object structure.
+
+
+### Non-goals
+
+
+
+*   Export arbitrary SavedModels. Use-cases will be covered. For example SavedModel supports multiple MetaGraphs, but the APIs proposed here may only ever export SavedModels with a single MetaGraph.
+*   Usable Python interfaces for all symbols in the TensorFlow API on import. Everything will of course be usable when graph building, but objects may not have many features when imported back into Python. The set of types with "nice" import representations is expected to increase over time.
+
+
+### Terminology
+
+
+
+*   A _signature_ identifies inputs and outputs for a computation, roughly the feeds and fetches of a single `session.run` call in TensorFlow 1.x.
+    *   In a SavedModel or MetaGraph, SignatureDefs identify input and output tensors in the GraphDef, possibly overlapping.
+*   _Concrete function graph_: A subgraph with a single signature which TensorFlow can execute natively to compute Tensor outputs from Tensor inputs.
+*   _Polymorphic function_: A Python callable that encapsulates several concrete function graphs behind one, more natural API.  For example it may use non-Tensor arguments to dispatch between the concrete graphs that compute outputs from the Tensor arguments.
+
+
+*   _FunctionDef:_ A protocol buffer representing a concrete TensorFlow function.
+*   _FunctionDefLibrary_: A protocol buffer containing multiple FunctionDefs and their gradient functions.
+*   _GraphDef_: A serialized "v1-style" TensorFlow graph. Includes a FunctionDefLibrary.
+*   _MetaGraph:_ A GraphDef + training checkpoint. Contains fields for signatures and other metadata, although these are often blank.
+*   [SavedModel: A collection of MetaGraphs with additional assets](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/saved_model/README.md). Signatures and other MetaGraph fields are always filled.
+
+
+*   _Checkpointable_: A mixin which manages dependencies between objects. Checkpointable objects have named dependencies on other checkpointable objects. Most TensorFlow objects already participate in this scheme.
+*   _tf.train.Checkpoint_: A checkpointable object with a save() method, used for writing training checkpoints. Used in several examples in this document as an easy way to make a checkpointable object, since there are no plans for a separate "Checkpointable" symbol.
+
+
+*   _Resource_: A data type (dtype) for tensors which point to state. Used to implement variables and tables. In TensorFlow 2.x, resources persist as eager tensors attached to a single Python object (e.g. a `tf.Variable` object), sharing a lifetime with that object. Operations take resource-dtype tensors in order to read or mutate the pointed-to state. The [RFC for variables in 2.x](https://github.com/tensorflow/community/blob/master/rfcs/20180817-variables-20.md) has more detail.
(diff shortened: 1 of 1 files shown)


ppwwyyxx:
There is another use-case which I'm not sure has been covered by this proposal: make large code modifications/refactor and restore the weights/states only (but not restore the computation).
For example, originally I have a network like this:
```python
class Net():
    def __init__(self):
        self.W1 = tf.Variable(...)
        self.b1 = tf.Variable(...)
        self.W2 = tf.Variable(...)
        self.b2 = tf.Variable(...)

    @tf.function
    def __call__(self, x):
        return self.W2 * (self.W1 * x + self.b1) + self.b2
```

And after a while I want to refactor it into:
```python
class Net():
    def __init__(self):
        self.L1 = Layer()
        self.L2 = Layer()

    @tf.function
    def __call__(self, x):
        return self.L2(self.L1(x))


class Layer():
    def __init__(self):
        self.W = tf.Variable(...)
        self.b = tf.Variable(...)

    @tf.function
    def __call__(self, x):
        return self.W * x + b
```

This is easily supported by `tf.train.Saver` in TF 1, as long as the variable names are unchanged between the two versions above. I'm not sure how this case would be handled with the proposed API, because it appears to me that saving a `Checkpointable` always save both the variables and the computation. And where a variable is loaded is strongly associated to how the code is originally written (i.e. the attribute names, etc).

allenlavoie:
@ppwwyyxx I do [mention the training checkpoint use-case in the introduction](https://github.com/tensorflow/community/blob/41c57c6959742867ac88b31938f7cb37fc6929a3/rfcs/20181116-saved-model.md#training). After that it's not mentioned much mostly because it's already implemented.

[`tf.train.Checkpoint`](https://www.tensorflow.org/api_docs/python/tf/train/Checkpoint) does not save computation, and like `Saver` in v1 is aimed at training checkpoints. It does use attribute names in objects rather than scope names, but should generally be a bit more flexible because of it.

ppwwyyxx:
@allenlavoie I'm aware of the paragraph, but "using attribute names in objects" is exactly what I'm concerned about. In the code example I gave above, due to refactoring, the object hierarchy has changed and therefore the attributes are different. How should users restore an old model after the code has been changed like the way above?

allenlavoie:
@ppwwyyxx one option with object-based checkpointing is to construct a checkpoint-compatible structure and a new structure which both reference the same variable objects, then load the checkpoint into the checkpoint compatible structure. Which is roughly equivalent to passing a dictionary of names to `Saver` after a large-scale refactor. (There is an equivalent refactoring issue when name-based code which opens scopes has to be restructured.)

I wrote a design document for `tf.train.Checkpoint` quite a while back. If it's helpful for background I can dig it up and post it.

ppwwyyxx:
Thanks! From this solution, it sounds to me that, in order to load an old checkpoint, the legacy code has to be kept after refactoring, right? This would be against the goal of refactoring. 

And users will still need the extra logic to "call old `Net()` (and then share with new `Net`) if loading old models, and call new `Net()` if loading new models" -- as a result, this solution is no better than nothing: users are still just going through the old codepath to load old models.

I agree to some extent that this is roughly like passing a dict of names to `Saver`, like you said. But keeping a legacy mapping of strings is much easier than keeping an entire legacy implementation of the model. (Think about how to write a script to convert the checkpoint: with `Saver` it only needs to translate the names, with `Checkpointable` it needs to have access to the model implementation).

Another complexity compared to `Saver`, is that with `Saver`, __code refactoring does not usually result in incompatibilities at all__. For example, in the two versions of `Net()` I gave above, refactoring from the 2nd `Net` to the 1st `Net` can be done with no changes to variable names. 

Because of this, I would disagree that "using attribute names is more flexible than using scope names",  since attribute names are more closely tied to code structure. In fact, using attribute names to save models is a major headache many others and I have had with PyTorch.

allenlavoie:
> Thanks! From this solution, it sounds to me that, in order to load an old checkpoint, the legacy code has to be kept after refactoring, right? This would be against the goal of refactoring.
> 
> And users will still need the extra logic to "call old Net() (and then share with new Net) if loading old models, and call new Net() if loading new models" -- as a result, this solution is no better than nothing: users are still just going through the old codepath to load old models.

It doesn't require a legacy implementation of the model, it just requires the same structure. This can be constructed of dummy `tf.train.Checkpoint` objects which just have the same attributes as before. I don't see why that would be any more difficult than names. And once you get down to Layers or submodules which are compatible, those can be mentioned as a whole. That's much nicer than having to go variable-by-variable in a `Saver` dictionary.

> Another complexity compared to Saver, is that with Saver, code refactoring does not usually result in incompatibilities at all. For example, in the two versions of Net() I gave above, refactoring from the 2nd Net to the 1st Net can be done with no changes to variable names.

I've broken many models by refactoring, so I have to disagree with you here. Scopes are pretty notorious for being fragile.

If naming each variable explicitly is an option, it's easy with the object-based API to save a dictionary of variables and nothing else. Mostly we needed to get rid of collections and `variable_scope`; I don't think we took away much freedom in terms of using names to save/restore.

> Because of this, I would disagree that "using attribute names is more flexible than using scope names", since attribute names are more closely tied to code structure. In fact, using attribute names to save models is a major headache many others and I have had with PyTorch.

I'm curious, a headache compared to what? There's almost a one-to-one mapping between object relationships and the scope-based naming that was common in v1. I agree that refactoring can still break checkpoints with the object-based API, but I think our main disagreement is over whether name-based APIs have any advantages there.

ppwwyyxx:
> It doesn't require a legacy implementation of the model, it just requires the same structure. This can be constructed of dummy tf.train.Checkpoint objects which just have the same attributes as before.

Thanks for the clarification! Now it makes more sense. Using a dummy Checkpoint object with only the attributes would be quite easy. 

If I'm not mistaken, to load an old model in the examples I gave above, I'll need to:
```python
new = NewNet()
old = DummyEmptyCheckpoint()
old.W1 = new.L1.W
old.W2 = new.L2.W
old.b1 = new.L1.b
old.b2 = new.L2.b
# or perhaps 
# old = DummyEmptyCheckpoint(W1=new.L1.W, W2=new.L2.W, b1=new.L1.b, b2=new.L2.b)
old.restore('checkpoint')
```
which is awesome and has the same complexity as `saver = tf.train.Saver({'W1': 'L1/W', 'W2': 'L2/W'... }); saver.restore('checkpoint')`.

> I'm curious, a headache compared to what? There's almost a one-to-one mapping between object relationships and the scope-based naming that was common in v1. I agree that refactoring can still break checkpoints with the object-based API, but I think our main disagreement is over whether name-based APIs have any advantages there

Compared to scope-based APIs, and mainly because code refactoring does not usually result in incompatibilities with scope-based API. I've given one example above. Another example that's more common:
```python
class OldNet():
    def __init__(self):
        self.W1 = tf.Variable(...)
        self.b1 = tf.Variable(...)
        self.W2 = tf.Variable(...)
        self.b2 = tf.Variable(...)

    def __call__(self, x):
        return tf.nn.conv2d(self.W2, self.W1 * x + self.b1) + self.b2

class NewNet():
    def __init__(self):
        self.L1 = MyFC()
        self.L2 = MyConv()

    def __call__(self, x):
        return self.L2(self.L1(x))
```
In this refactoring, variable names are unchanged as long as we don't enter a new scope in `MyFC()`.
This sort of refactoring is common, because often we just split things into submodules when the big module (the `OldNet`) has more and more functionalities and becomes crowded. However with attribute-based checkpointing, such refactoring will have to introduce incompatibilities, right?

martinwicke:
@ppwwyyxx I've seen a lot of code for which this is not true, and the failure is extremely vexing. Consider this:

ppwwyyxx:
Thanks! Very interesting example!

allenlavoie:
@alanhdu Interesting. Those do seem like reasonable things to support somehow. I haven't given it too much thought yet, but one option is to do automatic (limited) serialization for object attributes which contain only basic types like nests of list/tuple/dict/int/string/float. We already need to support saving these types for Python-only function arguments (used for dispatch, e.g. `training=True` vs. `training=False`), and we already need to inspect object attributes, so it wouldn't be a big stretch over the current proposal.

zakizhou:
@allenlavoie After an instance of a subclass of `tf.keras.Model` is serialized, will the loaded object be of type `tf.keras.Model`  or just `CheckpointableBase`?

allenlavoie:
@zakizhou generic `Checkpointable` to start, `Model` eventually. The hope is that we can then swap in a SavedModel implementation of `tf.keras.Model.save` and `load_model`.

alanhdu:
Hi! I don't know if this is the write thread to post this in, but I've been playing around with the TF 2.0 APIs and have run into a problem with the saving and loading `tf.Module`s.

Our actual use-case is to deploy RNNs working on streams of data. At inference time, we need to (1) be able to change the batch size between calls and (2) save the output state from one call to feed into the input state to another.

In TF 1.0, we handled this by explicitly passing in our input states and explicitly capturing the output states using a simple Python dictionary mapping  `input state placeholder`s to `output state tensor`s and stuffing some extra data into `assets.extra` at export time). I've tried a two different ideas in TF 2.0, but ran into problems with both: 

1. Allow each `tf.Module` to manage its own state (something Keras's `stateful` flag). The problem here is that I can't figure out how to allow a dynamically sized variable (i.e. the batch size of the state needs to be dynamic, because it might change between calls to `reset_state`).
2. Explicitly pass in the model stateless and explicitly pass in the input state and retrieve the output state. The problem here is that the most natural Python-API would use dictionaries and/or lists to pass in input states, which would make those explicitly unavailable to serving API:

```
 The `Tensor` arguments are the only ones serving APIs will know about, and these may not be nested inside lists, tuples, or dictionaries (since there is no way to specify nested `Tensor` arguments when serving).
```

Is it possible to either loosen this restriction, or to allow `tf.function`s to declare dependencies on dynamically sized variables?

allenlavoie:
@alanhdu

Dynamic variables should still work as long as the variable is created inside a `tf.function`:

```
import tensorflow as tf

class HasDynamicVariable(tf.Module):

  def __init__(self):
    super(HasDynamicVariable, self).__init__()
    self.v = None

  @tf.function(input_signature=[tf.TensorSpec([None, 2], tf.float32)])
  def make_dynamic_variable(self, initial_value):
    if self.v is None:
      self.v = tf.Variable(initial_value)
    return self.v

has_dynamic = HasDynamicVariable()
has_dynamic.make_dynamic_variable(tf.ones([10, 2]))
assert 20. == tf.reduce_sum(has_dynamic.v).numpy()
print(has_dynamic.v.shape)  # None, 2
has_dynamic.v.assign(tf.ones([15, 2]))
assert 30. == tf.reduce_sum(has_dynamic.v).numpy()
```
(Works for me running 2.0.0-dev20190226 on Python 3.6)

But getting TensorFlow Serving (and TensorFlow's C++ loader API) to allow nested structures is a good idea too. One thought (I forget whose) is that signatures could go away completely at some point (TF 4.x? We maintain SavedModel compatibility for at least one extra major version, and are exporting/consuming in 2.x.). Then Serving would support the same nested structures as tf.saved_model.load. Plenty of details to work out (how do you specify those in a request?) but I think it's workable. And we certainly don't need to wait for signatures to go away before allowing Serving to access arbitrary tf.saved_model.save functions.

alanhdu:
@allenlavoie Ah -- by "dynamically sized" I meant something like:

```python
class DyanmicSized(tf.Module):
    @tf.function
    def reset_state(self, batch_size):
        self.batch_size = tf.Variables(tf.zeros(batch_size, 32))
```

(that is, a `tf.Variable` with the `None` dynamic shape (`tf.placeholder(tf.float32, [None, 32])`), which I don't think is possible. That would mean for the LSTM case, we could do something like:

```python
lstm.reset_states(batch_size=64)
lstm(batch1)    # training
lstm.reset_states(batch_size=8)
lstm(batch2)    # inference
```

> But getting TensorFlow Serving (and TensorFlow's C++ loader API) to allow nested structures is a good idea too.

:+1: to that, although I think all our particular use case needs is to somehow map the Python nested structure to TF names, so we know what to feed into the C++ API.

allenlavoie:
That's what the snippet I posted does. The function just needs some indication of what the shapes should be (`input_signature`), otherwise it defaults to the most specific input shape.

alanhdu:
@allenlavoie Ah... I see. So you can get dynamic shapes by setting it via `tf.function(input_signature=...)` appropriately. I was thinking that there was a way to do `tf.Variable(..., shape=[None, ...])` or something (in the example I had, you'll notice that we pass in a `batch_size` parameter which is an integer scalar, but want the `tf.Variable` to have shape `[None, 32]`), since that intuitively seems the easiest to use.

(I should also say that for me, on Python 3.7 and `tf-nightly-2.0-preview==2.0.0.dev20190227`, I get the following warning):
```python
WARNING: Logging before flag parsing goes to stderr.
W0227 12:38:06.226005 140243216058176 tf_logging.py:161] Entity <method-wrapper '__call__' of weakref object at 0x7f8ce614c4f8> could not be transformed and will be staged without change. Error details can be found in the logs when running with the env variable AUTOGRAPH_VERBOSITY >= 1. Please report this to the AutoGraph team. Cause: Object conversion is not yet supported. If you are trying to convert code that uses an existing object, try including the creation of that object in the conversion. For example, instead of converting the method of a class, try converting the entire class instead. See https://github.com/tensorflow/tensorflow/blob/master/tensorflow/contrib/autograph/README.md#using-the-functional-api for more information.
```

That said, if this works then I think this should work for our use case (although we still might run into specifying `tf.TensorShape`s for tensors in nested inputs as the most "natural" programming model).

yueqiw:
@allenlavoie After saving a model in Tensorflow 2.0 using `tf.saved_model.save()`, how to load it back?

Below is how I created the model
```
...
model = tf.keras.Model(inputs, outputs)
... # train the model
print(model)
```
`<tensorflow.python.keras.engine.training.Model object at 0x7fa4f7e9df28>`

Now I want to save the model and load it back. However, the loaded object is no longer a `Model()` object and I cannot figure out a way to use it for inference. 
```
tf.saved_model.save(model, save_path)
imported = tf.saved_model.load(save_path)
print(imported)
``` 
`<tensorflow.python.saved_model.load._Loader._recreate_user_object.<locals>._UserObject object at 0x7fa49a073cc0>`

```
exported = tf.train.Checkpoint(model = model)
tf.saved_model.save(exported, save_path)
imported = tf.saved_model.load(save_path)
print(imported.model)
```
`<tensorflow.python.saved_model.load._Loader._recreate_user_object.<locals>._UserObject object at 0x7fa243812080>`

What's the correct way to use `tf.saved_model` ? Thanks!

allenlavoie:
@yueqiw

Yep, we won't have re-wrapping into Keras types implemented for the alpha. But if you're exporting a Model, the forward pass is available as a signature. `imported.signatures['serving_default']` will have a function.

I have a guide to 2.x SavedModel pending review.

yueqiw:
@allenlavoie Thanks, it works!

allenlavoie:
@goldiegadde added review notes. Does it need anything else?

zakizhou:
Hi @allenlavoie , will `tf.Module` be the base class for `tf.keras.layers.Layer` in the release of TF 2.0?

allenlavoie:
@zakizhou I believe that's still the plan.

rfcs/20181116-saved-model.md
```diff
@@ -0,0 +1,816 @@
+﻿# SavedModel Save/Load in 2.x
+
+| Status        | Proposed       |
```
goldiegadde: Can you please change the status to accepted, and then I will merge it.

goldiegadde:
> @goldiegadde added review notes. Does it need anything else?
Sorry , missed this. added a comment.

rfcs/20181116-saved-model.md
allenlavoie: Done, thanks.

zh794390558:
So, I can retrain, serving with saved model in the future? And how can i use the saved model with different device and distribution strategy? More detailed use case will be preferred.

allenlavoie:
@zh794390558 yes on the first question; see [the 2.x guide to SavedModel](https://www.tensorflow.org/alpha/guide/saved_model).

There isn't yet any integration between distribution strategies and SavedModels. Happy to give pointers if you or someone else is interested in starting on that; it's on my list, but that's a very long list right now.

allenlavoie:
Gentle ping @goldiegadde . I changed the status. Is this missing anything else?

PR #41, opened 2018-12-19
Add slack workspace for sig/io

yupbank:
https://tf-io.slack.com

cc @yongtang


ewilderj:
Hi folks. Our pattern to date has been to create Gitter channels for SIGs. We have one for [Addons](https://gitter.im/tensorflow/sig-addons), [Build](https://gitter.im/tensorflow/sig-build). I would be very happy to add one for SIG IO! We prefer this as we've already got some traction there, and it makes sense to cluster them together vs an entire Slack install just for one SIG. (It also means you can log in with GIthub which makes communication more straightforward.)

So, should I add a Gitter for sig-io?

terrytangyuan:
@ewilderj +1 to consistent entrypoint to access the SIGs. It’s much easier to manage. You made a good point on login via Github for Gitter which seems more open source friendly.

yupbank:
true... thanks for the very good points, yeah,  let's setup a Gitter channel...

PR #51, opened 2019-01-11
Small typo

gdequeiroz:


ewilderj:
cc @dynamicwebpaige to resolve the duplicate file thing. I'd like the final doc in sig build if possible vs governance. Now Martin's email has gone out and flown by, and the new testing group is in place, can we move it yet?

PR #54, opened 2019-01-15
RFC: Attention for Dense Networks on Keras

roumposg:
**The feedback phase will be open for 2 weeks until 2019-01-30**

# Attention for Dense Networks on Keras

| Status        | Proposed                                                                   |
:-------------- |:-------------------------------------------------------------------------- |
| **Author(s)** | Georgios Roumpos (roumposg@google.com)                                     |
| **Sponsors**  | Karmel Allison (karmel@google.com), Francois Chollet (fchollet@google.com) |
| **Updated**   | 2019-01-15                                                                 |

# Summary

This RFC proposes adding a layer for attention in `tf.keras.layers` that works with CNN/DNN networks. 
Recently people have had success using the Attention mechanism in dense layers,
e.g. CNN+Attention or Transformer networks. `tf.keras.layers` is the recommended way to build models in tensorflow, but it does not have a layer for attention that works with CNN/DNN networks. 

Note that this review focuses on Dense networks, namely CNN/DNN. It does not cover Recurrent Neural Networks (RNN).

### DIFF
--- a/rfcs/20190115-dense-attention.md
+++ b/rfcs/20190115-dense-attention.md
@@ -42,8 +42,9 @@ affects the output at T+1.
 and
 [tf.contrib.seq2seq.BahdanauAttention](https://www.tensorflow.org/api_docs/python/tf/contrib/seq2seq/BahdanauAttention)
 are implementations of dot-product (Luong) and additive (Bahdanau) Attention
-respectively for RNN in Tensorflow. This proposal is based on this
-implementation, but works with CNN/Dense networks.
+respectively for RNN in Tensorflow. There is ongoing work to implement those
+as Keras layers. Our proposal will follow the same implementation details,
+namely same mathematical operations, but will work with CNN/Dense networks.
 
 There is an implementation of Attention as a `tf.layers.Layer` subclass under
 https://github.com/tensorflow/models/tree/master/official/transformer,
--- a/rfcs/20190115-dense-attention.md
+++ b/rfcs/20190115-dense-attention.md
@@ -1,0 +1,506 @@
+# Attention for Dense networks on Keras
+
+| Status        | Proposed                                                                   |
+:-------------- |:-------------------------------------------------------------------------- |
+| **Author(s)** | Georgios Roumpos (roumposg@google.com)                                     |
+| **Sponsors**  | Karmel Allison (karmel@google.com), Francois Chollet (fchollet@google.com) |
+| **Updated**   | 2019-01-15                                                                 |
+
+## Objective and Motivation
+
+Recently people have had success using the Attention mechanism in dense layers,
+e.g. CNN+Attention or Transformer networks. Some examples are the
+["Attention is all you need"](https://arxiv.org/abs/1706.03762) paper, and
+models in
+[semantic text similarity](https://ai.googleblog.com/2018/05/advances-in-semantic-textual-similarity.html).
+`tf.keras.layers` is the recommended way to build models in tensorflow, but it
+does not have a layer for attention that works with CNN/DNN networks. We would
+like to contribute this capability.
+
+Keras is an API spec that can be implemented across different languages and
+backends, and `tf.keras` is a particular implementation of that spec. This
+document contains code examples for `tensorflow`, but the same API should work
+everywhere.
+
+### Recurrent Neural Networks
+
+Although not the primary focus of this proposal, the Attention layers in this
+proposal work with some configurations of RNN networks.
+Namely, when users create
+[tf.keras.layers.LSTM](https://www.tensorflow.org/api_docs/python/tf/keras/layers/LSTM)
+with `return_sequences=True`, the rest works the same way as CNN.
+
+Unfortunately, this technique does not cover sequence-to-sequence RNN models.
+In these models, the value is the states of encoder, and the query is the input
+of the decoder. The decoder needs to slide its input based on the timesteps, and
+feed them one by one. So, the output of the attention layer at timestep T
+affects the output at T+1.
+
+## Previous Work
+
+[tf.contrib.seq2seq.LuongAttention](https://www.tensorflow.org/api_docs/python/tf/contrib/seq2seq/LuongAttention)
+and
+[tf.contrib.seq2seq.BahdanauAttention](https://www.tensorflow.org/api_docs/python/tf/contrib/seq2seq/BahdanauAttention)
+are implementations of dot-product (Luong) and additive (Bahdanau) Attention
+respectively for RNN in Tensorflow. This proposal is based on this
+implementation, but works with CNN/Dense networks.
+
+There is an implementation of Attention as a `tf.layers.Layer` subclass under
+https://github.com/tensorflow/models/tree/master/official/transformer,
+specifically in
+https://github.com/tensorflow/models/blob/master/official/transformer/model/attention_layer.py.
+That file implements dot-product attention proposed in this file, and also
+supports multi-head. Our proposal is to expose such a method in
+`tf.keras.layers`. In addition, our proposal creates variables inside the
+`build()` method, rather than the `call()` method.
+
+There is ongoing work to add Attention in Keras, namely
+https://github.com/keras-team/keras/pull/11421. That proposal addresses
+Attention mechanism for RNN networks only. I cannot see a way to make it work
+for CNN/Dense networks, which are the motivation for our proposal.
+
+https://github.com/keras-team/keras/issues/9263 contains an example of a Keras
+Layer that implements a CNN+Attention network. That example merges CNN and
+Attention into the same class, whereas our proposal is modular. In the
+Examples section, we present an example of how to build a
+CNN+Attention model.
+
+https://github.com/keras-team/keras/issues/7341 is a request to add an Attention
+Layer. Our proposal will resolve that request.
+
+https://github.com/keras-team/keras/issues/7803 is a request for a Multi-Head
+Attention Layer. Multi-Head Attention is not covered in this proposal, but can
+be implemented as a follow-up, as discussed in the
+[Multi-Head Attention](#multi-head-attention) section.
+
+There are a few more issues that request Attention for RNN. They are covered
+either by https://github.com/keras-team/keras/pull/11421 or our proposal:
+
+*   https://github.com/keras-team/keras/issues/5738
+*   https://github.com/keras-team/keras/issues/4962
+*   https://github.com/keras-team/keras/issues/2525
+
+## Design Proposal
+
+We propose to implement the following common attention layers:
+
+*   `Attention`: Basic dot-product attention, a.k.a. Luong-style attention.
+    Follows
+    [tf.contrib.seq2seq.LuongAttention](https://www.tensorflow.org/api_docs/python/tf/contrib/seq2seq/LuongAttention).
+    This attention has two forms.
+    *   The first is standard dot-product attention, as described in: Minh-Thang
+        Luong, Hieu Pham, Christopher D. Manning. "Effective Approaches to
+        Attention-based Neural Machine Translation." EMNLP 2015.
+        https://arxiv.org/abs/1508.04025.
+    *   The second is the scaled form inspired partly by the normalized form of
+        additive (Bahdanau-style) attention. To enable the second form,
+        construct the object with parameter `scale=True`.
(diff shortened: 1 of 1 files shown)


rfcs/20190115-dense-attention.md
```diff
@@ -0,0 +1,506 @@
...
+document contains code examples for `tensorflow`, but the same API should work
+everywhere.
+
+### Recurrent Neural Networks
+
+Although not the primary focus of this proposal, the Attention layers in this
+proposal work with some configurations of RNN networks.
+Namely, when users create
+[tf.keras.layers.LSTM](https://www.tensorflow.org/api_docs/python/tf/keras/layers/LSTM)
+with `return_sequences=True`, the rest works the same way as CNN.
```
alextp: My impression is that tf.keras.layers.LSTM returns the output in shape [sequence_length, batch_size, ...] while CNNs operate in [batch_size, sequence_length, ...].

So maybe mention here that you'll need a transpose to use your attention layer on an LSTM?

rfcs/20190115-dense-attention.md
```diff
@@ -0,0 +1,506 @@
...
+
+Unfortunately, this technique does not cover sequence-to-sequence RNN models.
```
alextp: By "this technique" do you mean your proposal?

If so I think your proposal should support masking, as it's very common.

rfcs/20190115-dense-attention.md
```diff
@@ -0,0 +1,506 @@
...
+
+*   https://github.com/keras-team/keras/issues/5738
+*   https://github.com/keras-team/keras/issues/4962
+*   https://github.com/keras-team/keras/issues/2525
+
+## Design Proposal
+
+We propose to implement the following common attention layers:
+
+*   `Attention`: Basic dot-product attention, a.k.a. Luong-style attention.
```
alextp: This section should put the equations for both forms, for clarity.

rfcs/20190115-dense-attention.md
```diff
@@ -0,0 +1,506 @@
...
+    typically calculated from the query length tensor. Used to mask the output
+    tensor. This is similar to the `mask` argument of
+    [tf.keras.backend.rnn](https://www.tensorflow.org/api_docs/python/tf/keras/backend/rnn).
+*   `value_mask` (optional) of shape `[batch_size, Tv]`. Boolean tensor,
+    typically calculated from the value length tensor. It is used to mask
+    `value` elements beyond this length so they do not contribute to the result.
+
+The output is of shape `[batch_size, Tq, dim]`.
+
+Following the pattern of other Keras layers, we pass the list `[query, value]`
```
alextp: This feels awkward, specially passing a tuple as the mask argument. Why not have query_mask and value_mask arguments separately, for better self-documenting code?

Similarly for query and value; it's nice if users can tell easily which is which.

rfcs/20190115-dense-attention.md
```diff
@@ -0,0 +1,506 @@
...
+Although not the primary focus of this proposal, the Attention layers work with
+RNN networks, such as
+[tf.keras.layers.LSTM](https://www.tensorflow.org/api_docs/python/tf/keras/layers/LSTM):
+When creating the LSTM, users need to set `return_sequences=True`, and the rest
+works the same way as CNN. It is unclear whether this method suffices to create
+the most common RNN+Attention models.
+
+We will first work on the implementation for Tensorflow.
+
+### Self-Attention
```
alextp: Most uses of self-attention that I know of are seq2seq and so want masking.

rfcs/20190115-dense-attention.md
```diff
@@ -0,0 +1,506 @@
...
+
+The Self-Attention variant can be implemented by passing the same tensor to both
+`query` and `value`.
+
+### Multi-Head Attention
```
alextp: I think this is covered by your proposal; just use multiple Attention layers; each will represent one attention head.

rfcs/20190115-dense-attention.md
```diff
@@ -0,0 +1,506 @@
...
+rather than writing custom ones.
+
+### Position Representations
+
+DNN+Attention networks do not model relative or absolute position information in
+their structure. Instead, position information is modeled as an additional term
+in the model output. The proposed techniques can be implemented as an additional
+feature in the Attention API.
+
+*   https://arxiv.org/abs/1803.02155 describes how relative position
```
alextp: It'd help to clarify here whether the position-dependent attention implementations compose with this layer or would require new layers.

rfcs/20190115-dense-attention.md
```diff
@@ -0,0 +1,506 @@
...
+    representation can be added to dot-product attention.
+*   https://arxiv.org/abs/1503.08895 and https://arxiv.org/abs/1706.03762 show
+    how absolute position information can be added as a deterministic function
+    of position.
+
+### 2D and 3D
```
alextp: It feels to me like we should be able to support n-d attention (not just 2d or 3d) with ~the same code, so we should do it.

LanceNorskog:
The amount of "does this but does not do that" cautions in the proposal confirms a suspicion I've had for awhile: the Keras API is a little too low-level.

rfcs/20190115-dense-attention.md
roumposg: Thanks. I put pseudocode in the pydoc in the following section. Let me add that here, too (once I figure out  how to do this in github, sorry for the delay).

rfcs/20190115-dense-attention.md
roumposg: Francois suggested a list, because the same pattern is used in other places, such as tf.keras.layers.Add. See the code in https://github.com/tensorflow/tensorflow/blob/r1.12/tensorflow/python/keras/layers/merge.py#L205

rfcs/20190115-dense-attention.md
roumposg: Can you clarify what you mean? I think the mask argument this api supports is good enough for DNN/CNN networks, such as transformer. WDYT?

rfcs/20190115-dense-attention.md
roumposg: Good point. It is not very pretty, but it works. Let me update the text.

rfcs/20190115-dense-attention.md
roumposg: Good point. For relative position representation, we need to modify the attention layer. Absolute position can be implemented as a separate layer that composes with this layer. Let me update the text.

rfcs/20190115-dense-attention.md
roumposg: That's true. Let me update the text.

roumposg:
@LanceNorskog thanks for the feedback. Do you have any suggestions on how to make it more powerful, starting from this review? Thanks.

rfcs/20190115-dense-attention.md
alextp: Let's discuss in the design review.

rfcs/20190115-dense-attention.md
```diff
@@ -0,0 +1,532 @@
...
+feed them one by one. So, the output of the attention layer at timestep T
+affects the output at T+1.
+
+## Previous Work
+
+[tf.contrib.seq2seq.LuongAttention](https://www.tensorflow.org/api_docs/python/tf/contrib/seq2seq/LuongAttention)
+and
+[tf.contrib.seq2seq.BahdanauAttention](https://www.tensorflow.org/api_docs/python/tf/contrib/seq2seq/BahdanauAttention)
+are implementations of dot-product (Luong) and additive (Bahdanau) Attention
+respectively for RNN in Tensorflow. This proposal is based on this
```
guillaumekln: By "based", do you mean it will extend classes from `tf.seq2seq`? It's unclear to me if this proposal has this recent change in mind: https://github.com/tensorflow/tensorflow/commit/b797012b39f89c45f13336b36c6b73e9ad93d815.

roumposg: Thank you for you comment. I mean that it is inspired from it, and that the implementation is consistent. Let me update the text.

Yes, I am aware of this change. It covers RNN networks, and it moves in the same direction as my proposal. Namely, implementing as keras.layer. Let me update the text to clarify that.

rfcs/20190115-dense-attention.md
roumposg: I see. This is a valid use case. Let me spend some more time to better understand this case and get back to you, thanks.

rfcs/20190115-dense-attention.md
guillaumekln: Multi head attention is usually batched so I think it requires a special case.

roumposg: All tensors are assumed to be batched. Namely, input tensors are `[batch_size, dim]` etc. Is this what you mean?

guillaumekln: I meant multi head attention is usually a single layer that takes `[batch_size, num_heads, time, dim]` and not `num_heads` layers that take `[batch_size, time, dim]`.

rfcs/20190115-dense-attention.md
roumposg: I thought about this, and I don't think this can be implemented as a manual mask or a composable layer. It needs to be a feature of the layer.

I added a couple of options about how this can be implemented in the "Self-attention" section. Let me know what you think, thanks.

rfcs/20190115-dense-attention.md
roumposg: Well, then the user needs to split the tensor into `num_heads` tensors of shape `[batch_size, time, dim]`. This is one way to do it.

rfcs/20190115-dense-attention.md
qlzh727: The output shape for LSTM is [batch, timestep, unit] when return_sequences=True. In the case of return_seqences=False, the output shape is [batch, unit].

karmel:
Notes from the review meeting:

*   Substantial changes?
    *   Some alternatives were brought up-- BaseAttention, Alex brought up another and Self Attention.
    *   Public comments were addressed with edits.
*   Layers will need to share some code. Shared method is one option, BaseAttention class has the advantage of inheritance being a common paradigm in Keras.
    *   Would the base class be sufficient for all the variations of attention? ie, multi-head attention, which has an extra dimension.
    *   Multi-head could be done with reshaping, ie a Reshape layer. 
    *   Reshaping is (maybe) expensive on TPU.
    *   But the user shouldn't reshape-- would a child class of base attention be able to implement that?
    *   **[TODO]: create a simple GNMT/Transformer, not necessarily working, to show what this would look like.**
*   What about Recurrent Attention-- will this work for that?
    *   There have been attempts to add this to Keras that were complicated.
    *   RNN attention is more complicated. The generated value has to be combined with states, fed to the next step. Inputs are only seen per timestep. 
    *   In other words, the implementation is totally different, it just happens to share a name.
    *   **[TODO]: ideally, come up with a name that distinguishes from Recurrent Attention-- BaseDenseAttention, BaseCausalAttention. Make sure it's clear in the docstring, point to recurrent implementations**
    *   Is there anything that is common, even if just the interface? The API should be reuasable in some parts, but it's tricky. Probably not worth trying to force into the same API.
*   Arg passing -- (query, value), (q_mask, v_mask) versus named kwargs.
    *   Named is nice, but keras convention is lists/tuples
    *   There are some downsides to not going with the convention in terms of how data is passed through to Keras, ie layers that generate masks expect to pass them through as masks=(0 … N)
    *   **Conclusion: use inputs=(...), masks=(...)**
*   Self-attention wants to prevent flow of info from the future to the past. Needs an additional mask. How would this work with the current proposal?
    *   Would have to be implemented by the layer, with a user-passed flag.
    *   causal_mask? Boolean.
    *   In convolutions there is precedent for padding=causal, so we can use that. But padding doesn't make sense in this context.
    *   **use_causal_mask seems fine. Analogous to use_bias.**
    *   But causal mask only makes sense in self-attention. Throw an error in normal cases? It doesn't make sense in terms of the model, but it does make mathematical sense. Let people do it, as you can imagine cases where that makes sense. Throw an error if the sequences are different lengths.
*   Sync to keras-team/keras: we can push out a project proposal for Keras community.
*   In attention is all you need, there is a "key"-- where is that? 
    *   They always use the same for key and value. 
    *   In seq2seq, there is a different key and value.
    *   **You could allow inputs=(query, value, key)**
    *   We should be able to flesh this out in the example noted above.

PR #55, opened 2019-01-17
RFC: Embedding and Partitioned Variables in TF 2.0 

yuefengz:
**Review period will close at the end of 2019-01-30**

| Status        | Accepted                                                                  |
| :------------ | :------------------------------------------------------------------------ |
| **Author** | Yuefeng Zhou (yuefengz@google.com)                                        |
|**Sponsors**| Alexandre Passos (apassos@google.com), Martin Wicke (wicke@google.com), Josh Levenberg  (joshl@google.com) |
|**Updated** | 2019-1-16                                                                 |

Summary: this RFC describes the design and interfaces for embeddings and partitioned variables in TF 2.0 based on Distribution Strategy. We propose to support embeddings and sharded layers with DistributionStrategy and Keras' layer API. We also propose to shard embeddings and layers only in `ParameterServerStrategy` and mirror them in `MirroredStrategy` or `CollectiveAllReduceStrategy.`

### DIFF
--- a/rfcs/20190116-embedding-partitioned-variable.md
+++ b/rfcs/20190116-embedding-partitioned-variable.md
@@ -277,6 +277,34 @@ def embedding_lookup(
 We will remove `parition_strategy` from all other library code.
 
 
+#### Save and Restore
+
+
+The current mechanism of saving and restoring `PartitionedVariable` which relies on private attributes on variables called `_save_slice_info` can continue to work. Saver reads `_save_slice_info` to save all partitions into one tensor, which allows this one tensor to be restored to different numbers of partitions as long as the new `PartitionedVariable` has the same full shape.
+
+To make the logic more TF 2.0-like, we can implement it differently via `CheckpointableBase`:
+
+```python
+class PartitionedVariable(checkpointable.CheckpointableBase):
+
+  def _gather_saveables_for_checkpoint(self):
+    def _saveable_factory(name=self._name):
+      return _PartitionedSaveable(name, self._partitions)
+    return {checkpointable.VARIABLE_VALUE_KEY: _saveable_factory}
+
+ 
+class PartitionedSaveable(saveable_object.SaveableObject):
+  
+  def __init__(self, name, partitions):
+    pass
+  
+  def restore(self, restored_tensors, restore_shapes):
+    pass
+```
+
+The implementations details may subject to change. The saving and restoring of `ShardedVariable` should be similar.
+
+
 ### Post-TF 2.0 Work
 
 
--- a/rfcs/20190116-embedding-partitioned-variable.md
+++ b/rfcs/20190116-embedding-partitioned-variable.md
@@ -131,7 +131,7 @@ Since Distribution Strategy is a library that is designed to parallelize and sca
 
 |          |Embeddings|Partitioned Layer (for Loadbalance Only)|
 |:-------- |:-------- |:----------------|
-|Mirrored/ CollectiveAllReduce Strategy|Pre-TF 2.0: <br>w/ Keras' Embedding layer API: won't be sharded but will be mirrored on each replica. <br><br>Post-TF 2.0: <br>w/ Distribution Strategy's API directly: possibly allow sharding; allow being placed on host; or dense update. <br><br>`Heuristics` in Embedding layer to auto-set performance improvement hints.|Pre-TF 2.0: <br>w/ Keras' layer API: won't be sharded but will be mirrored on all replicas, just like a normal layer. <br><br>Post-TF 2.0: w/ Distribution Strategy's API directly: allow being mirrored on each host. <br><br>Heuristics in Embedding layer to auto-place larger layers on host.|
+|Mirrored/ CollectiveAllReduce Strategy|Pre-TF 2.0: <br>w/ Keras' Embedding layer API: won't be sharded but will be mirrored on each replica. <br><br>Post-TF 2.0: <br>w/ Distribution Strategy's API directly: possibly allow sharding. <br><br>Add heuristics in Distribution Strategy to auto-set performance improvement hints, such as whether to place an embedding on host or whether to cast sparse updates to dense tensors.|Pre-TF 2.0: <br>w/ Keras' layer API: won't be sharded but will be mirrored on all replicas, just like a normal layer. <br><br>Post-TF 2.0: w/ Distribution Strategy's API directly: allow being mirrored on each host. <br><br>Add heuristics in Distribution Strategy to place some layers on host.|
 |ParameterServer Strategy|Pre-TF 2.0:<br>w/ Keras' Embedding layer API: probably sharded, according to the `partitioner` passed to its constructor.<br><br>Post-TF 2.0:<br>w/ Distribution Strategy's API directly: allow creating embedding variables with different partitioners.|Pre-TF 2.0:<br>w/ Keras' layer API: probably sharded, according to the `partitioner` passed to its constructor which applies globally to all layers.|
 
 In the `ParameterServerStrategy`, the goal of supporting partitioned variable is to match the existing behavior in 1.x. We will propose its APIs for TF 2.0 below. 
@@ -276,6 +276,10 @@ def embedding_lookup(
 
 We will remove `parition_strategy` from all other library code.
 
+We understand this is not a smooth API transition for users since the default `partition_strategy` of `tf.nn.embedding_lookup` is "mod". Also, the saved model populated by 1.x TensorFlow using "mod" strategy will probably not work in 2.0.
+
+Since `embedding_lookup` won't be a public API in TensorFlow 2.0, we recommend users use Keras' `Embedding` layer, even in `Estimator`'s model function. The other option is to use the legacy `tf.compat.v1` symbol if users have to handle embeddings by themselves.
+
 
 #### Save and Restore
 
@@ -313,6 +317,14 @@ In parameter server architecture, model variables can sometimes be `AggregatingV
 The following is planned to be done after TF 2.0's release. We describe their high-level ideas and their details are subject to change.
 
 
+#### Rebalance Embeddings in Embedding layer
+
+
+The main reason users use "mod" partition strategy is because they would like to use "mod" to achieve their load-balancing goal when their embeddings are not evenly distributed in terms of frequency of being referenced. Therefore in order to accomplish this goal with only "div" partition strategy, we can rebalance embeddings in the Keras Embedding layer, optionally given a frequency map/list, and remember the mapping in a variable or a hash table for lookup.
+
+This task will be prioritized after TF 2.0.
+
+
 #### Add Partitions in `experimental_create_sharded_variable`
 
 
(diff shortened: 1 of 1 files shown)


byronyi:
Gentle ping to @yangjunpro and @wangsiyu; it seems you guys were eager to this feature in the past.

https://github.com/tensorflow/tensorflow/pull/22473 https://github.com/tensorflow/tensorflow/pull/22937 https://github.com/tensorflow/tensorflow/pull/23254

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+1.  shard storage: occasionally embedding variables can be too large to fit in the memory of one machine.
+1.  reduce contentions in asynchronous training: updates from different workers could possibly end up on different partitions which reduces possibility of overriding each other's update and using locking, although this depends on the implementation of embedding update.
+
+
+##### Partition Strategy: div or mod
+
+
+There are two strategies to look up embedding vectors on a sharded embedding variable: "div" and "mod". See an explanation [here](https://www.tensorflow.org/api_docs/python/tf/nn/embedding_lookup). "Mod" strategy may be useful when references to embedding slices are not evenly distributed over their indices.  However, "mod" strategy is a poor approximation to the actual load balancing users want. For example, when vocabulary is sorted by frequency, with "mod" strategy the first parameter server always has larger load than the second one. Furthermore, its current checkpointing mechanism also prevents users from migrating to a cluster with different number of parameter servers.
```
byronyi: Would it be possible to support user-provided hash function, or a step further, consistent hashing? It enables PS level 1 fault tolerance, and it would be particularly helpful in the case of large-scale embeddings (thus more ps and workers).

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+
+## Background
+
+
+### Embeddings
+
+
+An [embedding](https://www.tensorflow.org/guide/embedding) is a mapping from discrete objects, such as words, to vectors of real numbers. In TensorFlow implementation, we use a variable to hold all these vectors for a feature column. All vectors in an embedding variable have the same size but the sizes of embedding variables may vary. 
+
+We don't always use the whole embedding variable in a training step, the result of embedding lookup is <code>[IndexedSlices](https://www.tensorflow.org/api_docs/python/tf/IndexedSlices)</code> which is essentially a dict mapping from indices to vectors. So is the gradient to an embedding variable.
```
alextp: The result of embedding lookup is not indexedslices; the gradient of embedding lookup is indexedslices

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+
+These APIs `tf.get_variable` and `tf.variable_scope` will be gone in TF 2.0.
+
+
+### Possible Problems with The `PartitionedVariable` Class 
+
+
+Although we have a class called `PartitionedVariable`,  we often concatenate all of their components to pretend it is a single variable.
+
+For example, `tf.matmul(partitioned_var, inputs)` would concatenate the list of variables in the `partitioned_var` before the execution of `matmul` as if the `partitioned_var` is a single variable. With the concatenation, the `partitioned_var` would only serve for loadbalancing purposes.
```
alextp: And this will in fact be wrong if you want mod partitioning instead of div partitioning.

(another issue is that keeping track of mod vs div is delegated to users who might no longer have access to this information)

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+    pass
+```
+
+Users may have to concatenate variables themselves when dealing with it directly in post-TF 2.0 and using it for loadbalancing purposes rather than embeddings. More convenience methods will be added post-TF 2.0.
+
+
+#### Distribution strategy
+
+
+##### ParameterServerStrategy
```
alextp: There should be a way to create PartitionedVariables still for the load-balancing case, right?

I don't see it here. Clarify?

rfcs/20190116-embedding-partitioned-variable.md
yuefengz: Done.

rfcs/20190116-embedding-partitioned-variable.md
yuefengz: Done. Layers should create load-balancing variables if `partitioner` is specified in `ParameterServerStrategy`.

rfcs/20190116-embedding-partitioned-variable.md
yuefengz: That sounds more of general feature request for `ParameterServerStrategy` rather than for embeddings. 

Consistent hashing, if I understand it correctly, is more for scenarios that need high availability. When a server goes down, you'll need a mechanism to migrate its key-values to other servers. However, PS only checkpoints once in a while, so you'll need heads-up from container orchestration system in order to prepare variable migration before a PS goes down. You'll also need to tell workers about the configuration change and workers may have to rebuild the whole graph for the re-assignment of variables. This can be something built on top of TensorFlow although I guess a simpler solution is just to restart the whole cluster when a server needs to be removed since high availability is not the most important goal.

rfcs/20190116-embedding-partitioned-variable.md
yuefengz: I was talking about non-embedding loadbalancing variables. I rephrased it and added a paragraph for the ugliness of the `partition_strategy` argument.

yangjunpro:
> tensorflow/tensorflow#22473

Sorry for missing this discussion thread for a while. We will take a serious look at this RFC and provide feedback then.

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+1.  shard storage: occasionally embedding variables can be too large to fit in the memory of one machine.
+1.  reduce contentions in asynchronous training: updates from different workers could possibly end up on different partitions which reduces possibility of overriding each other's update and using locking, although this depends on the implementation of embedding update.
+
+
+##### Partition Strategy: div or mod
+
+
+There are two strategies to look up embedding vectors on a sharded embedding variable: "div" and "mod". See an explanation [here](https://www.tensorflow.org/api_docs/python/tf/nn/embedding_lookup). "Mod" strategy may be useful when references to embedding slices are not evenly distributed over their indices.  However, "mod" strategy is a poor approximation to the actual load balancing users want. For example, when vocabulary is sorted by frequency, with "mod" strategy the first parameter server always has larger load than the second one. Furthermore, its current checkpointing mechanism also prevents users from migrating to a cluster with different number of parameter servers.
```
yangjunpro: Based on the above sentence, seemingly it wants to tell us that "div" strategy performs bettern than "mod" when vocabulary is sorted by frequency. However, due to the nature of "mod" and "div" partition strategy, I think "mod" could bring more balanced partition. So I am a little bit puzzled about the example provided here.

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+
+
+### Possible Problems with The `PartitionedVariable` Class 
+
+
+Although we have a class called `PartitionedVariable`, we often concatenate all of their components to pretend it is a single variable. With the concatenation, the `partitioned_var` would only serve for loadbalancing purposes. For example, `tf.matmul(partitioned_var, inputs)` would concatenate the list of variables in the `partitioned_var` before the execution of `matmul` as if the `partitioned_var` is a single variable. 
+
+However, to parallelize computation, people have to write methods that respect the partitions, e.g. `tf.nn.sampled_softmax_loss`. Therefore, this class has been overloaded by different use cases.
+
+On the other hand, when `PartitionedVariable` is used for sharded embeddings, `partitioned_strategy` has to be kept consistent when it is required by several methods down the stream such as `tf.nn.embedding_lookup` and `tf.nn.nce_loss`. This is uncessary for users and any inconsistency would lead to subtle bugs.
```
yangjunpro: Maybe we could elaborate more about the justification.
```suggestion
On the other hand, when `PartitionedVariable` is used for sharded embeddings, `partitioned_strategy` has to be kept consistent when it is required by several methods down the stream such as `tf.nn.embedding_lookup` and `tf.nn.nce_loss` which requires an additional `partition_strategy` argument. This is uncessary for users and any inconsistency would lead to subtle bugs. 
```

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+
+On the other hand, when `PartitionedVariable` is used for sharded embeddings, `partitioned_strategy` has to be kept consistent when it is required by several methods down the stream such as `tf.nn.embedding_lookup` and `tf.nn.nce_loss`. This is uncessary for users and any inconsistency would lead to subtle bugs. 
+
+
+## Goals
+
+
+1.  Support partitioned embedding and partitioned layer in TF 2.0 in parameter server architecture via Distribution Strategy's and Keras layer's API. 
+1.  Better support for embeddings in mirrored and collective allreduce architecture. We will not shard them in this architecture in our pre-TF 2.0 design.
+2.  We will only support "div" partition strategy but as a post-TF 2.0 work we will support re-balancing embeddings in the Keras' `Embedding` layer. That means we will not support "mod" partition strategy any more.
```
yangjunpro: So, the reason for choosing "div" partition strategy is regarding to the checkpoint compatibility issue?

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+3.  We also would like the API to enable people to easily switch between different architectures without needing to modify their models.
+
+
+## Non-goals
+
+
+1.  We don't have plan to support any flavor of model parallelism beyond the current implementations of embedding lookup and loss functions that respect partitioning.
```
yangjunpro: Why do we emphasize the "loss function" here in addition to the "embedding lookup"?
In my understanding, with partition layer support, we could do more than partitioning the loss functions?

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+The following is planned to be done before TF 2.0's release.
+
+
+#### A new type `ShardedVariable`
+
+
+As discussed above, `PartitionedVariable` concatenates component variables silently which can only serve the purpose of loadbalancing. So we need another object for uses cases where computation needs to be parellelized or storage needs to be sharded. Users don't have to deal with it as long as they use Keras' layer API and it won't be exposed as a public API in pre-TF 2.0 stage.
+
+```python
+class ShardedVariable(object):
```
yangjunpro: Will it be better by providing some specific use cases with SharededVariable to demonstrate its difference against the PartitionedVariable?

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+
+
+##### Single host
+
+
+On a single host, embeddings can be placed on host memory or mirrored on devices.
+
+When an embedding variable is small, it may benefit from being mirrored on devices. Lookup can be performed local to each device. Updates can be casted to dense tensors and use the existing allreduce primitive to exchange gradients. This can still be faster than running allgather on their corresponding sparse updates.
+
+When an embedding variable is larger, we'll need an efficient allgather primitive to exchange updates between devices. Alternatively, we can place it on host memory at the cost of transferring embeddings from host to devices in the forward pass and gathering updates from devices to host in the backward pass.
```
yangjunpro: Have we considered for embedding partition using GPU memory rather than host memory? This may bring  more complexity, but may also bring more performance gains since host2device traffic can be reduced. 
Also, it looks that many embedding-dominating models are not complex so that GPU device memory may be a potential choice for holding the embedding partition.
Actually, inside Alibaba, we are working on prototyping a new strategy called as PEARLStrategy(Partitioned Embedding And RepLicated variables ) to support GPU-based embedding partitioning.

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+##### Multiple hosts
+
+
+After excluding sharding across devices, there are still three ways of handling embeddings on multiple hosts:
+
+*   sharding across all workers' host memory;
+*   mirroring on all workers' host memory;
+*   mirroring on all replicas, i.e. devices.
+
+On multiple hosts, mirroring could be cheaper than sharding in terms of communication cost since it requires smaller number of communications although it transfers more data. This is true even compared to the optimal implementation of sharding. When updates to an embedding are not very sparse, converting them to dense updates and applying allreduce in the mirrored case can be faster than sharding which relies on all-to-all personalized communication or all-gather. Furthermore, mirroring is much easier to implement.
```
yangjunpro: Can we elaborate it more? I think this only holds true when the sparsity is not high.

rfcs/20190116-embedding-partitioned-variable.md
yuefengz: The `ShardedVariable` is just like `ParititionedVariable` excepts it doesn't auto-concats partitions.  It will be passed to embedding lookup and a few loss functions. So it is pretty dummy.

rfcs/20190116-embedding-partitioned-variable.md
```diff
@@ -0,0 +1,498 @@
...
+       dtype: the data type of the variable.
+    """
+    pass
+```
+
+
+#### Layers
+
+
+We will need to call `strategy.experimental_create_sharded_variable()` in Keras' `Embeeding` layer. Under `ParameterServerStrategy` scope, all variables can potentially be `PartitionedVariable` for loadbalancing only.
```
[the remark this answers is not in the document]
yuefengz: Done.

rfcs/20190116-embedding-partitioned-variable.md
yuefengz: I've just updated these paragraphs. Sharding on GPUs might be faster but more difficult to implement. We'll implement mirroring embeddings and allgather first.

I am not sure how faster it is to shard across GPUs than to put it on host memory. You've probably done some experiment : ). If it worths the complexity, we'll definitely implement it.

rfcs/20190116-embedding-partitioned-variable.md
yuefengz: Maybe some sentence is confusing. I meant when some gradients are not very sparse, we can densify and all-reduce them, which would be faster than allgather. This is one of the benefits of mirroring embedding.

liyi193328:
any progress here?

pavanky:
@yuefengz is there a development branch we can follow for this feature ?

PR #56, opened 2019-01-17
RFC: Stateful Containers with tf.Module

tomhennigan:
**This review is open for public comment until 2019-02-11**

| Status        | Accepted                                                          |
| :------------ | :---------------------------------------------------------------- |
| **Author(s)** | DeepMind: tomhennigan@google.com, mareynolds@google.com           |
|               | Google: agarwal@google.com, apassos@google.com, dberth@google.com |
| **Sponsor**   | apassos@google.com                                                |
| **Updated**   | 2019-01-16                                                        |

**Summary**

We propose a lightweight base class (`tf.Module`) for stateful containers in TensorFlow. This base class offers three main features:

-   **Naming** - variables and ops created inside modules are created within a name scope so they can easily be tracked back to their creator.
-   **Variable tracking** - variables referenced by a module (directly or via dependencies on other modules) can be retrieved via a simple API.
-   **Module tracking** - given a module a simple API exposes all modules that this module depends on.

Users are encouraged to subclass `tf.Module` to create their own stateful components and we expect and encourage an ecosystem of third party modules to evolve.

Thanks in advance for your feedback!

### DIFF
--- a/rfcs/20190117-tf-module.md
+++ b/rfcs/20190117-tf-module.md
@@ -403,7 +403,7 @@ class ParentModule(tf.Module):
 'parent_module/v:0'
 
 >>> p.child_module.name
-'child_module'
+'child'
 >>> p.child_module.name_scope.name
 'parent_module/child/'
 >>> p.child_module.v.name
--- a/rfcs/20190117-tf-module.md
+++ b/rfcs/20190117-tf-module.md
@@ -178,7 +178,6 @@ class Module(six.with_metaclass(ModuleMetaclass, tf.checkpoint.Checkpointable)):
   ...     self.b = tf.Variable(tf.zeros([output_features]), name='b')
   ...
   ...   def __call__(self, x):
-  ...     x = tf.convert_to_tensor(x, name='x')
   ...     y = tf.matmul(x, self.w) + self.b
   ...     return tf.nn.relu(y)
 
@@ -553,8 +552,6 @@ class Linear(tf.Module):
     self.output_size = output_size
 
   def __call__(self, x):
-    x = tf.convert_to_tensor(x, name='x')
-
     if not hasattr(self, 'w'):
       input_size = x.shape[1]
       initial_w = tf.random.truncated_normal(shape=[input_size, self.output_size],
--- a/rfcs/20190117-tf-module.md
+++ b/rfcs/20190117-tf-module.md
@@ -296,9 +296,9 @@ class Module(six.with_metaclass(ModuleMetaclass, tf.checkpoint.Checkpointable)):
     >>> c = tf.Module()
     >>> a.b = b
     >>> b.c = c
-    >>> assert list(a.children) == [b]
-    >>> assert list(b.children) == [c]
-    >>> assert list(c.children) == []
+    >>> assert list(a.owned_submodules) == [b]
+    >>> assert list(b.owned_submodules) == [c]
+    >>> assert list(c.owned_submodules) == []
 
     Returns:
       A generator over all child modules.
--- a/rfcs/20190117-tf-module.md
+++ b/rfcs/20190117-tf-module.md
@@ -1,0 +1,634 @@
+# tf.Module
+
+| Status        | Proposed                                                          |
+| :------------ | :---------------------------------------------------------------- |
+| **Author(s)** | DeepMind: tomhennigan@google.com, mareynolds@google.com           |
+|               | Google: agarwal@google.com, apassos@google.com, dberth@google.com |
+| **Sponsor**   | apassos@google.com                                                |
+| **Updated**   | 2019-01-16                                                        |
+| **Obsoletes** | N/A                                                               |
+
+## Overview
+
+We propose a lightweight base class (`tf.Module`) for stateful containers in
+TensorFlow. This base class offers three main features:
+
+-   **Naming** - variables and ops created inside modules are created within a
+    name scope so they can easily be tracked back to their creator.
+-   **Variable tracking** - variables referenced by a module (directly or via
+    dependencies on other modules) can be retrieved via a simple API.
+-   **Module tracking** - given a module a simple API exposes all modules that
+    this module depends on.
+
+Users are encouraged to subclass `tf.Module` to create their own stateful
+components and we expect and encourage an ecosystem of third party modules to
+evolve.
+
+## Motivation
+
+When defining neural networks it is convenient to describe the network as a DAG
+of connected components (e.g. datasets, layers, optimizers). In Python it is
+convenient to represent nodes in this DAG as objects which encapsulate node
+state (e.g. `tf.Variable`) and functionality (e.g. forward functions).
+
+We propose a lightweight "module" abstraction that serves as a base class for
+all types of node in this DAG (e.g. layers, optimizers, metrics) which is not
+tied to a particular higher level framework (e.g. Keras, Estimator or Sonnet).
+The module abstraction standardizes naming of variables and ops (to simplify
+debugging/visualisation) and dependency tracking (e.g. retrieving `tf.Variable`s
+and `tf.Module`s referenced by a given root module).
+
+Having a standard abstraction for stateful TensorFlow components will encourage
+users to create loosely coupled components which are easy to reason about in
+isolation while providing a standard mechanism for common tasks in a larger
+assembly (e.g. finding all trainable variables for a module tree given a root).
+
+We see such a lightweight abstraction as being particularly important for
+cutting edge research, where it is not possible for library/framework authors to
+reason about what their users will create. We believe that lightweight
+abstractions which can be extended or composed to build more complex systems are
+the way forward.
+
+## Background
+
+TensorFlow V1 has a number of abstractions for handling TensorFlow state (e.g.
+`tf.Variable`s). Broadly you can split them into "collections based" (e.g.
+`tf.variable_scope`, `tf.get_collection`, `tf.make_template`, `tf.layers.Layer`)
+and "Object Oriented" (e.g. `tf.keras.layers.Layer`).
+
+TensorFlow V2 is standardising on OO and `tf.keras` as the only stateful
+abstraction inside the default TensorFlow pip installation.
+
+Outside of TensorFlow there exist a number of third party libraries which also
+offer abstractions for managing TensorFlow state. For example the Sonnet library
+from DeepMind is built on top of `tf.make_template`.
+
+## Requirements
+
+### Functional
+
+**Naming `tf.Variable`s (and `tf.Operation`s)**
+
+-   Each module instance must have a name scope. This name should include the
+    name scope from when the module is constructed.
+-   Every method on the module must enter this name scope before calling user
+    code (such that variables and operations are correctly named).
+
+**`tf.Variable` creation and reuse**
+
+-   Modules should be able to create and reuse variables.
+-   Users should be able to intercept variable creation without changing module
+    code (e.g. via `tf.variable_creator_scope`).
+-   Users should be able to access and change variables on a module in an
+    intuitive way (e.g. being able to copy variables between modules
+    `module.w = other_module.w`).
+
+**`tf.Variable` tracking**
+
+-   Variables created, owned or used by a module should be retrievable (e.g.
+    mod.variables should provide a collection of variables for mod and all of
+    its descendent modules).
+-   Which variables are tracked should be obvious (e.g. all object properties)
+    and "opt-out" rather than "opt-in" (e.g. the most obvious way to create and
+    reuse a variable should automatically be tracked).
+
+**`tf.Module` tracking**
+
+-   Modules should track dependencies (e.g. a sequential should provide access
+    to its layers) and support checkpointing as a graph.
+
+**Performance**
+
+-   TensorFlow 2 is eager first meaning Python code is on the critical path.
+    tf.Module should have as little runtime overhead as reasonable to avoid
+    getting in the way.
+-   Additionally tf.Module should be fully compatible with tf.function allowing
+    high performance graphs to be extracted when appropriate.
+
+### Non-functional
+
+**Wide adoption**
+
+-   `tf.Module` (or types derived from it) should be the obvious choice for
+    research code and useful in general.
+-   It should not favor a specific theme (e.g. supervised learning) vs. another
+    (e.g. adversarial training).
+-   Having shared abstractions is critical for a research teams to facilitate
+    easy collaboration.
+
+**Researcher approved**
+
+-   `tf.Module` should be designed in collaboration with researchers and the API
+    should match their preferences.
+
+**Unbloated**
+
+-   `tf.Module` should be a thin abstraction which is trivial to reason about.
+-   The implementation should be clean and readable, ideally no more than a few
+    100 lines of code.
+-   The external API should have very few public methods.
+
+**Tightly specified**
+
+-   The specification for a `tf.Module` should be clear and agreed upon.
+-   Future additions to `tf.Module` should be critically evaluated against the
+    original design goals and moved elsewhere when they do not align.
+-   New features (e.g. supporting mixed precision training) should not require
+    changes to `tf.Module`.
+
+**Loosely coupled**
+
+-   Using `tf.Module` should not tie you to a particular framework or mentality.
+-   It should be trivial to integrate a subclass of `tf.Module` into any
+    codebase.
+
+**Defensive**
+
+-   The design of `tf.Module` should optimize for helping avoid common mistakes
+    (e.g. accidentally not tracking variables).
+-   If functionality is required (e.g. calling `super(..).__init__()` in
+    subclasses) we should error early (e.g. immediately after the module
+    constructor returns) to remind users if they forget.
+
+## Detailed Design
+
+### Class structure
+
+We propose a pure Python class `tf.Module` which has a metaclass
+`tf.ModuleMetaclass` to wrap all methods in the modules name scope. This design
+allows for submodules or variables to created in any method (including
+`__init__` and `__call__`). There is more detail on this decision in the
+detailed design below.
+
+```python
+class Module(six.with_metaclass(ModuleMetaclass, tf.checkpoint.Checkpointable)):
+  """Base neural network module class.
+
+  A module is a named container for `tf.Variable`s, other `tf.Module`s and
+  functions which apply to user input. For example a dense layer in a neural
+  network might be implemented as a `tf.Module`:
+
+  >>> class Dense(tf.Module):
+  ...   def __init__(self, in_features, output_features):
+  ...     super(Linear, self).__init__()
+  ...     # NOTE: We're creating variables in the ctor, but you can actually
+  ...     #       create them in any method (including __call__) you like.
+  ...     self.w = tf.Variable(
+  ...         tf.random.normal([input_features, output_features]), name='w')
+  ...     self.b = tf.Variable(tf.zeros([output_features]), name='b')
+  ...
+  ...   def __call__(self, x):
+  ...     x = tf.convert_to_tensor(x, name='x')
+  ...     y = tf.matmul(x, self.w) + self.b
+  ...     return tf.nn.relu(y)
+
+  You can use the layer as you would expect:
+
+  >>> d = Dense(input_features=64, output_features=10)
+  >>> d(tf.ones([100, 64]))
+  <tf.Tensor: ...>
+
+  By subclassing `tf.Module` instead of `object` any variables created inside
+  the module are automatically created within the modules name scope:
+
+  >> d.w.name
+  "dense/w:0"
+
+  In eager mode this is useful for debugging, and when used with `@tf.function`
+  the use of name scopes gives operations (e.g. matmul) useful names as well.
+
+  As well as automatic naming, the Dense module inherits methods for tracking
+  its variables:
+
+  >>> d.variables
+  (<tf.Variable 'dense/b:0' ...>, <tf.Variable 'dense/w:0' ...>)
+  """
+
+  @property
+  def name(self):
+    """Returns the name of this module as passed or determined in the ctor.
+
+    NOTE: This is not the same as the `self.name_scope.name` which includes
+    parent module names.
+    """
+
+  @property
+  def name_scope(self):
+    """Returns a `tf.name_scope` instance for this class.
+
+    NOTE: The module name scope includes all parent modules as determined
+    when this module was created.
+    """
+
+  @property
+  def variables(self):
+    """Returns a tuple of variables owned by this module and it's submodules.
+
+    Returns:
+      A tuple of variables for the current module (sorted by attribute name)
+      followed by variables from all submodules recursively (depth first).
+    """
+
+  @property
+  def owned_variables(self):
+    """Returns a tuple of variables that are attributes of the current module.
+
+    See `variables` for a property which returns all variables from the current
+    module and all it's submodules recursively.
+
+    Returns:
+      A tuple of variables which are attributes of the current module. Will
+      yield variables inside nested structures (lists etc) but not in other
+      modules.
+    """
+
+  @property
+  def trainable_variables(self):
+    """Returns a tuple of variables owned by this module and it's submodules.
+
+    Returns:
+      A tuple of variables for the current module (sorted by attribute name)
+      followed by variables from all submodules recursively (depth first).
+    """
+
+  @property
+  def owned_trainable_variables(self):
+    """Returns a tuple of variables that are attributes of the current module.
+
+    See `variables` for a property which returns all variables from the current
+    module and all it's submodules recursively.
+
+    Returns:
+      A tuple of variables which are attributes of the current module. Will
+      yield variables inside nested structures (lists etc) but not in other
+      modules.
+    """
+
+  @property
+  def submodules(self):
+    """Returns an iterator of all submodules recursively.
+
+    Submodules are modules which are properties of this module, or found as
+    properties of modules which are properties of this module (and so on).
+
+    >>> a = tf.Module()
+    >>> b = tf.Module()
+    >>> c = tf.Module()
+    >>> a.b = b
+    >>> b.c = c
+    >>> assert list(a.submodules) == [b, c]
+    >>> assert list(b.submodules) == [c]
+    >>> assert list(c.submodules) == []
+
+    Returns:
+      A generator over all submodules.
+    """
+
+  @property
+  def owned_submodules(self):
+    """Returns an iterator of immediate child modules.
+
+    Child modules are modules which are found as properties of the current
+    module.
+
+    >>> a = tf.Module()
+    >>> b = tf.Module()
+    >>> c = tf.Module()
+    >>> a.b = b
+    >>> b.c = c
+    >>> assert list(a.children) == [b]
+    >>> assert list(b.children) == [c]
+    >>> assert list(c.children) == []
+
+    Returns:
+      A generator over all child modules.
+    """
+
+class ModuleMetaclass(type):
+  def __new__(mcs, name, bases, clsdict):
+    """Modifies module subclasses when they are defined.
+
+    Wraps all methods and properties in clsdict such that enter the modules
+    name scope before evaluation.
+
+    >>> class MyModule(tf.Module):
+    ...   def __call__(self, x):
+    ...     if not hasattr(self, 'w'):
+    ...       self.w = tf.Variable(tf.zeros([x.shape[1], 64]), name='w')
+    ...     return tf.matmul(x, w)
+
+    Without any user intervention __call__ will enter the modules name scope,
+    allowing variables and submodules to be created. In some cases this is not
+    desirable, so there exists an opt out annotation(`@no_auto_name_scope`)
+    which instructs the metaclass to not wrap these functions.
+    """
+
+  def __call__(cls, *args, **kwargs):
+    """Creates new module instances.
+
+    To allow other modules/variables to be created in the constructor of a
+    module (and be named appropriately) we enter the module name scope inside
+    the base (Module) constructor. We close the name scope in the metaclass
+    before returning the new instance.
+
+    >>> class MyModule(tf.Module):
+    ...   def __init__(self):
+    ...     super(MyModule, self).__init__()
+    ...     # ^ Enters the module name scope but does not exit.
+    ...
+    ...     self.v = tf.Variable(1., name='v')
+    ...     # ^ Will be named "my_module/v"
+    ...
+    ...     self.m = MyOtherModule()
+    ...     # ^ Will be named "my_module/my_other_module"
+
+    >>> m = MyModule()
+    >>> # ^ Name scope will be closed thanks to metaclass.
+    """
+```
+
+### Naming
+
+Having human readable names for ops and variables is essential for debugging.
+Each `tf.Module` will have a name scope which is determined in the constructor.
+By default this will be a `lower_camel_case` version of the `ClassName`:
+
+```python
+class MyModule(tf.Module):
+  def __init__(self, name=None):
+    super(MyModule, self).__init__(name=name)
+    self.v = tf.Variable(1., name='v')
+
+>>> m = MyModule()
+>>> m.name
+'my_module'
+>>> m.name_scope.name
+'my_module/'
+>>> m.v.name
+'my_module/v:0'
+
+>>> m = MyModule(name='custom_name')
+>>> m.name
+'custom_name'
+>>> m.name_scope.name
+'custom_name/'
+>>> m.v.name
+'custom_name/v:0'
+```
+
+Modules created inside methods of other modules inherit their parent module
+name:
+
+```python
+class ParentModule(tf.Module):
+  def __init__(self):
+    super(ParentModule, self).__init__()
+    self.child_module = MyModule(name='child')
+    self.v = tf.Variable(1., name='v')
+
+>>> p = ParentModule()
+
+>>> p.name
+'parent_module'
+>>> p.name_scope.name
+'parent_module/'
+>>> p.v.name
+'parent_module/v:0'
+
+>>> p.child_module.name
+'child_module'
+>>> p.child_module.name_scope.name
+'parent_module/child/'
+>>> p.child_module.v.name
+'parent_module/child/v:0'
+```
+
+In TensorFlow 2 string matching of variable names is not required for
+distributed training (you can create remote variables eagerly, also see
+[DistributionStrategy](https://github.com/tensorflow/community/blob/master/rfcs/20181016-replicator.md))
+or checkpointing (this is done based on Python object structure) so the naming
+of variables and ops is just to aid with usability and visualisation.
+
+### Variables
+
+Users are free to create `tf.Variable` instances in any method at any time (as
+long as the super constructor has run).
+
+Variable reuse should be achieved using standard object oriented programming
+(e.g. assigning to object properties). `tf.Module` doesn't do anything behind the
+scenes to help (or hinder) variable reuse.
+
+We considered including a default implementation of `__call__` since lazy
+building is extremely common (e.g. constructing model parameters the first time
+an input is observed), however we backtracked on this for a number of reasons:
+
+-   Even among the authors there existed different preferences for how to enable
+    this pattern, it seemed unlikely we could make a choice which would satisfy
+    the community.
+-   Not all modules want a `__call__` method (e.g. an `Optimizer`) and we got
+    strong feedback that it is annoying to ignore unused methods inherited from
+    your base class.
+-   Encouraging a lazy pattern might make it harder to create modules with
+    multiple forward methods (e.g. a VAE module might be implemented with an
+    `encode` and `decode` method with no `__call__`).
+
+As such we leave the problem of how to enable lazy parameter construction to
+library authors who build on `tf.Module`s.
+
+### Checkpointing
+
+`tf.Module` will extend `tf.checkpoint.AutoCheckpointable` which automatically
+tracks variables which are object properties and enables restore on create etc.
+Users who want to learn more about checkpointing should look into this type.
+
+It is worth noting that object structure (e.g. the names of object parameters)
+are encoded into checkpoints, so library authors should take care to lay out
+their objects in a consistent way and defend their structure. `tf.Module` is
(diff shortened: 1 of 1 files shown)


rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+**Naming `tf.Variable`s (and `tf.Operation`s)**
+
+-   Each module instance must have a name scope. This name should include the
+    name scope from when the module is constructed.
+-   Every method on the module must enter this name scope before calling user
+    code (such that variables and operations are correctly named).
+
+**`tf.Variable` creation and reuse**
+
+-   Modules should be able to create and reuse variables.
```
alextp: What do you mean by reuse here?

My understanding is that modules only create variables, and reuse is done by reusing modules, right?

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+      loss = loss_fn()
+    if variables is None:
+      variables = tape.watched_variables()
+    self.apply_gradients(zip(variables, tape.gradients(loss, variables)))
+    return loss
+
+  def apply_gradients(self, vars_and_grads):
+    for variable, grad in vars_and_grads:
+      variable.assign_sub(grad * self.learning_rate)
+```
```
alextp: I'd like to comment here on the following points:

 1. Module should be the base class for tf.keras.Layer
 2. We explicitly chose not to force a call signature or build() strategy
 3. Pros and cons of the name Module (familiarity with pytorch being a big issue here)

awav:
@tomhennigan, That is awesome! It is very similar to what we have in [GPflow](https://github.com/GPflow/GPflow) - `Parameterized` class. With this change in TensorFlow we will be able to serialize our objects (currently we have custom saver) and integrate them with Keras models. 

Although, RFC could be extended with array-like modules, but don't confuse it with keras sequence. Keras sequence implies dependency between layers/modules, whereas array-like module is a different approach for organising submodules with indexing restricted to integer.

E.g. in [GPflow](https://github.com/GPflow/GPflow) we can construct a sum of the kernels (modules) and represent it as a super-kernel class or a combination kernel. Essentially, a combination kernel is a list of kernels, and when we call, let's say, __call__ method, it sums over sub-kernels results. 

I'm happy to give more examples if needed. Thanks!

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+By default this will be a `lower_camel_case` version of the `ClassName`:
+
+```python
+class MyModule(tf.Module):
+  def __init__(self, name=None):
+    super(MyModule, self).__init__(name=name)
+    self.v = tf.Variable(1., name='v')
+
+>>> m = MyModule()
+>>> m.name
```
leandro-gracia-gil: What happens if the module name ends in '/'? Usually that is used to set _absolute_ name scopes ignoring the current stack. Similarly, what happens if a child module does it?

leandro-gracia-gil:
Thank you very much for this proposal! We have a remarkably similar class in our custom framework, where we use it as the base for all our own layers and composable networks. It makes me happy to see something like this being added to TF 2.0.

Given the similarity, I'll try to keep an eye to the review in case I notice anything that reminds me of any problems we had developing or using our own approach, but so far I'd say it looks pretty good.

guillaumekln:
I was under the impression that inheriting from `tf.keras.layers.Layer` was the recommended way to create stateful and reusable components in TensorFlow 2.0. This proposal makes it more nuanced. IMO it should clarify when one should extend `tf.Module` versus extend `tf.keras.layers.Layer`.

rfcs/20190117-tf-module.md
tomhennigan: Your understanding is correct. Variable creation (`self.w = tf.Variable(..)`) and reuse (`tf.whatever(some_input, self.w)`) is enabled via OOP (no magic).

We considered re-implementing `variable_scope` and `get_variable` but felt that the OOP design was a lot cleaner and more Pythonic.

rfcs/20190117-tf-module.md
tomhennigan: Agreed re (1) and re (2) I call out why a bit in the detailed design if anyone wants more context.

Re (3) I feel strongly that `Module` is the right name and what users will expect. The term `Module` and its usage in ML is almost as old as I am, the oldest reference I know about (h/t @jekbradbury for the reference) is from SN/Lush [0]:

> In Lush, building and training a complex system is done by assembling basic blocks, called modules. Modules are subclasses of the class gb-module.

[0] http://lush.sourceforge.net/lush-manual/51d13a16626fd68f.html

tomhennigan:
> I was under the impression that inheriting from `tf.keras.layers.Layer` was the recommended way to create stateful and reusable components in TensorFlow 2.0. This proposal makes it more nuanced. IMO it should clarify when one should extend `tf.Module` versus extend `tf.keras.layers.Layer`.

If this RFC is accepted @alextp and I hope that we can make `tf.keras.layers.Layer` extend `tf.Module` (instead of `Checkpointable` as it does today). `Module` can help make the implementation of `Layer` simpler by implementing part of it's API (e.g. the `Layer.variables` property). We have not attempted to do this yet and I'm not sure if there would be any technical blockers.

Conceptually I see Keras `Layer` as enriched `Module`s with additional features (e.g. `build(shapes)`/`call(inputs)` pattern, json/yaml serialization, regularizers, losses, metrics, masking etc). If these features are useful to you, or if you are using your subclass in a Keras `Model` then Keras `Layer` is a good choice.

If the additional functionality of Keras is not useful for you then you may prefer to subclass `tf.Module` which is simpler with far fewer features. If `tf.Module` is not useful you may want to subclass `Checkpointable` or just `object` :smile:

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+>>> m = MyModule(name='custom_name')
+>>> m.name
+'custom_name'
+>>> m.name_scope.name
+'custom_name/'
+>>> m.v.name
+'custom_name/v:0'
+```
+
+Modules created inside methods of other modules inherit their parent module
```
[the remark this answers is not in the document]
tomhennigan: Whilst the design works in TF1 we have designed with TF2 in mind. In TF2 there are no variable scopes (we use name scope to name variables and tensors). In eager mode `tf.Tensor` does not have a `name` property, although if you inspect the graph from a `@tf.function` you do see tensor names.

Module inherits the `name_scope` from when the module is constructed. I've run the following with TF1 and as you can see the 'use' `name_scope` makes no difference inside the module but is applied when using the result of calling the module:

```python
class BarModule(tf.Module):
  def __call__(self):
    v = tf.Variable(1., name='v')
    t = tf.constant(1., name='t')
    return v, t

with tf.Graph().as_default():  # Enable TF1 graph mode.
  with tf.name_scope('create'):
    bar = BarModule()

  with tf.name_scope('use'):
    for i in range(2):
      print('iter', i)
      v, t = bar()
      x = v * t
      print('- v', v.name)
      print('- t', t.name)
      print('- x', x.name)
      print()
```

```
iter 0
- v create/bar_module/v:0
- t create/bar_module/t:0
- x use/mul:0

iter 1
- v create/bar_module/v_1:0
- t create/bar_module/t_1:0
- x use/mul_1:0
```

rfcs/20190117-tf-module.md
tomhennigan: In the experimental implementation of this I've enforced (in the constructor) that the module name is a valid Python 2 identifier [0], so `/` is not allowed.

[0] https://docs.python.org/2/reference/lexical_analysis.html#identifiers

tomhennigan:
> @tomhennigan, That is awesome! It is very similar to what we have in [GPflow](https://github.com/GPflow/GPflow) - `Parameterized` class. With this change in TensorFlow we will be able to serialize our objects (currently we have custom saver) and integrate them with Keras models.

Serialization (as in saving/restoring parameters) should work great (since we extend `Checkpointable`). I think you'll have a better time integrating with Keras `Model` if you extend `Layer` instead. We're hoping to make `tf.Module` a base class for Keras `Layer` but that is not included as part of this proposal (see [this comment](#issuecomment-455502533) for more context).

> Although, RFC could be extended with array-like modules, but don't confuse it with keras sequence. Keras sequence implies dependency between layers/modules, whereas array-like module is a different approach for organising submodules with indexing restricted to integer.
> 
> E.g. in [GPflow](https://github.com/GPflow/GPflow) we can construct a sum of the kernels (modules) and represent it as a super-kernel class or a combination kernel. Essentially, a combination kernel is a list of kernels, and when we call, let's say, __call__ method, it sums over sub-kernels results.

I think this should "just work". I tried creating a simple example, please let me know if I've completely misunderstood what you're trying to achieve:

```python
class MulKernel(tf.Module):
  def __init__(self, m, dtype=tf.float32, name=None):
    super(MulKernel, self).__init__(name=name)
    self.m = tf.Variable(tf.constant(m, dtype=dtype), name='m')

  def __call__(self, x):
    return self.m * 1

class ArrayModule(tf.Module):
  def __init__(self, kernels):
    super(ArrayModule, self).__init__()
    self.kernels = kernels

  def __call__(self, x):
    x = tf.convert_to_tensor(x, name='x')
    results = [kernel(x) for kernel in self.kernels]
    return tf.reduce_sum(results)

a = ArrayModule([MulKernel(i, name='mul_%d' % i) for i  in range(10)])
y = a(1.)
```

Which produces this rather pretty graph:

screenshot from 2019-01-18 11-09-26

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+### `tf.Variable` tracking
+
+It is very convenient to be able to collect all the variables for a given module
+and all of its dependencies. We achieve this in `tf.Module` using reflection to
+walk object properties. When walking if we encounter a nested structure (e.g. a
+`list`, `dict` etc) we will flatten it using `nest.flatten(value)` and evaluate
+each leaf. This means variables nested in properties will be found (e.g.
+variables in `list`s or `tuple`s on the module).
+
+This "walk" function is used to provide the following properties:
```
superbobry: Are there properties computed once per instance, or are they fully dynamic?

awav:
@tomhennigan

> I think you'll have a better time integrating with Keras Model if you extend Layer instead.

The finest abstraction is not actually deep learning model (is actually keras still a deep learning dedicated framework?), and not a model at all. The `tf.Module` perfectly fits our needs. Still, in some cases this abstract blocks can be included in layer compostion, e.g. head of the sequence is a CNN and output of CNN passes though a GP, https://arxiv.org/abs/1511.02222.

> If this RFC is accepted @alextp and I hope that we can make tf.keras.layers.Layer extend tf.Module (instead of Checkpointable as it does today). Module can help make the implementation of Layer simpler by implementing part of it's API (e.g. the Layer.variables property). We have not attempted to do this yet and I'm not sure if there would be any technical blockers.

I'd say, making keras layers depend on `tf.Module` should be one of the priorities for this RFC - unification of interfaces in TensorFlow. It diminishes a barrier for users, who inherit from `tf.Module` and want to take advantage of keras features.

> Serialization (as in saving/restoring parameters) should work great (since we extend Checkpointable)

That's great!

> I think this should "just work". I tried creating a simple example, please let me know if I've completely misunderstood what you're trying to achieve: ...

If `tf.Module` can find sub-module's in an attribute, which is actually a list, and properties like `variables`, `trainable_variables` and etc return correct objects then the problem is solved.

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+
+-   Even among the authors there existed different preferences for how to enable
+    this pattern, it seemed unlikely we could make a choice which would satisfy
+    the community.
+-   Not all modules want a `__call__` method (e.g. an `Optimizer`) and we got
+    strong feedback that it is annoying to ignore unused methods inherited from
+    your base class.
+-   Encouraging a lazy pattern might make it harder to create modules with
+    multiple forward methods (e.g. a VAE module might be implemented with an
+    `encode` and `decode` method with no `__call__`).
```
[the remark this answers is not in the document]
tomhennigan: We prototyped it such that you didn't have to implement `__call__` (e.g. `_call` was not `@abstractmethod`), but if users called your module then we throw `NotImplementedError`. Even this has downsides though, like introspection/generated docs making it look like you had a call method.

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+  def name_scope(self):
+    """Returns a `tf.name_scope` instance for this class.
+
+    NOTE: The module name scope includes all parent modules as determined
+    when this module was created.
+    """
+
+  @property
+  def variables(self):
+    """Returns a tuple of variables owned by this module and it's submodules.
```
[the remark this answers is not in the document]
tomhennigan: Thanks for the IRL discussion, I'll use "Collection" in the docstring and still return a tuple at least for now.

rfcs/20190117-tf-module.md
tomhennigan: Fully dynamic, trying to reason about when to invalidate the caches would be non-trivial.

tomhennigan:
@awav

> I'd say, making keras layers depend on `tf.Module` should be one of the priorities for this RFC - unification of interfaces in TensorFlow.

We've been mindful to design `tf.Module` in such a way that it could be a stateful container for a number of components in TensorFlow (not just `tf.keras`), but I think integrating it well with `tf.keras` requires a separate design/RFC process. I'm not planning to address that here.

> If `tf.Module` can find sub-module's in an attribute, which is actually a list, and properties like `variables`, `trainable_variables` and etc return correct objects then the problem is solved.

Yep, this works as you would expect (we use `tf.contrib.framework.nest.flatten(prop)` for each object property and walk all leaves).

rfcs/20190117-tf-module.md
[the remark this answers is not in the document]
gehring: @tomhennigan, I agree 100% with this decision. There is no need to bloat the `tf.Module` class more than is absolutely required. Since the goal of this class is to handle stateful DAGs of modules, it is best to focus efforts on that.

As pointed out in this doc, there are many use cases where we would want to handle stateful DAGs but don't require a `__call__` method. I see it the responsibility of other libraries/sub-package to extend the functionality of `tf.Module`. I find specialized classes much easier to extend and reason about.

rfcs/20190117-tf-module.md
gehring: I'm thinking the performance cost of having this dynamic should be well documented so that users know to cache the results of these calls if using eager and good performance is required. Though, if everything works seamlessly with `tf.function`, then maybe this wouldn't be necessary and encouraging users to use `tf.function` for better performance would be sufficient.

tomhennigan: Agreed re documentation, I'll add a note in the docstrings.

For a 100 linear layer sequential it takes 2ms to collect all 200 parameters (at least on my workstation). This could be worse for modules with deep hierarchies or large nests.

Assuming the call to `trainable_variables` (or whatever) is inside the `tf.function` this cost is only paid when the function is traced (~once).

rfcs/20190117-tf-module.md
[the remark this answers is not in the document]
josh11b: In general, returning a tuple makes the API a bit more efficient since you don't have to return a copy to avoid callers mutating your internal data structures.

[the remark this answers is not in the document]
tomhennigan: We agree returning a tuple is the right decision, but the signature/contract of the method should be less specific to allow this to change in future (e.g. if there is a faster immutable collection we might want to use that instead).

If we could include type hints in TF code I would go with the following:

```python
@property
def variables(self) -> typing.Collection[tf.Variable]:
  ...
```

martinwicke:
Or Iterable? We could adjust the docstring now.

Come 2020, we will drop Py2 support, and all this typing can actually
happen for real.

tomhennigan:
> Or Iterable? We could adjust the docstring now.

`Collection` is better since it indicates the returned value supports contains and len (iterable just indicates it has `__iter__`). Agreed updating the docstring is clearest.

> Come 2020, we will drop Py2 support, and all this typing can actually
> happen for real.

Can't wait 😄count me in on a fixit to go through and add types!

sguada:
+1 for this proposal, it strikes a great balance between abstraction and usability.

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+-   Variables created, owned or used by a module should be retrievable (e.g.
+    mod.variables should provide a collection of variables for mod and all of
+    its descendent modules).
+-   Which variables are tracked should be obvious (e.g. all object properties)
+    and "opt-out" rather than "opt-in" (e.g. the most obvious way to create and
+    reuse a variable should automatically be tracked).
+
+**`tf.Module` tracking**
+
+-   Modules should track dependencies (e.g. a sequential should provide access
```
lc0: Something that started on twitter -> https://twitter.com/lc0d3r/status/1090351739537145858

Since we are trying to also 
```
track dependencies (e.g. a sequential should provide access to its layers) and support checkpointing as a graph.
```

Would it make sense to think about the next step and have on the model level an approachable way to have access to DAG.

The idea is, once you inherit from tf.Module you can run something similar to `keras` `model.summary`  and even more exciting - `model.plot_model`.

Sorry for a bit confusing way to formulate this suggestion

tomhennigan: Hey @lc0, thanks for the suggestion! We've been thinking about something similar. I'm not sure if we should solve this inside tf.Module or if this is something we should leave to framework authors, what do you think?

FWIW the current module APIs expose a lot of useful information to do a modest job of this outside of Module:

```python
import tabulate

def weight_size(m):
  size = 0
  for v in m.variables:
    size += v.numpy().nbytes

  suffix = 'B'
  suffixes = ['KB', 'MB', 'GB', 'TB']
  while size > 1024 and suffixes:
    size /= 1024
    suffix = suffixes.pop(0)
  return '%d %s' % (size, suffix)

def summarise(module):
  rows = []
  for m in module.submodules:
    row = [m.name, m.__class__.__name__, weight_size(m)]
    rows.append(row)
  print(tabulate.tabulate(rows, headers=['attr', 'class', 'size'], tablefmt='grid'))
```

```
+----------+---------+--------+
| attr     | class   | size   |
+==========+=========+========+
| linear_0 | Linear  | 8 KB   |
+----------+---------+--------+
| linear_1 | Linear  | 4 MB   |
+----------+---------+--------+
| linear_2 | Linear  | 4 MB   |
+----------+---------+--------+
| linear_3 | Linear  | 40 KB  |
+----------+---------+--------+
```

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+    super(MyModule, self).__init__(name=name)
+    self.v = tf.Variable(1., name='v')
+
+>>> m = MyModule()
+>>> m.name
+'my_module'
+>>> m.name_scope.name
+'my_module/'
+>>> m.v.name
+'my_module/v:0'
```
pluskid: Are the variable names (with scope) attached upon the variable creation? An alternative might be the pytorch way that a variable does not know about its (parent's) scope, and the full scoped name is only constructed on the fly depending on which parent module are we dealing with (because a module can be re-used in different parent modules, and they might even be constructed externally and pass in via the constructor of a custom module).

I'm thinking about the following pseudocode example:
```python
class FunnyAutoEncoder(tf.Module):
  def __init__(self, encoder, decoder, name=None):
    super(...)
    self.encoder = encoder
    self.decoder = decoder

  def __call__(self, x):
    z = self.encoder(x)
    z2 = tf.funny_op(z)
    return self.decoder(z2)


encoder = MLP(...)
decoder = MLP(...)
ae = FunnyAutoEncoder(encoder, decoder)
```

then `ae.encoder.layerX.w.name` will probably not show anything related to being under an auto encoder? Instead, it will be just with a top-level `'mlp/'` scope?

I think a major scenario of having full scoped names available is to iterate through all the variables in a module, and do some inspection / processing to the variable according to the names. In this case, it might be more useful to have the (scoped) names reflect what role each variable is playing in the module.

However, I guess one drawback might be to detect potential duplication. For example, if one does this
```python
encoder = ResidualBlock(...)
decoder = encoder
ae = FunnyAutoEncoder(encoder, decoder)
```
then the same variables might be listed twice under different (scoped) names. But I guess this situation is rather rare, and it is probably to explicitly define a `FunnyAutoEncoderSharedEncoderDecoder` that has only one `encoder_and_decoder` field.

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+  ...   def __init__(self, in_features, output_features):
+  ...     super(Linear, self).__init__()
+  ...     # NOTE: We're creating variables in the ctor, but you can actually
+  ...     #       create them in any method (including __call__) you like.
+  ...     self.w = tf.Variable(
+  ...         tf.random.normal([input_features, output_features]), name='w')
+  ...     self.b = tf.Variable(tf.zeros([output_features]), name='b')
+  ...
+  ...   def __call__(self, x):
+  ...     x = tf.convert_to_tensor(x, name='x')
```
pluskid: I feel it is important to avoid having the users to call `tf.convert_to_tensor` every time they define their own `__call__` function. One might want to define implicit pre-call hook to do the conversion automatically. Or better approach might be to explicitly require the API to only accept `Tensor` (or nested tuple of tensors) as inputs, and ask user to do explicit conversion before calling from top level.

This is because many of the modules are probably to be used as building blocks in bigger architectures (e.g. residual blocks inside a resnet). They will not directly interact with the user most of the time, so there is probably no need for them to call `convert_to_tensor` every time, but it might be better to enforce the API to have consistent behavior.

gehring: While I agree with you that it would be good for any callable subclass of `tf.Module` to have some automatic conversion to tensor (e.g., in an implementation of Layer 2.0), this is beyond the scope of `tf.Module`. The `Dense` class example only serves to demonstrate how one would subclass `tf.Module`. Let me try to lay out what I understand the purpose of `tf.Module` to be and what the reason for its limited scope is. (I am not affiliated with the TF team)

After the shift to object-oriented storing of `tf.Variable`'s, variable reuse is done by storing variables within object attributes. It would be impractical and clunky to store all variables in a unique container object (which would essentially recreate TF 1.0's collections). To avoid this, we prefer to have variables contained within the objects that would use them, e.g., `kernel` and `bias` variables as attributes of the `Dense` instance that uses it.  While this new way of organizing variables has many advantages, it does make checkpointing and operating on the "global" variables more difficult (e.g., collecting variables for gradient descent).

By design, `tf.Module` is lightweight and focuses on addressing the challenges of managing nested stateful objects. Its sole purpose is to abstract away the DAG of modules for the purpose of operating on all relevant variables. This allows us to act like our top-level/root module explicitly contains all our variables. This is done by providing traversal methods which collect all the variables. As a bonus, `tf.Module` automatically handles the complexities of checkpointing when variables are allowed to be created before and after the checkpoint is loaded, e.g., hydration and instantiation of variables.

By limiting `tf.Module`'s responsibilities to managing nested stateful modules, we end up with code that is easy to understand, document and extend. As a result, any developer wishing to implement some stateful containers can easily do so without intricate knowledge of how best to manage "global" state by subclassing `tf.Module`. Since handling stateful DAGs of modules is a very common requirement, `tf.Module` could easily be used by most sub-packages and libraries with stateful operations, e.g., `tf.optimizers`, `tf.metrics`, `tf.estimator`. (Note, this specific proposal is arguing for the existence/inclusion of `tf.Module` in TF 2.0. The advantages of using `tf.Module` as a base class throughout tensorflow will likely be examined on case by case basis in a separate proposal.)

pluskid: @gehring Yes, I agree with you. I don't think doing automatic conversion is a good idea. But enforcing the inputs are Tensors are probably good (even only implicit as a statement in the documentation). Anyhow, this is a minor thing in the whole proposal.

rfcs/20190117-tf-module.md
tomhennigan: So currently the proposal is that name scopes are determined when the module is constructed and from that point it is fixed per module. For name based filtering you can do something like this:

```python
encoder = MLP(..., name='encoder')
decoder = MLP(..., name='decoder')
ae = FunnyAutoEncoder(encoder, decoder)

encoder_vars = encoder.variables
# .. or ..
encoder_vars = [v for v in ae.variables if v.name.startswith('encoder/')]
```

It's worth pointing out that you can also enter a name scope and build modules inside it if you want to group a set of modules that way:

```python
with tf.name_scope('encoder'):
  encoder = MLP(...)  # name_scope will be `encoder/mlp/`
decoder = MLP(..., name='decoder')  # name_scope := `decoder/`
ae = FunnyAutoEncoder(encoder, decoder)  # name_scope := `funny_auto_encoder/`
```

Is that flexible enough for your use cases? We've had a similar pattern in Sonnet and I've not heard any complaints from users about it, but perhaps there's something I'm missing!

pluskid: @tomhennigan Thanks! I'm not sure if it is because I'm too used to the other way of thinking. Let me state the use cases in my mind and we can see if it is easy to do in the current proposal. I think the main inflexibility of the current approach is that in order to have a meaningful namescope hierarchy, we can only construct the model from top-down according to the hierarchy. It will be difficult, for example, to construct some part of the component (e.g. the `encoder` from the example above) without knowing *beforehand* what that component is going to be used for (i.e. in your example, you explicitly pass the `name='encoder'` argument, indicating you know that it is going to be used as an encoder).

This problem might be more pronounced when one want to write a generic library of commonly used (sub)nets. For example, in library1 (say Sonnet2), I have a `Module` that defines a default ResNet implementation. In library2, which is a general RL training framework, that allow one to pass in a `Module` to be used for each specific component (e.g. the Vision part of the agent). Now as the user, if I want to use the ResNet in the vision part of the agent, I might need to use some hack to figure out what proper `name` to pass into when constructing the ResNet to make it consistent with the naming hierarchy in the agent model. Moreover, if the agent is grouping all its vision part code under the name scope `agent/vision`, then I cannot pass `agent/vision` as `name=` parameter to `ResNet` because it is not a proper python variable name.

My guess that this is not a problem in Sonnet(v1) is that Sonnet(v1) is still graph based library with explicit `build()` function. In other words, when you create the Python object, the underlying graph building (therefore the naming and scoping) has not been called yet. Therefore, one can also change the name scoping *when* the build is actually being called. Here is some pseudo-code of my understanding of what might happen in Sonnet V1
```python
class Agent(...):
  def __init__(self, vision_net, ...):
    self.vision_net = vision_net
    # ....

  def build(self, ...):
    with tf.name_scope(self.name + '/vision'):
      self.vision_net.build(...)
```
On the other hand, with the current proposal, users could (and it seems recommended) create variable in the python object `__init__` function, which will attach name to the variable upon creation. Therefore, one need to know the exact name of the parent *hierarchy* upon creating that python object. Once the `Module` object is created, the parent module can no longer re-wire its name scope as in the case of Sonnet V1. Is my understanding correct here?

rfcs/20190117-tf-module.md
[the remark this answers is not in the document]
chr1sj0nes: Perhaps `typing.Sequence` would be better here? Importantly, the returned collection has a well-defined, deterministic ordering.

rfcs/20190117-tf-module.md
awav: I'm strongly against auto-conversion to tensors and agree with @tomhennigan, it is up to a scientists, engineers or anyone else, what to do with non tensor types.

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+  def __call__(cls, *args, **kwargs):
+    """Creates new module instances.
+
+    To allow other modules/variables to be created in the constructor of a
+    module (and be named appropriately) we enter the module name scope inside
+    the base (Module) constructor. We close the name scope in the metaclass
+    before returning the new instance.
+
+    >>> class MyModule(tf.Module):
+    ...   def __init__(self):
```
brilee: To me, it seems more likely that users will end up using a dependency-injection style, with my_other_module being passed in as an argument to `__init__`. This way, if you want to do something like share a set of weights across multiple parts of a neural network, then you would just pass the same module instance to each of those parts.

In this scenario, MyOtherModule() would be created before MyModule() is created. I'm not sure the `@name_scope` property would still work in a DI world.

rfcs/20190117-tf-module.md
brilee: I agree with @pluskid that dependency injection should at least be possible. My preference is for DI to be the recommended way to build up modules, but I'd understand if others disagreed.

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+
+## Requirements
+
+### Functional
+
+**Naming `tf.Variable`s (and `tf.Operation`s)**
+
+-   Each module instance must have a name scope. This name should include the
+    name scope from when the module is constructed.
+-   Every method on the module must enter this name scope before calling user
```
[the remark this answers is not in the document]
tomhennigan: In eager op names are not a thing so that's good.

For `tf.function` names are uniqueified within the function body, so you would only get an _N suffix if there were actually two relu's within **the same** module (or you called your module multiple times sequentially).

As a nit, scope names are re-entrant (as in you can do `with tf.name_scope("foo/"): ...` as many times as you like) but op names need to be unique.

Example: https://gist.github.com/tomhennigan/a183a6570d2a6fb9bce08758ee2d78b6

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,634 @@
...
+class Module(six.with_metaclass(ModuleMetaclass, tf.checkpoint.Checkpointable)):
+  """Base neural network module class.
+
+  A module is a named container for `tf.Variable`s, other `tf.Module`s and
+  functions which apply to user input. For example a dense layer in a neural
+  network might be implemented as a `tf.Module`:
+
+  >>> class Dense(tf.Module):
+  ...   def __init__(self, in_features, output_features):
+  ...     super(Linear, self).__init__()
```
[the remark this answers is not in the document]
tomhennigan: Updated, thanks!!

rfcs/20190117-tf-module.md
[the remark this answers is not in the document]
brilee: Thanks for the correction. I see now that this works, although I'm still a bit confused at where the _1 suffix comes from.
```
@tf.function
def f(x):
  name_scope = tf.name_scope('blah')
  with name_scope:
    with name_scope:
      return x + 1

g = f.get_concrete_function(tf.TensorSpec([1, 64], tf.float32)).graph
for node in g.as_graph_def().node:
  print(node.name)
  
...
x
blah_1/add/y
blah_1/add
Identity
```

rfcs/20190117-tf-module.md
```diff
@@ -250,19 +237,6 @@ class Module(six.with_metaclass(ModuleMetaclass, tf.checkpoint.Checkpointable)):
...
-    module and all it's submodules recursively.
-
-    Returns:
-      A collection of variables which are attributes of the current module. Will
-      yield variables inside nested structures (lists etc) but not in other
-      modules.
-    """
-
   @property
   def submodules(self):
```
gehring: @tomhennigan Is this done depth first or breadth first? I feel like it could be a good thing for `tf.Module` to offer a way to infer the depth of submodules/variables. Without `owned_*`, what would be the best way for a user to do this without re-implementing `tf.Module`'s logic?

tomhennigan: Breadth first.

So there are two options (frustratingly some recent commits haven't been sync'd to GitHub so I can't just point you at the code!).

Firstly we've made it really easy for subclasses to implement this if the want by exposing `_flatten`:

```python
class MyModule(tf.Module):
  @property
  def owned_submodules(self):
    return tuple(self._flatten(recursive=False, predicate=lambda v: isinstance(v, tf.Module)))
```

I also discussed with brilee@ and pluskid@ earlier about their feedback on variable names. We were aligned that what they both wanted was not really `Variable.name`, but rather the path to all variables (e.g. allowing the same variable instance to be reachable via multiple paths). We're planning to add an option to `_flatten` to allow flattening with paths. So for example in the case where you have:

```python
encoder = Encoder()
decoder = encoder
auto = Auto(encoder, decoder)
```

You can implement `variables_and_paths` in one line with `self._flatten(..., with_path=True)`:

```python
>>> auto.variables_and_paths()
{"encoder.layers[0].w": <tf.Variable ...>,
 "encoder.layers[1].w": <tf.Variable ...>, 
 "decoder.layers[0].w": <tf.Variable ...>,
 "decoder.layers[1].w": <tf.Variable ...>}
```

It seems to me like this is another way of representing depth. WDYT?

tomhennigan: Actually the CL has been pushed 😄, `_flatten` is in tensorflow/tensorflow@5076adf6

gehring: I didn't have a particular use case in mind but I believe it is likely for someone to require that type of info (if not simply for human readability). The `with_path` solution is lightweight and should pretty much cover any uses cases related to the topology of the variables/sub-modules. I'm 100% on board, nice job!

rfcs/20190117-tf-module.md
tomhennigan: Thanks for the IRL discussions @pluskid, @brilee, @superbobry and @chr1sj0nes.

To summarize our discussions there are many cases where there are multiple paths to weights in a model (e.g. `model.encoder.w` might be the same instance as `model.decoder.w`) and this is somewhat confusing since in TensorFlow variables have a single, immutable name (e.g. `w`'s name would be `encoder/w:0` in both cases if encoder created it).

Perhaps the "right" solution is to remove the name from variables in Python, and instead use Python variable names to "name" variables. Sadly this is not in scope for TensorFlow 2, and we agreed that the current strategy of variable naming at creation time was OK for simple cases and better than not naming.

We discussed that the python variables or object structure can be used instead to associate a path which each parameter and in the above case you would have two paths to `w`. We agreed that it would be convenient to have some easy way to traverse the structure of arbitrary `Module`s in a way that have some simple protocol (e.g. somewhere inside the model there is a property called `encoder` and somewhere inside the model there is a property called `decoder`) to filter these paths.

In tensorflow/tensorflow@1d5fd9219 I added `with_path` to Module's `_flatten` method. This allows you to do exactly that. For example you can find paths in some arbitrary model to all `tf.Variable`s inside a (possibly deeply nested) `Module` in a property called `encoder`:

```python
>>> class MyModule(tf.Module):
...   @property
...   def state_dict(self):
...     return dict(self._flatten(
...         predicate=lambda v: isinstance(v, tf.Variable), with_path=True))

>>> mod = MyModule()
>>> mod.encoder = Encoder()
>>> mod.decoder = mod.encoder

>>> mod.state_dict
{('encoder', 'w'), <tf.Variable: ...>,
 ('decoder', 'w'), <tf.Variable: ...>}

>>> {k: v for k, v in mod.state_dict.items() if 'encoder' in k}
{('encoder', 'w'), <tf.Variable: ...>}
```

I hope that helps :smile:. h/t @gehring who was asking about how to determine "depth" in another comment. I think this `with_path` option to `_flatten` (or the `recursive` flag) would help solve that too.

rfcs/20190117-tf-module.md
tomhennigan: I think [this answer](https://github.com/tensorflow/community/pull/56#discussion_r255048762) covers this case too? Please let me know if I'm wrong.

rfcs/20190117-tf-module.md
tomhennigan: FYI `with_path` is in tensorflow/tensorflow@1d5fd921 :smile:

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,629 @@
...
+
+>>> p.name
+'parent_module'
+>>> p.name_scope.name
+'parent_module/'
+>>> p.v.name
+'parent_module/v:0'
+
+>>> p.child_module.name
+'child_module'
```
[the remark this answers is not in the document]
tomhennigan: Whoops! Thanks have updated this 😄

rfcs/20190117-tf-module.md
brilee: Yep. Thanks for taking the time to answer our questions!

rfcs/20190117-tf-module.md
```diff
@@ -0,0 +1,629 @@
...
+    """
+
+  @property
+  def trainable_variables(self):
+    """Returns a collection of variables owned by this module and it's submodules.
+
+    Returns:
+      A collection of variables for the current module (sorted by attribute name)
+      followed by variables from all submodules recursively (depth first).
+    """
```
MarkDaoust: I'm a bit late to the party, but I don't see this in the discussion anywhere:

Keras layers and models implement a `trainable` flag as well.
When set to False, none of the variables in the module, or submodules are returned in by trainable_variables.

The current implementation of module does not recognize a module-level trainable flag, it's not clear if that's intentional, or an omission.

Looking at the implementation... it would require a separate [predicate for recursion](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/module/module.py#L419).

This seems like a minor change that could reduce surprises and increase compatibility. Especially with the plan to make `layer` and `model` sub-classes of module.

WDYT?

tomhennigan: This is intentional. Module uses the immutable "training" bit on `tf.Variable` to determine which variables are trainable. If you want to implement layer freezing like Keras does for a `Module` subclass you can do:

```python
class KerasLike(tf.Module):

  def __init__(self, trainable=True, name=None):
    super(KerasLike, self).__init__(name=name)
    self.trainable = trainable

  @property
  def trainable_variables(self):
    if not self.trainable:
      return ()
    return super(KerasLike, self).trainable_variables
```

MarkDaoust: Great. Thanks for the clarifications.

rfcs/20190117-tf-module.md
episodeyang: The table seems excessive. Here is something with less chrome:

```
 attr      class    size   
+========+=======+======+
 linear_0  Linear   8 KB   
 linear_1  Linear   4 MB   
 linear_2  Linear   4 MB   
 linear_3  Linear   40 KB  
```

PR #59, opened 2019-01-28
Add Bryan to SIG-IO release team

BryanCutler:
This change adds Bryan to the SIG-IO release team with GitHub and PyPi handles.


BryanCutler:
cc @yongtang @ewilderj

PR #67, opened 2019-02-08
RFC: Replace SWIG with pybind11

av8ramit:
**The review period for this RFC will end on 2019-02-26**

# RFC: Replace SWIG with pybind11

| **Author**    | amitpatankar@google.com                              |
| **Sponsor**   | wicke@google.com, gunan@google.com, jkline@google.com|
| **Updated**   | 2019-02-09                                           |

## Summary

[SWIG](http://www.swig.org/tutorial.html) is an interface compiler that connects code written in C++ with our Python API. Currently we are using SWIG across all of TensorFlow. We would like to deprecate SWIG and transition over to using pybind11. 

This document describes a system to improve:

1.  code readability
2.  build times
3.  binary size
4.  performance of the Python API

### DIFF
--- a/rfcs/20190208-pybind11.md
+++ b/rfcs/20190208-pybind11.md
@@ -1,6 +1,6 @@
 # Replace SWIG with pybind11
 
-| Status        | Proposed                                             |
+| Status        | Accepted                                             |
 :-------------- |:---------------------------------------------------- |
 | **Author**    | amitpatankar@google.com                              |
 | **Sponsor**   | wicke@google.com, gunan@google.com, jkline@google.com|
--- a/rfcs/20190208-pybind11.md
+++ b/rfcs/20190208-pybind11.md
@@ -1,0 +1,327 @@
+# Replace SWIG with pybind11
+
+| Status        | Proposed                                             |
+:-------------- |:---------------------------------------------------- |
+| **Author**    | amitpatankar@google.com                              |
+| **Sponsor**   | wicke@google.com, gunan@google.com, jkline@google.com|
+| **Updated**   | 2019-02-09                                           |
+
+
+
+## Objective
+
+[SWIG](http://www.swig.org/tutorial.html) is an interface compiler that connects code written in C++ with our Python API. Currently we are using SWIG across all of TensorFlow. We would like to deprecate SWIG and transition over to using pybind11. 
+
+This document describes a system to improve:
+
+1.  code readability
+2.  build times
+3.  binary size
+4.  performance of the Python API
+
+
+## Motivation
+
+
+### What is pybind11?
+
+[Pybind11](https://github.com/pybind/pybind11) is a header-only library that exposes C++ functions and classes to Python to create Python bindings of existing C++ code. Pybind11 is lightweight, and unlike other libraries that serve a similar purpose, it is purely focussed on binding. Pybind11 is relatively new, but is rapidly growing in contributors and adopters. It's an adaptation of [Boost.Python](https://www.boost.org/doc/libs/1_66_0/libs/python/doc/html/index.html) which is quite reputable, but has a lot of extra libraries. Pybind11 is basically the only the binding generation portions of Boost.Python.
+
+
+### Cons with SWIG
+
+1. SWIG auto-generated code is not optimal for performance.
+2.  SWIG needs the swig toolchain.
+3.  SWIG requires .i or .swig files which require knowledge of SWIG directives.
+    1.  We've changed .swig to .i files to get around readability issues
+4.  SWIG combined with C++ is relatively complicated to understand.
+
+
+### Why not CLIF?
+
+1.  The CLIF wrapping directives are more complex [[Link](https://github.com/google/clif/blob/master/clif/python/README.md#api-specification-language)]
+  1.  Need a separate .cliff file
+    2.  Similar to SWIG, adds a toolchain dependency beyond just the C++ compiler (and adds another dependency on a different subset of LLVM)
+2.  Most literature is optimized for cmake as opposed to bazel
+3.  Fewer developers and adopters in open source
+    1.  [No activity](https://github.com/google/clif) for nearly a year
+
+
+### Pro/Cons with pybind11
+
+
+| Pros: | Cons: |
+|-------|-------|
+| No need to link against any additional libraries|Less support|
+|Function signatures are precomputed at compile time which leads to smaller binaries|Our cases may be more complicated than their examples|
+|No auto-generated code|Heavy reliance on another third party library|
+|Much simpler and easier to debug||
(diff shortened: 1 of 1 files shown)


rfcs/20190208-pybind11.md
```diff
@@ -0,0 +1,327 @@
...
+        compatible_with = compatible_with,
+    )
+```
+
+### Issues
+
+
+
+1.  PyObject pointers are not inherent to pybind11 and require some minor code modification. [[GitHub Issue Link](https://github.com/pybind/pybind11/issues/1201#issue-278504106)]. As you notice in diagram 2 of the rollout we have to create a small wrapper function around the function. Pybind11 cannot cast a python object to `PyObject*` so you need to provide a `pybind11::handle` reference and cast it internally. We could potentially modify the IsSequence function directly, but that would require additional unnecessary modification in the .cc file. 
+2.  Complicated use cases where we export weird classes are not properly documented or seen before. We unfortunately don't know what we don't know when it comes to our migrations. [Other users](https://community.lsst.org/t/using-pybind11-instead-of-swig-to-wrap-c-code/1096) have also complained about the more complicated use cases.
```
superbobry: Could you give an example of a problematic case in TensorFlow? Also note that the LSST post could be out of date (published in 2016), e.g. `pybind11` does support sequence types now via `pybind11::sequence`.

rfcs/20190208-pybind11.md
```diff
@@ -0,0 +1,327 @@
...
+## Design Proposal
+
+We should deprecate SWIG and replace the wrapper with pybind11. It would be impractical to create a single cl where we transfer everything so we will do a transition in place. Each [.i file](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/tensorflow.i) is a swig file that has directives to wrap and export various C++ classes and functions. There are 60 files across all of TensorFlow. On average there are roughly 20 symbols per file. We can do a rolling transition where we export certain files twice. We can export some symbols with SWIG and the remaining ones with pybind. 
+
+
+### Impact
+
+We can get rid of roughly 60 bloated SWIG files by transitioning to pybind11. We can also make documentation and exporting functions much easier. We can save countless hours on adding classes and ops by making exporting them to Python extremely easy. 
+
+1.  Faster build times by decreasing code generation and simply exposing the code directly.
```
superbobry: Interestingly, the build times reported in the LSST blog post (see the "Issues" section below) are comparable between pybind11/SWIG versions. Do you have another reference where the build times are consistently faster?

byronyi:
What about other bindings (e.g. Java) that could, in theory, use SWIG but not pybind11?

perfinion:
Can you add a comparison to Cython as well? Other projects ive seen recently that moved off SWIG went to Cython so its worth looking at. https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html

Also TF supports many languages, will those work with pybind11? or will there be a C api that they interface with?

bhack:
As a side note. As in tf 2.0 Keras will be one of the main promoted interfaces and in production/inference many developers need to use c++ code in c++ projects do you think that you will improve/default solutions like http://www.bitbionic.com/2017/08/18/run-your-keras-models-in-c-tensorflow/?

bhack:
I think this need to be re-evaluated with TF 2.0 as the Pythonic nature of the Keras API:
https://github.com/keras-team/keras/issues/4491

skye:
@byronyi @perfinion @bhack I believe we only use SWIG for TF's Python library. Other language bindings use different mechanisms to wrap the [C API](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/c_api.h). So other languages are outside the scope of this proposal. (@av8ramit correct me if I'm wrong.)

av8ramit:
@skye yes that is correct! This proposal is only to replace the SWIG generated bindings for [`pywrap_tensorflow`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/BUILD#L4227). I will update the description accordingly.

bhack:
Yes I understand but what about wrapping an api that is designed in python like the new Keras roles in 2.0. 
So I am talking about the inverse problem c++ call on a new main API like Keras that it is python.
So I suppose that you are not introducing pybind11 to use embedding and that you think that a new semi-core API like tf.Keras will be only confined in python right? https://pybind11.readthedocs.io/en/stable/advanced/embedding.html

bhack:
To summarize tf.keras will be so central in TF 2.0 but it will never enter in C API and it will be not available other than python. 
In 2.0 we still need two use two different api in other language bindings vs python.

bhack:
@skye Ok, but I think that TF team has done its own strategic evaluation to what to expose or not expose to other language bindings (including c++) in the TF 2.0 API refactor. 
Here the clarification that pybind11 is only introduced -> Python it is enough.

martinwicke:
We have a long-standing desire to pull more of the API into C++ (and C!). That is a prerequisite for exposing the functionality in other languages (without rewriting it entirely; an exception is the R bindings which actually call the Python).

This RFC is completely orthogonal to that effort. We would like to change how the *Python* bindings are created for the reasons outlined in the RFC.

Neither effort is tied to TF2.0, which is mainly about the Python API (which is also the only API for which we make semver guarantees), so anything related to other languages can proceed in independently and in parallel.

bhack:
@martinwicke But the python dominant API shift (tf.keras marketing in 2.0) will impact the evaluation of what to have in C++/C and so it is strictly related to TF2.0 and production (if for you production have a special regard for C/C++).

I think that with tf.keras promotion the gap it is becoming larger with bindings and this is a TF2.0 topic but out the boundaries of this ticket.

rfcs/20190208-pybind11.md
av8ramit: I don't have an official metrics. I will definitely document the differences as I roll out implementation. We are making the assumption based on getting rid of a few dependency links in the middle (Swig generated python file, pywrap_tensorflow.py) between C++ and the python function call.

rfcs/20190208-pybind11.md
av8ramit: Yes you are right, that post is out of date. Fortunately from the minor conversions I've made I haven't seen any roadblocking issues. This was a more generic comment that could impact the roll-out timeline.

av8ramit:
@perfinion with regards to Cython, I don't have a performance argument for you, but one of the most appealing things about pybind11 is no longer requiring an [intermediate file](https://cython.readthedocs.io/en/latest/src/userguide/wrapping_CPlusPlus.html#add-public-attributes) like we currently have with SWIG. I haven't used Cython before so happy to continue this discussion. pybind11 is lightweight and it's sole intention fits relatively well with our goals.

PR #69, opened 2019-02-26
RFC: TF on Demand Project

hyeygit:
*Comment period is open until 2019-03-13* 

| **Author(s)** | Gunhan Gulsoy (gunan@google.com), Hye Soo Yang (hyey@google.com) |
| **Sponsor**   | Gunhan Gulsoy (gunan@google.com)                |
| **Collaborators** | Christoph Goern (goern@redhat.com), Subin Modeel (smodeel@redhat.com)         |
| **Updated**   | 2019-02-25                                           |

## Objective

This document proposes a system to build optimized binaries in the cloud and deliver the artifacts to users of TensorFlow. With this system, we aim to achieve the following:

*   Hide the complexity of the build system of TF from beginner and intermediate users.
*   Improve the out-of-the-box performance for users.
*   Build on the strength of TensorFlow's community partners.

### DIFF
--- a/rfcs/20190225-tf-on-demand.md
+++ b/rfcs/20190225-tf-on-demand.md
@@ -1,6 +1,6 @@
 # TensorFlow On Demand
 
-| Status        | Proposed      |
+| Status        | Accepted      |
 :-------------- |:---------------------------------------------------- |
 | **Author(s)** | Gunhan Gulsoy (gunan@google.com), Hye Soo Yang (hyey@google.com) |
 | **Sponsor**   | Gunhan Gulsoy (gunan@google.com)                |
--- a/rfcs/20190225-tf-on-demand.md
+++ b/rfcs/20190225-tf-on-demand.md
@@ -1,0 +1,191 @@
+# TensorFlow On Demand
+
+| Status        | Proposed      |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | Gunhan Gulsoy (gunan@google.com), Hye Soo Yang (hyey@google.com) |
+| **Sponsor**   | Gunhan Gulsoy (gunan@google.com)                |
+| **Collaborators** | Christoph Goern (goern@redhat.com), Subin Modeel (smodeel@redhat.com)         |
+| **Updated**   | 2019-02-25                                           |
+
+## Objective
+
+This document proposes a system to build optimized binaries in the cloud and deliver the artifacts to users of TensorFlow. With this system, we aim to achieve the following:
+
+*   Hide the complexity of the build system of TF from beginner and intermediate users.
+*   Improve the out-of-the-box performance for users.
+*   Build on the strength of TensorFlow's community partners.
+
+## Motivation
+
+### Overview
+
+TF build is difficult and the friction present during building phase is apparent in our user experiences. For a successful build, users needs to be aware of requirements and configurations that they might not be too familiar with. This can be extremely challenging for beginner and intermediate users.
+
+Currently, TF python release artifacts are hosted on PyPI as a single source of TF binaries for download. However, PyPI is quite limited in capability; it is unable to recognize and/or deploy on user machines.
+
+Historically, this has been a problem when trying to satisfy the following requirements for our artifacts:
+
+*   **Portability**: The artifacts should be able to run on as many platforms as possible.
+*   **Performance**: The artifacts should run as fast as possible.
+*   **Size**: The artifacts should be as small as possible in size.
+
+To achieve the above, we propose a system that takes inputs from the user, build artifacts based on the inputs and send the output back to the user. We are proposing two user endpoints to the system:
+
+1.  Command line tool (python)
+1.  Web interface
+
+An overview of the system:
+
+![drawing](https://docs.google.com/drawings/d/11jAVBtR4nV4bkDVW1WrHhLXdzX2ZouOPwsztW2hZQz4/export/png)
+
+A system flow chart summarizing the ordering of events for all use cases of the system:
+
+![drawing](https://docs.google.com/drawings/d/1lSSHaYktst8MNTqF860cPlkfbgc3Ft_t_8PWzH0VkW4/export/png)
+
+
+### Collaboration with Red Hat
+
+Red Hat already built a system to build and deliver TF on the cloud. Reference artifacts released by Red Hat are listed below:
+
+*   [tensorflow-build-s2i ](https://github.com/thoth-station/tensorflow-build-s2i)
+*   [Tensorflow-release-job](https://github.com/thoth-station/tensorflow-release-job)
+*   [tensorflow-serving-build](https://github.com/thoth-station/tensorflow-serving-build)
+
+We propose collaborating with the Red Hat team and leveraging the team's open-source software as much as possible. The server side of the project, in particular, can have concentrated involvement from the team given the backend GKE cluster that already has a lot of functionality built and available. (Please refer to the "Server" block in the diagram above for details on the server-side event flow.) Based on Additional support and functionality will be built around and on top of it to match what TensorFlow currently supports and should support moving forward.
+
+## Design Proposal
+
+### Front-end
+
+**Web UI (download.tensorflow.org)**
+
+We propose a simple web interface that allows users to manually enter necessary system specs and desired build options for requesting a custom TF binary. The UI will be straightforward and easy-to-use. A sample mock up UI is shown below:
+
+![drawing](https://docs.google.com/drawings/d/1Krze2no7zjfqe7nldOm-ArECOGXVVjgtk0VMXCabzkw/export/png)
+
+Once all fields are filled in, the system backend will check if the corresponding binary has already been built and present in cache. If it exists, then it will provide the user a URL to the binary for download. If it does not exist, then it will ask the user for an email address to receive a link for downloading the newly built binary.
+
+Before sending out the binary, the system will first check whether the binary is supported (or unsupported) by going through what is being tested in CI. It will then inform the users accordingly. 
+
+**Command Line Tool (TF Downloader Binary)**
+
+We propose a simple binary which will detect and fill out most of the inputs that the Web UI requires. It will then send the request to the backend to build. A sample execution (not final) is shown below:
+
+```
+> python tfdownloader.py
+Downloader will now detect your system:
+	- Detecting CPU…….  Found Intel Core i7-8700K
+	- Detecting GPU……. Found NVIDIA GTX 980
+	- Detecting CUDA…… Found 9.2
+	- Detecting cuDNN….. Found 7.4
+	- Detecting Distribution……. Found Ubuntu 18.04
+
+Requesting TF build with options:
+    CPU options: -mavx2
+    GPU enabled: yes
+    CUDA version: 9.2
+    cuDNN version: 7.4
+    CUDA compute capability: 5.2
+    GCC version: 7.3
+```
+
+
+It is possible that the system this script is run on may be incompatible for running TF. Such incapabilities and what the script is able to do can be categorized into two failure modes:
+
+
+
+*   *Hard Constraint Failures*
+    *   32-bit OS
+    *   GPU too old (Cuda Compute Capability older than 3.0)
+*   *Soft Constraint Failures*
+    *   CUDA version too old
+    *   cuDNN missing
+    *   Other runtime libraries missing or too old (libstdc++)
+    *   Unsupported python version found
+
+In hard constraint failure cases, the tool will point users to cloud options. In soft constraint failure cases, the tool will offer alternative build configurations (CPU only) or request users to install the missing software.
+
+We propose distributing this binary through PyPI. However, please note that binary distribution method is still under discussion.
+
+### Back-end
+
+**TF Builder Service**
+
+TF builder service will require a proto with build options. This proto will define all options the system recognizes to build TF. Once the request is received, the system will check the cache to see if we already have such a package built. If yes, then it will send back the URL for downloading this package. A sample proto (not final) is shown below:
+
+```
+proto build_options {
+   string version = 0;
+   enum CpuOptions {
+       SSE3 = 0;
+       SSE4 = 1;
+       SSE4_1 = 2;
+       SSE4_A = 3;
+       SSE4_2 = 4;
+       AVX = 5;
+       AVX2 = 6;
+       AVX512F = 7;
+   }
+   repeated CpuOptions cpu_options = 1;
+
+   // CUDA options
+   bool cuda_enabled = 2;
+   enum NvidiaGPUGeneration {
+       KEPLER = 0;
+       MAXWELL = 1;
+       PASCAL = 2;
+       VOLTA = 3;
+       TURING = 4;
+   }
+   NvidiaGPUGeneration gpu_generation = 3;
+   string CUDA_version = 4;
+   string cuDNN version = 5;
+……
+   // Free formed string of options to append to bazel.
+   string extra_options = 100;
+}
+```
+
+If no such package exists in cache, then the system will execute the following commands with the appropriate flags and environment variables to newly build it.
+
+```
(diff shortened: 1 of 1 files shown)


rfcs/20190225-tf-on-demand.md
```diff
@@ -0,0 +1,191 @@
...
+We propose a simple binary which will detect and fill out most of the inputs that the Web UI requires. It will then send the request to the backend to build. A sample execution (not final) is shown below:
+
+```
+> python tfdownloader.py
+Downloader will now detect your system:
+	- Detecting CPU…….  Found Intel Core i7-8700K
+	- Detecting GPU……. Found NVIDIA GTX 980
+	- Detecting CUDA…… Found 9.2
+	- Detecting cuDNN….. Found 7.4
+	- Detecting Distribution……. Found Ubuntu 18.04
```
fridex: Some of these be found at https://github.com/thoth-station/thamos/blob/master/thamos/discover.py#L33

rfcs/20190225-tf-on-demand.md
```diff
@@ -0,0 +1,191 @@
...
+……
+   // Free formed string of options to append to bazel.
+   string extra_options = 100;
+}
+```
+
+If no such package exists in cache, then the system will execute the following commands with the appropriate flags and environment variables to newly build it.
+
+```
+git checkout <tag>
```
justin1121: Hi,

Curious to know if this `<tag>` could reference any branch known to the tensorflow repo (maybe impossible to reach forks?) or will be limited to released tensorflow versions or release candidates? Think it would pretty interesting to be able to build custom tensorflow builds based of a fork submitted for consideration to the tensorflow github repo.

rfcs/20190225-tf-on-demand.md
```diff
@@ -0,0 +1,191 @@
...
+yes "" | ./configure
+bazel build tensorflow/tools/pip_package:build_pip_package
+./bazel-bin/tensorflow/tools/pip_package/build_pip_package
+```
+
+Once the build is complete, the package will be stored in the cache before sending out a link to the user for download.
+
+**Package Cache**
+
+We will design the package cache as a simple GCS bucket supported by a simple database. The database can have the following schema:
```
fridex: How long will the built artifacts be available for download? Will it be possible to list available builds (e.g. have listing of `gs://tensorflow-artifacts/on-demand/`?

IMHO its worth to consider keeping build options somewhere handy (e.g. publish well defined proto or JSON/YAML alongside with the resulting wheel files) so that a user can see build configuration of a wheel file before downloading it or cross-checking it. This is also handy for automated systems that would like to consume such metadata when using built wheel files.

Another option (instead of creating a metadata files) is to keep prefix consisting of build configuration options  as we did originally [here](http://tensorflow.pypi.thoth-station.ninja/index/) the issue is in maintaining such prefix and keeping it parse-able.

rfcs/20190225-tf-on-demand.md
```diff
@@ -0,0 +1,191 @@
...
+
+```
+PackageCacheDataStore {
+   // This is the primary key to our datastore.
+   // As build_options is such a complex data type,
+   // this will be the the build_options fed through a hash function
+   string build_options_key;
+
+   // The path to the artifact. Will look like:
+   // gs://tensorflow-artifacts/on-demand/<hash>/packagename.whl
```
fridex: It would be great to ffollow wheel file name convention as defined in [PEP standards around wheel files](https://www.python.org/dev/peps/pep-0491/#id9). The name of file is relevant when installing a wheel file and Python's toolchain uses it during installation to verify if the given wheel file fits platform. One can then directly install the downloaded wheel file, also installing via HTTPS is an option (if the listing will support it).

With the above, wheel files will carry tensorflow version. It's worth to consider to put versions built with the same configuration options into same prefix (the description makes this not clear - the build system would override `packagename.whl` if a different tensorflow version is built as only hash of `build_options_key` distinguishes these builds). This way one can use the proposed system as a TensorFlow-specific PyPI - by providing URL that would like like `https://<path-to-gs>/tensorflow-artifacts/on-demand/<hash>`, when doing ``pip install --index-url <URL>`` or in other tools that try to deal with multi-package source index configuration (such as [Pipenv](https://pipenv.readthedocs.io/en/latest/advanced/#specifying-package-indexes)). Package updates/new version releases for specific build configuration options would work out-of-the-box.

rfcs/20190225-tf-on-demand.md
```diff
@@ -0,0 +1,191 @@
...
+```
+
+
+It is possible that the system this script is run on may be incompatible for running TF. Such incapabilities and what the script is able to do can be categorized into two failure modes:
+
+
+
+*   *Hard Constraint Failures*
+    *   32-bit OS
+    *   GPU too old (Cuda Compute Capability older than 3.0)
```
alextp: People with old GPUs can still build TF for CPUs, right?

rfcs/20190225-tf-on-demand.md
```diff
@@ -0,0 +1,191 @@
...
+*   Improve the out-of-the-box performance for users.
+*   Build on the strength of TensorFlow's community partners.
+
+## Motivation
+
+### Overview
+
+TF build is difficult and the friction present during building phase is apparent in our user experiences. For a successful build, users needs to be aware of requirements and configurations that they might not be too familiar with. This can be extremely challenging for beginner and intermediate users.
+
+Currently, TF python release artifacts are hosted on PyPI as a single source of TF binaries for download. However, PyPI is quite limited in capability; it is unable to recognize and/or deploy on user machines.
```
frreiss: Does this proposal mean that TensorFlow binaries will no longer be released on PyPI?

rfcs/20190225-tf-on-demand.md
martinwicke: Always. And <3.0 was never supported for GPU anyway.

rfcs/20190225-tf-on-demand.md
jackwish: Guess not.

> Does this proposal mean that TensorFlow binaries will no longer be released on PyPI?

Guess not.

rfcs/20190225-tf-on-demand.md
```diff
@@ -0,0 +1,191 @@
...
+1.  Command line tool (python)
+1.  Web interface
+
+An overview of the system:
+
+![drawing](https://docs.google.com/drawings/d/11jAVBtR4nV4bkDVW1WrHhLXdzX2ZouOPwsztW2hZQz4/export/png)
+
+A system flow chart summarizing the ordering of events for all use cases of the system:
+
+![drawing](https://docs.google.com/drawings/d/1lSSHaYktst8MNTqF860cPlkfbgc3Ft_t_8PWzH0VkW4/export/png)
```
jackwish: The diagrams seems of low resolution...

byronyi:
Would it be possible for the TF team to take a step back and encourage the downstream vendors to come up with their own distros, like RHEL/Debian or Certified Kubernetes?

I am not saying the downloader/installer idea isn't great. Most users do have severer pain points customizing TF builds. I am just saying it might be strategically preferred to create a void in the ecosystem, so third-party packaging teams could find a way to contribute by filling up this gap. We already have SIG-Build; what is their role in this initiative?

IMHO, an official TF downloader is good, but it adds little value to what we already have (PyPI entries that supports all major OSs and at least one mainstream CUDA version). I am afraid once we have this RFC perfectly implemented, the complexity will be hidden away from not only users but developers as well. This might, in the long term, be of little help for the community developers to acquire a better understanding of the TF build infrastructure.

rfcs/20190225-tf-on-demand.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Front-end
+
+**Web UI (download.tensorflow.org)**
+
+We propose a simple web interface that allows users to manually enter necessary system specs and desired build options for requesting a custom TF binary. The UI will be straightforward and easy-to-use. A sample mock up UI is shown below. (Please refer to [PyTorch](https://pytorch.org) download UI as an additional reference.)
+
+![drawing](https://docs.google.com/drawings/d/1Krze2no7zjfqe7nldOm-ArECOGXVVjgtk0VMXCabzkw/export/png)
+
+Once all fields are filled in, the system backend will check if the corresponding binary has already been built and present in cache. If it exists, then it will provide the user a URL to the binary for download. If it does not exist, then it will ask the user for an email address to receive a link for downloading the newly built binary.
```
jackwish: To me, the binaries collection are a fixed one - fixed combination of platforms/configurations from the Build system. So, for the configuration not supported by default Build system, people need to build by their own, right?

And, what about the addon (contrib), as they are moving out from TF repo. Are they included in the selectable configurations?

rfcs/20190225-tf-on-demand.md
```diff
@@ -0,0 +1,191 @@
...
+    *   GPU too old (Cuda Compute Capability older than 3.0)
+*   *Soft Constraint Failures*
+    *   CUDA version too old
+    *   cuDNN missing
+    *   Other runtime libraries missing or too old (libstdc++)
+    *   Unsupported python version found
+
+In hard constraint failure cases, the tool will point users to cloud options. In soft constraint failure cases, the tool will offer alternative build configurations (CPU only) or request users to install the missing software.
+
+We propose distributing this binary through PyPI. However, please note that binary distribution method is still under discussion.
```
jackwish: Is it possible to integrate into the existing TF PyPI packages? People may get confused when there are too many PyPI packages to choose.

rfcs/20190225-tf-on-demand.md
sub-mod: It can reference any branch

rfcs/20190225-tf-on-demand.md
sub-mod: These builds are for specific combination of OS/python/CUDA/arch etc. PyPI currently lacks the capability to differentiate wheels based on platform.
PyPi will continue to have the tensorflow wheels with generic arch.

rfcs/20190225-tf-on-demand.md
sub-mod: As new combinations are requested we can expand the list.

r4nt: Why don't we expect basically ~all configurations that don't have active CI to be broken? That is, what happens with random compile errors that will show up in a lot of different combinations of libc/compiler/cuda?

rfcs/20190225-tf-on-demand.md
fridex: Check this ML thread: https://www.mail-archive.com/wheel-builders@python.org/msg00334.html

rfcs/20190225-tf-on-demand.md
goern: From my point of view its worth to think about publishing TF Wheels in a different thread: there are a few issues like naming (mentioned by Subin above), include build time configuration in wheel file, ...

I am sure we can not change the way PyPi works short term, so we need to come up with a cleaver structure of index hosts and directories.

rfcs/20190225-tf-on-demand.md
goern: @hyeygit ?

hyeydummy1: For now, the built artifacts will be available indefinitely. In the future, however, we may think about cleaning them up. 

Also, thank you for the suggestions. We will definitely consider the options.

hyeygit: For now, we plan to have the built artifacts be available indefinitely. In the future, however, we may think about cleaning them up.

hyeygit: Regarding storing the build options, we are still actively discussing, trying to come up with a solution that would be manageable and scalable. The second option you provided has actually been brought up. I do agree that giving the users the option to retrieve the build configuration of a wheel file would be handy. Will make sure to consider the first option.

rfcs/20190225-tf-on-demand.md
hyeygit: All diagrams are updated with better resolution!

rfcs/20190225-tf-on-demand.md
hyeygit: The initial version will include core tensorflow only (and no addons like contrib). However, as @sub-mod explained above, we will consider including addons and/or sub-projects based on incoming requests.

hyeygit: For unsupported configurations, the tool will suggest best-known alternative solutions. Depending on the case (and most likely), the user will need to build the wheel on his/her own.

rfcs/20190225-tf-on-demand.md
hyeygit: PyPI lacks the ability to distinguish different build configurations. The system will not be integrated with PyPI (and therefore it won't introduce confusion to users trying to download TF PyPI packages).

foxik:
Are there any plans to support libtensorflow as well? We are thinking of distributing a C++ application using the C libtensorflow as a backend, and having a simple way of compiling the library for the user system would be a huge help.

rfcs/20190225-tf-on-demand.md
hyeygit: The system will automatically file bugs for such compile errors and suggest alternative solutions to the users to help complete the build. 

We currently run a large set of builds on tf-nightly across multiple platforms to make sure the builds are successful and to catch any such random compile errors upon failures. Based on our observation, we do not anticipate running into these issues, at least not too frequently. 

Upon receiving these compiler error bugs, we will triage them and aim to improve the system appropriately.

Looping in @gunan for any additional explanation.

rfcs/20190225-tf-on-demand.md
```diff
@@ -0,0 +1,191 @@
...
+   // Maybe, maybe not
+   date last_built;
+   date last_downloaded;
+   int download_count;
+}
+```
+
+### Other Details
+
+**Bug Filing**
```
claynerobison: What is the plan to allow feature requests/PR submissions? For example, how would I request for the system to build a binary with compiler flags for the latest Xeon instruction set?

claynerobison: And how frequently will the build system compiler be updated to support updated `-m` flags?

sub-mod: As long the compiler flags are available for the OC+gcc combination it can be passed in to the build environment.

rfcs/20190225-tf-on-demand.md
agramesh1: Would there be a way to request a specific version of gcc to be used for compilation?

rfcs/20190225-tf-on-demand.md
claynerobison: `OC`?

sub-mod: I meant OS(operating System)

rfcs/20190225-tf-on-demand.md
sub-mod: yes we can do that. Alos for centos/rhel  OS we can install a certain devtoolset based on user input.

hyeygit:
@foxik Regarding your question above --we will start with pip packages first and then will expand to supporting others like libtensorflow later.

Looping in @gunan for any additional comments.

ewilderj:
@hyeygit  Please push a change with "Status: Accepted" and I will merge this RFC.

hyeygit:
**Summary**
- Built TF artifacts will be available indefinitely.
- TF binaries will still be release on PyPI. This project is a separate effort.
- Initial release for this project will include core TF only (excluding addons / sub-projects / libtensorflow etc.).
- Additional supports will be included in the follow up releases depending on incoming requests. 
- Supporting new instruction set(s) will be possible as long as the compiler flags are available (for the OS and GCC combination) and can be passed into the build environment.
- Wheel file naming convention will be finalized during the implementation phase as it is subject to change. The solution will aim to respect the PEP standards around wheel files and provide users the ability to retrieve build configurations of a wheel file.

PR #77, opened 2019-03-12
RFC: Modular TensorFlow

gunan:
**Comment period is open until 2019-04-04**

# Modular TensorFlow

| **Author(s)** | Gunhan Gulsoy (gunan@google.com)                     |
| **Updated**   | 2019-03-06                                           |

Objective: Propose an architecture where smaller, more focused TensorFlow modules can be created, maintained and released separately.


rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+### Definitions
+
+This section will briefly describe the terms we will use in the rest of this design.
+
+**Modules:** These are the components of core TF library that will "accept" plugins to expand their capabilities. Examples of modules are networking, filesystem, (graph) optimizers.
+
+**Plugins:** Plugins are extensions to different modules. For example, filesystem module can have a GCP plugin, an S3 plugin, an HDFS plugin.
+
+**Shared objects:** These are  dll/so/dylib files that can house **one or more** plugins for **one or more **modules.
```
jackwish: > `plugins for **one or more **modules.`

failed markdown highlighting syntax :)

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+**Packages: **Python pip packages which may include Python files and/or shared objects.
```
jackwish: same here.

rfcs/20190305-modular-tensorflow.md
gunan: Fixed. Thanks for reporting!

rfcs/20190305-modular-tensorflow.md
gunan: Fixed. Thanks for reporting!

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+#### Long build times
+
+Many volunteers and developers outside Google use their laptops for development. On such systems, TF development cycles require building all of tensorflow that takes around 2 hours. While there is an argument to be made that bazel caching should ensure that they build all of TF only once, without forge, bazel caching is not working as [well as expected](https://docs.bazel.build/versions/master/remote-caching.html#known-issues).
+
+#### Adding support for new hardware is very difficult and not scalable
+
+The ML ecosystem is constantly expanding. New hardware for accelerating ML applications is being worked on by many teams inside and outside Google. As the most popular machine learning framework, TF is expected to add support for many of these hardware as quickly as possible.
+
+Currently, this means that all such hardware developers need to check in their code into the [main tensorflow repository](http://github.com/tensorflow/tensorflow). This means that all the changes are required to go through TF team's review. This can make merging support for new hardware very very difficult.
```
r4nt: ... or maintain a fork of TF, right?

byronyi: The kernel way without GPL enforcing ends you up with Caffe, as the saying goes by, “there are a thousand TensorFlow in a thousand contributors’ eyes”.

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+When this plugin is loaded by TF at runtime, `TF_InitPlugin` method will be called. This method will register the plugin as prescribed by the module, and exit.
+
+Plugin shared objects will need to follow some standards:
+
+
+
+1.  These cannot export any symbols into global namespace to avoid symbol collisions.
+1.  They need to communicate to TF through any provided C APIs.
+1.  They can link anything they like, anyway they like in addition to TF.
```
claynerobison: How would TF core APIs be called from within a plugin? e.g. If an MKL plugin signals that it can handle a usage, but as part of that usage, the MKL plugin needs to call the eigen implementation in TF core, how would that handoff be managed?

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+![alt_text](20190305-modular-tensorflow/big_picture.png "Overview of modular TensorFlow")
+
+A summary of the above is:
+
+
+
+*   Core TF functionality will be implemented in C++
+*   Core TF functionality can be extended using shared objects.
+*   On top of the core C++ libraries, we will have the language bindings (Using the C API)
```
karllessard: As discussed in [this developer thread](https://groups.google.com/a/tensorflow.org/forum/?utm_medium=email&utm_source=footer#!msg/developers/TuN9YDVVOxE/bUP_KwZQCgAJ), it is still not very clear to me how protobufs will be exposed by the C API.

For example, the language bindings will need to retrieve the list of available operations, which is currently represented by a list of `OpDef` described in [opdef.proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/op_def.proto). How will that information cross the C API? 
1. by returning the protobufs as a series of bytes that should be deserialized by the bindings, like in [TF_GetAllRegisteredKernels](https://github.com/tensorflow/tensorflow/blob/1f18aeee0fe45207eb27b4a3d0a6ebbaae165c78/tensorflow/c/c_api.h#L1724)?
2. by exposing a snapshot of the protobufs as a text file that should be parsed by bindings, like in [ops.pbtxt](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/ops/ops.pbtxt)?
3. by wrapping protobuf messages in a C structure exposed in the API?

In other words, can a `.proto` file of TensorFlow be part of the contract with its modules or should they be used only internally by the core?

gunan: Our solution to this will be 3.
We would like to remove all protobufs from our APIs.
We will push for using protobufs for only serialization formats, for RPCs or data we would store on disk.

rfcs/20190305-modular-tensorflow.md
gunan: Yes, here are two examples I am aware of:
https://github.com/ROCmSoftwarePlatform/tensorflow-upstream
https://github.com/codeplaysoftware/tensorflow

rfcs/20190305-modular-tensorflow.md
gunan: This is an interesting use case. I can think of two potential solutions:

1 - MKL plugin can use the global kernel registry to fetch the eigen kernel and execute it.
2 - MKL plugin can rebuild all eigen kernels and also distribute all.

We can brainstorm more on this issue, I think it is certainly worth considering.

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+*   All different pieces need to use Stable public APIs with backwards compatibility guarantees.
+
+
+### Definitions
+
+This section will briefly describe the terms we will use in the rest of this design.
+
+**Modules:** These are the components of core TF library that will "accept" plugins to expand their capabilities. Examples of modules are networking, filesystem, (graph) optimizers.
+
+**Plugins:** Plugins are extensions to different modules. For example, filesystem module can have a GCP plugin, an S3 plugin, an HDFS plugin.
```
claynerobison: This is defining an SDK. When can we get drop? That will generate lots of feedback/questions...

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+
+When this plugin is loaded by TF at runtime, `TF_InitPlugin` method will be called. This method will register the plugin as prescribed by the module, and exit.
+
+Plugin shared objects will need to follow some standards:
+
+
+
+1.  These cannot export any symbols into global namespace to avoid symbol collisions.
+1.  They need to communicate to TF through any provided C APIs.
```
claynerobison: Thread-pool management? How do we coordinate between TF Core and plugins so as to not over/undersubscribe?

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+**Please note that the above is a toy example to explain the abstract ideas described above. Exact implementation can vary across different modules.**
+
+Interface has to be pure C at the ABI level. We can have C++ header-only libraries built on top of these C ABI/APIs.
+
+
+#### Plugins
+
+Plugins need to include implementation of the interfaces declared by one module. If the module interface requires Init and Compute methods, it will need to implement these two functions, plus a TF_InitPlugin function which will be called at load time. This function will also need to register the plugin as prescribed by the module.
+
```
claynerobison: SDK/API architecture has the potential to influence performance of plugins for better and worse. If an API design favors one plugin and hurts another, what is the plan to resolve the conflict?

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+![alt_text](20190305-modular-tensorflow/releases.png "Release plans")
+
+
+To summarize the above timeline:
+
+*   Different packages set their own release cadences
+*   Each package will set version boundaries for each of their dependencies.
+*   Each package is responsible for ensuring that all of their public APIs are working without any changes until the next major release
+*   Packages do not need to modify the minimum version requirements unless they start using newly introduced public API symbols.
+*   TF metapackage releases may choose to hold back individual packages in favor of faster releases. But dependency requirements have to be respected when doing so.
```
claynerobison: What Validation will TF Core team do on plugins to ensure interoperability?

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+![alt_text](20190305-modular-tensorflow/initial_tf_deps.png "Initial package structure")
+
+In the current setup, we need to test all of the above packages for different Python versions, operating systems, accelerators (CPU, GPU), compilers, and more variants combined. In the modularized world, each of these packages only need to be unit tested for the following:
+
+
+*   tensorflow-base: Operating systems, compiler versions and python versions only with CPU 
+*   tf-gpu: With GPU only, for different operating systems.
+*   tf-estimator: Only for different python versions
+
+When testing a package that has dependencies, such as tf-estimator, or tf-gpu, tensorflow-base will be installed with its latest stable release, to ensure to avoid any flakes by this package.
```
claynerobison: How will CI be done? When someone submits a PR to TF Core, can plugins trigger CI to ensure that the PR doesn't cause regressions in the plugin?

claynerobison: And vice versa: when someone submits PR to Plugin, how does TF Core prevent regression?

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+
+![alt_text](20190305-modular-tensorflow/initial_tf_deps.png "Initial package structure")
+
+In the current setup, we need to test all of the above packages for different Python versions, operating systems, accelerators (CPU, GPU), compilers, and more variants combined. In the modularized world, each of these packages only need to be unit tested for the following:
+
+
+*   tensorflow-base: Operating systems, compiler versions and python versions only with CPU 
+*   tf-gpu: With GPU only, for different operating systems.
+*   tf-estimator: Only for different python versions
```
claynerobison: Who will own publication of Tensorflow + MKL binaries?
- anaconda packages
- pypi packages
- container images

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+  ..........
+}
+
+void TF_InitPlugin() {
+  M_RegisterPlugin("A", A_Init, A_Execute);
+}
+```
+
+
+When this plugin is loaded by TF at runtime, `TF_InitPlugin` method will be called. This method will register the plugin as prescribed by the module, and exit.
```
claynerobison: How will conflicts and priority be decided between plugins (e.g. CPU and GPU)?

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+Potential issues:
+
+*   B depends on unreleased APIs A exposed in a newer version
+    *   B needs to release a patch version, with just a change in the minimum required version of A
+
+
+#### Scenario 3: C and B depend on different minimum versions of A
+
+As both C and B have to define a range of versions they require from A, the max version should satisfy both constraints.
```
claynerobison: How do plugins catch these dependency chain conflicts with sufficient time to resolve them before a release cycle?

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+
+#### Scenario 4: User installs C first, but then uninstalls A
+
+This is a pip issue. To help diagnose the problem, B and C should print that A is missing, and user needs to install that to use B or C.
+
+
```
claynerobison: Who owns documentation for plugins? Is there plan for a shared site with plugin contribution?

rfcs/20190305-modular-tensorflow.md
byronyi: Each plugin shall manage their depending TF version by itself.

rfcs/20190305-modular-tensorflow.md
byronyi: Each plugin resides in its own repository, so does the doc site (separated from TF core docs).

rfcs/20190305-modular-tensorflow.md
gunan: What kind of conflicts did you have in mind?
As described here, when we set the max required version to the next major release, plugins can release independently, and as each plugin is independently released pip should be able to take care of satisfying the constraints for 'A'. with the max required A version the same for all plugins, the requirement intervals of versions of 'A' are guaranteed to have an intersection.

rfcs/20190305-modular-tensorflow.md
gunan: Just describing from the current state (as official builds do not have MKL enabled yet), TF team releases main TF package, and Intel releases tf-mkl plugin. Again,  this is reflecting only the current state, and not taking into account any potential changes to mkl support in TF. We can evaluate those changes orthogonally. To clarify the 3 points you raised in the above described scenario:

- Anaconda packages: TF team does not own any anaconda packages. We are separately evaluating anaconda, so it is an orthogonal issue.
- pypi packages: `tensorflow` pypi package is owned by TF team, and `tensorflow-mkl` is a plugin owned by Intel.
- container images: same as pypi packages, but Intel has the option to extend the TF container images just with `pip install tensorflow-mkl` command to build the MKL images on top of TF images.

rfcs/20190305-modular-tensorflow.md
gunan: @sjamesr is working on this. We are trying to build the API based on `OpKernel` and `OpKernelContext` classes.

rfcs/20190305-modular-tensorflow.md
gunan: CC @sjamesr and @m3bm3b to comment on this.

rfcs/20190305-modular-tensorflow.md
gunan: This is a great question. From our side, we will try to be as neutral as possible. Initial APIs will likely be based on the current C++ APIs TF has. But the API designs will be individually reviewed through different RFCs. We hope that we will be able to evaluate all concerns during these reviews. If we make mistakes favoring one or the other framework, we hope to alleviate these through addition of new symbols, and complete redesigns with the next major version releases.

rfcs/20190305-modular-tensorflow.md
gunan: For a select number of plugins, we will run continuous integration tests. For Filesystems, these will be as simple as reading and writing files. For ops and kernels, we will run convergence and performance testing on select models.

rfcs/20190305-modular-tensorflow.md
gunan: CC @lamberta
It is actually a mix. There will have to be some plugins, like estimator we will declare as a part of core TF APIs, and document as such. For some other plugins, what @byronyi said will apply.

gunan: If such regressions can be detected with unit tests, those would be the detectors at presubmit time.  If not, we will run these after merging changes. Most of our regression tests run after merging PRs right now.

rfcs/20190305-modular-tensorflow.md
gunan: We will try to avoid conflicts.
For ops, we have this rough idea: Each op plugin is required to use a separate "namespace" assigned to them for creating new ops. But we will have to improve on this idea.

rfcs/20190305-modular-tensorflow.md
gunan: CC @lamberta
It is actually a mix. There will have to be some plugins, like estimator we will declare as a part of core TF APIs, and document as such. For some other plugins, what @byronyi said will apply.

rfcs/20190305-modular-tensorflow.md
gunan: No, our goal is for each repository to run their own unit tests only for PRs. Currently, TF's situation is becoming unmanageable, and it is a big developer concern on TF side.
Plugins will rely on TF core to keep its API syntactically and semantically compatible. Then they will run their own tests.
If the integration tests after PRs are merged uncover backwards incompatible changes in TF, or the other dependencies in the ecosystem these will be rolled back.

rfcs/20190305-modular-tensorflow.md
claynerobison: A global kernel registry sounds interesting. @agramesh1, @nhasabni do you see performance implications here?

rfcs/20190305-modular-tensorflow.md
claynerobison: Do you know what `select number of plugins` includes?

rfcs/20190305-modular-tensorflow.md
claynerobison: CC @agramesh1

rfcs/20190305-modular-tensorflow.md
claynerobison: Clarification: if two plugins register the ability to handle the same op (or whatever granularity of operation), how will priority be decided between them?

rfcs/20190305-modular-tensorflow.md
m3bm3b: Our expectation for the moment is that threadpool management will be much as it is today.   
That is, plugins will typically use a threadpool provided to them 
(such as via OpKernelContext); the key difference would be that the interface 
would be provided as part of a stable ABI, so that it can be used across the boundaries 
of shared libraries.   Thus, we don't expect that interactions will be substantially 
better or worse than today.

We cannot prevent  the author of a plugin from creating a threadpool local to that plugin,
but, as you are perhaps suggesting, to do so risks making interactions with other 
threadpools within the address space be less predictable.  That might well make such a 
plugin less desirable by clients.

rfcs/20190305-modular-tensorflow.md
mahmoud-abuzaina: For MKL, right now we depend on many cc_library(s) from core in bazel build files. Considering a separate MKL plugin, would we be able to do the same? Or will the core build files somehow expose the libraries to depend on?

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+### Problems addressed
+
+#### Spaghetti dependencies
+
+Everything being in a single repository encourages everyone to freely use all other code in the repository, even when this does not make sense. This ensures the library is very difficult to modify later.
+
+#### Long build times
+
+Many volunteers and developers outside Google use their laptops for development. On such systems, TF development cycles require building all of tensorflow that takes around 2 hours. While there is an argument to be made that bazel caching should ensure that they build all of TF only once, without forge, bazel caching is not working as [well as expected](https://docs.bazel.build/versions/master/remote-caching.html#known-issues).
+
```
mahmoud-abuzaina: With the new design, is using bazel still a must for building the separate plugins (like MKL) or other build tools (like cmake) can be used?

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+When this plugin is loaded by TF at runtime, `TF_InitPlugin` method will be called. This method will register the plugin as prescribed by the module, and exit.
+
+Plugin shared objects will need to follow some standards:
+
+
+
+1.  These cannot export any symbols into global namespace to avoid symbol collisions.
+1.  They need to communicate to TF through any provided C APIs.
+1.  They can link anything they like, anyway they like in addition to TF.
+1.  They can be built and distributed separately
```
mahmoud-abuzaina: Can different compilers other that GCC be used to build plugins (ex: Intel ICC compiler)?

gunan: Yes. But when using different compilers, it becomes very important to only use C symbols at the ABI level.

rfcs/20190305-modular-tensorflow.md
gunan: Plugins that are not in the main tensorflow repository are free to use any build system they would like.

rfcs/20190305-modular-tensorflow.md
gunan: We are right now working to create a C API for all libraries accessible through OpKernel and OpKernelContext classes. We are hoping that will cover all the required libraries for kernels. We can discuss if MKL kernels need more symbols.

rfcs/20190305-modular-tensorflow.md
gunan: In the current implementation of the kernel registry, you can add extra labels to each kernel being registered, and use those to select which kernels you would like to pick:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/kernel_def.proto#L35
Our initial idea is to leverage this, but @sjamesr may have other ideas.

rfcs/20190305-modular-tensorflow.md
gunan: At the moment, "select number of plugins" will be the "features currently officially supported by TF", such as CUDA, cloud file systems, GRPC based networking.
This list may grow or shrink over time.

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+*   Environment (google3, opensource)
+*   Operating system (Linux, windows, macos, mobile)
+*   Architecture (x86, arm, ppc64, …)
+*   Accelerator (CPU, GPU,  TPU)
+*   Compiler (GCC, Clang, MSVC)
+*   Python version (2, 3.4, 3.5, 3.6,  3.7)
+
+More can be added to this list where we have to rebuild TensorFlow to support different network architectures, CUDA versions, SIMD instruction sets, etc.
+
+Having a monolithic repository means we need to rebuild all of our code for all of these different combinations. However, it makes no sense to rebuild all of our C++ code if the only difference is the Python version. Or rebuild all of our CPU kernels for different CUDA versions. Modularizing our code means we only need to rebuild and test the modules that are directly impacted by the dimensions we are changing in the support matrix.
```
nammbash: Can you touch up a little more on this PR process envisioned with this change.
Currently Intel push changes to google and google owns the process of verifying for 
1. coding standards
2. API compatibility
3. backwards compatibility
4. Unit and regression tests.
5. others

Now for future PR's for Intel MKL repo, will Intel police its own PRs into this separated module Repo?

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+#### Large TF support matrix
+
+The TF support story is a unique beast. Our support matrix has a lot of orthogonal dimensions. Below are some of the more prominent dimensions:
+
+*   Environment (google3, opensource)
+*   Operating system (Linux, windows, macos, mobile)
+*   Architecture (x86, arm, ppc64, …)
+*   Accelerator (CPU, GPU,  TPU)
+*   Compiler (GCC, Clang, MSVC)
+*   Python version (2, 3.4, 3.5, 3.6,  3.7)
```
nammbash: After the modularized approach,(Lets say for the Intel MKL Module:)
1. who will be running these tests for these support matrix or we care only about the Linux environment alone?
2. Who Kicks off the CI process to test MKL. Intel or Google.

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+More can be added to this list where we have to rebuild TensorFlow to support different network architectures, CUDA versions, SIMD instruction sets, etc.
```
nammbash: How do we handle MKL PR by outsiders? What should Intel do?

rfcs/20190305-modular-tensorflow.md
gunan: We already have a global kernel registry in TF:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/op_kernel.cc#L1048

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+*   Different packages set their own release cadences
+*   Each package will set version boundaries for each of their dependencies.
+*   Each package is responsible for ensuring that all of their public APIs are working without any changes until the next major release
+*   Packages do not need to modify the minimum version requirements unless they start using newly introduced public API symbols.
+*   TF metapackage releases may choose to hold back individual packages in favor of faster releases. But dependency requirements have to be respected when doing so.
+*   Major releases still need to be coordinated.
+
+
+## Packaging Issue Scenarios
```
nammbash: How do we maintain release dependence? 
Intel syncs dot releases with the major release of Google?
If Google release TF 3.0 with MKL 2.17 dependency, the next MKL is 2.18 (working with TF 2.0 and TF 3.0) or TF-mkl release 2.18 or 3.1 or both?

shengfuintel:
Some of GPU kernels and CPU kernels actually shared the templated functions. If they are separated into different modules, there are lots of duplicated code.

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+Currently, this means that all such hardware developers need to check in their code into the [main tensorflow repository](http://github.com/tensorflow/tensorflow). This means that all the changes are required to go through TF team's review. This can make merging support for new hardware very very difficult.
+
+#### Long PR review queue
+
+TensorFlow is a very successful opensource project. It is the [4th most forked](https://github.com/search?o=desc&q=stars:%3E1&s=forks&type=Repositories) and [5th most starred](https://github.com/search?q=stars%3A%3E0&s=stars&type=Repositories) project on github. This also means that TF receives quite a lot of opensource contributions. The TensorFlow team has to review all contributions to TensorFlow itself. This creates a bottleneck for merging the changes to the main repository.
+
+#### Flexibility for collaborators
+
+Currently, any partner or contributor that would like to work with us are subject to all the rules within of the main repository. Some of these can be relaxed through modularization, where work can happen in a separate repository.
```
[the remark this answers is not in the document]
gunan: It will be up to the owners of each split repository.
They can be under github tensorflow org, they can be closed source, they can be on github and different org, or hosted on different websites.

rfcs/20190305-modular-tensorflow.md
gunan: If Intel would like to own the MKL repository, yes, intel will have full control over all of these. You will have to provide backwards compatibility on your public APIs.
Each repository owner will have full control over their repository.
For regression testing, we will work together to make sure the continuous integration of Base TF packages and all plugins will continue to work together.

rfcs/20190305-modular-tensorflow.md
gunan: For MKL module, assuming Intel taking full ownership of the MKL module (which we can discuss), will be intel's decision for both. Intel only wants to support Linux, sure. Intel wants to support all operating systems, sure.
TF and MKL presubmits will be independent, and their CIs will be owned by the repository owners. The integration tests will be a collaboration.

rfcs/20190305-modular-tensorflow.md
gunan: That would be entirely intel's call.
Intel may choose to completely close the source of the MKL plugin, or manage the plugin code just like another project, just like this one: https://github.com/intel/mkl-dnn

rfcs/20190305-modular-tensorflow.md
gunan: I am not sure what you mean. Main TF package can either depend on MKL, or be depended on MKL. We will not accept a circular dependency.
If TF depends on MKL, MKL will depend on tensorflow-base. If MKL chooses to depend on TF, that will work like an add on.

What we would like to do is, just like described here, enable everyone to do independent releases. Therefore, avoiding circular dependencies is a must.

gunan:
@shengfuintel we are looking into how to split the current kernels, as a part of https://github.com/tensorflow/community/blob/master/rfcs/20180604-dynamic-kernels.md.
We will likely create two shared objects, where GPU kernels depend on CPU kernels, and they are built together. In the future, we may look into separately optimizing the two.

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+
+### C++
+
+This project aims to implement similar plugin architectures for multiple components of TF code. While these will be listed separately, **there will be a lot of shared pieces between these components**. The modules we would like to handle are:
+
+1.  Networking module, with verbs, gdr plugins initially
+1.  Filesystems module, with GCP, AWS and HDFS support
+1.  Kernels module, 
+1.  Optimizers/Graph rewrite module,
```
nhasabni: (1) Will optimizers/graph rewrite modules offer same capabilities (such as different phases of rewrite such as pre-placement, post-placement, etc) that exists in current design? (2) What if plugins/backends want to have plugin-specific or backend-specific optimizers?

nhasabni: Also, relevant question is if a plugin/backend needs a specific core component. To be precise, MKL-CPU backend currently uses BFCAllocator from TF core. How can we continue to ensure that this dependency is met?

gunan: Our aim is to provide all the current capabilities, and we are working with grappler team and a few more groups internally to see how we should rework it. But our aim is to functionally match the current way the graph rewriting is possible with checking code into tensorflow.

Currently, for kernels our plan is to provide semantic equivalents to everything in OpKernelContext. While I cannot see a public interface to BFCAllocator there, we will be able to add more symbols to the API, as long as previous symbols are not changed.

rfcs/20190305-modular-tensorflow.md
mdfaijul: @gunan It is not clear how to add new operators to a module plugin. Currently, we could create new operators through `OpDef` and optionally set visibility hidden. We have been doing this for MKL-DNN quantization work.

gunan: The idea is to use a mechanism similar to custom ops, but simplified. With a C ABI, we will be able to make custom ops much easier to build and distribute.

rfcs/20190305-modular-tensorflow.md
agramesh1: Follow on to @claynerobison 's comment. Currently we use different default threadpool settings and also set some MKL specific settings when MKL is enabled (see https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/common_runtime/threadpool_device.cc)  Would there be a way to do this from the plugins?

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+*   **TF will become a suite of multiple packages, built by multiple authorities. What if the bugs get blamed on TF team**: With the modular model, we expect testing of 3rd party code to become easier. This can also be mitigated if the error messages are better, and if they can clearly point out which module the issue stems from. Finally, we can create an apple-swift like testing model, where we run a Jenkins setup that people can donate their machines to, and we can run continuous integration tests on their plugins.
+*   **Why not have APIs but still have a monolithic repository: **When everything is in a single repository, this enables developers to bypass the APIs, and depend on internals. Moreover, we cannot grant full control over different folders on our repository to our partners in a single repository. As long as they are in a single repository, they are still constrained by our build system and license. Finally, in a single repository we do not provide the option of closed source plugins for contributors.
```
karllessard: What will be the constraint in regard of the licences? I guess if a module is outside the TF organization, license could be any that complies with the one with TensorFlow but what about modules that remains under github.com/tensorflow but outside the main repo?

byronyi: https://github.com/tensorflow/community/pull/18#issuecomment-424389704

aaroey:
@pooyadavoodi Hi Pooya, would you please find people on your side to take a look at this RFC? Thanks.

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+typedef struct MInterface {
+  void (*M_Init)(M_context*, string);
+  void (*M_Execute)(M_context*);
+};
+
+// The plugin registry for module M
+// The registry here implemented uses C++ class std::map
+// This is OK as it is not exposed in our example here.
+// As far as module implementations are concerned, they only
+// need to see and use M_RegisterPlugin method, and that method takes // care of their addition into this registry.
```
aaroey: Split the line into two.

rfcs/20190305-modular-tensorflow.md
```diff
@@ -0,0 +1,481 @@
...
+
+void M_Execute(M_context* ctx) {
+  ctx->interface.M_Execute(ctx);
+}
+```
+
+
+**Please note that the above is a toy example to explain the abstract ideas described above. Exact implementation can vary across different modules.**
+
+Interface has to be pure C at the ABI level. We can have C++ header-only libraries built on top of these C ABI/APIs.
```
alsrgv: Great idea to offer header-only C++ libraries that will make C ABI more accessible.

ewilderj:
@gunan as the design review meeting has happened, can you post any notes from that into the comments here, please?

After that, we're good to push a change amending status to "Accepted", and I can merge this.

rfcs/20190305-modular-tensorflow.md
bhack: So is the official c/c++ API just a core API?
I.e. If I model/train in python with the new "use Keras in TF 2.0" mantra/marketing I am not sure that I can use the same API in c++ to integrate inference in a c++ app.
I still not find a way on how we can achieve this in the roadmap.
Train with python API, quickly export the model and load it in c++, make some ops call in c++ at inference. 
I.e. Just to check the competitor I see Aten has a very good roadmap to stay on par with python API https://pytorch.org/cppdocs/ and Torchscript makes trivial to import a just trained model in c++ on the fly with just with a one or two Trocscript call.

rfcs/20190305-modular-tensorflow.md
karllessard: Quick question, how will the TensorFlow libraries be published to be used by language bindings?

Right now, pretty much all bindings have their own distribution (PiP package, Maven artifacts...) that bundles both the client and the TF core libraries. But with module TF, will the core libraries built, tested and published by Google so the clients just need to pick them for distribution?

bhack:
What is the status of this?

mihaimaruseac:
It is work in progress, the first module will be the filesystems (#101)

bhack:
Is this megiable or not?

gunan:
It will be merged, but I have to edit some of the sections.
I am sorry for taking so long to merge this proposal.
While this is not merged, as you can see from other RFCs the work is definitely under way.

theadactyl:
@gunan any way we could resolve this RFC?

bhack:
Can we try to extend the deadline if needed? It Is still set on 2019-04-04

rfcs/20190305-modular-tensorflow.md
gunan: A further clarification.
The SDK will start as what it is today.
We will try to make all the kernel code work as is. Therefore, you can think of the current TF C++ headers as v0 of what is to come for TF C++ APIs.

PR #81, opened 2019-03-14
Update 20180920-unify-rnn-interface.md

zh794390558:


qlzh727:
LGTM, thanks for the fix.

bbrito:
Is this update already implemented in TF 2.0? Any example of how to migrate from static_rnn to keras.rnn?

Thanks!

limjq45:
I am trying to change my RNN to a static RNN, but I couldn't get it to work.
```
Old : model.add(Bidirectional(LSTM(128, return_sequences=True)))
```
```
new : model.add(Bidirectional(tf.keras.layers.RNN(tf.compat.v1.nn.rnn_cell.LSTMCell(128), return_sequences=True, unroll=True)))
```

The error I got was : 
```
TypeError: __init__() missing 1 required positional argument: 'units'
```

PR #83, opened 2019-03-16
RFC: Control Flow in TensorFlow Lite

miaout17:
Comment period is open until 2019-04-01

Status        | Proposed
:------------ | :---------------------------------------------------------------
**Author(s)** | Yu-Cheng Ling (ycling@google.com)
**Sponsor**   | Andrew Selle (aselle@google.com), Jared Duke (jdduke@google.com)
**Updated**   | 2019-03-15

**Summary**

We aim to make TensorFlow Lite easy to use. One of the ultimate goals is to be able to convert any TensorFlow model to TensorFlow Lite and run it correctly. The Select TensorFlow operators project increased hundreds of supported ops by running TensorFlow kernels in TensorFlow Lite. However, some of the TensorFlow ops are not supportible by this approach. Control flow is one of the biggest missing features.

This RFC discusses how control flow is defined, converted from TensorFlow, and implemented in TensorFlow Lite.


ewilderj:
Please add any notes from the design review meeting to this PR, and send a new push changing the status from Proposed to Accepted.

Thanks!

cena001plus:
hi,all
i face a issue ,   i convert ckpt to pb , pb model can give  right result,  then i convert pb model to tflite , it's success , but i test  and load tflite model,  it give a wrong result. i hava tried many method, but question is still stay, can you give me some suggestion?  @ewilderj @miaout17 @googlebot

miaout17:
@ewilderj sorry for the long delay since I missed your previous comment. 

I updated the status and here's the review notes:
 
* Discussion around control flow V1 support
  * We consider to support control flow V1 for legacy models
  * In addition to control flow V1, TensorList is also a problem. Legacy models use TensorArray but the new conversion flow will only support TensorList. 
  * Raising control flow from V1 to V2 isn't always trivial and it's not guaranteed to work. This will be best-effort behavior. The converter will attempt and error out if unable to.
  * Decision: Focus on TF 2.0 supported ops first (Control Flow V2, TensorList). 
* What control flow ops are we expected to see in v2?
  * If and While, For can be done if needed but is not generable by the user
* Function recursion  * Not a high priority because TensorFlow doesn't support this either. 
  * Can not be supported but should not be excluded from design. No real use case for now

miaout17:
@cena001plus sorry that I didn't see your post earlier. 
This is a RFC for discussing a new design. 

For your question, I'd suggest to post it on StackOverflow or [Github Issues on TensorFlow repo](https://github.com/tensorflow/tensorflow/issues). Thanks!

PR #84, opened 2019-03-25
RFC: Sustainability of TensorFlow Addons

dynamicwebpaige:
**Comment period ends 2019-04-08**

# Sustainability of TensorFlow Addons

| **Author(s)** | Paige Bailey (webpaige@google.com), Sean Morgan (seanmorgan@outlook.com), Yan Facai (facai.yan@gmail.com) |
| **Sponsor**   | Karmel Allison (karmel@google.com)                 |
| **Updated**   | 2019-03-23                                         |

## Objective
 
[TensorFlow Addons](https://github.com/tensorflow/addons) is a repository of contributions that conform to well-established API patterns, but implement new functionality not available in core TensorFlow. For TensorFlow Addons to support [code that is moved from `tf.contrib` to `addons`](https://github.com/tensorflow/community/blob/ef626896f30130dfc3b5e75126c94624b689a943/rfcs/20181214-move-to-addons.md#code-to-be-moved-from-tfcontrib-to-addons), as well as new functionality, requirements for maintaining and retiring those submodules must be defined and enforced.

This document details requirements and responsibilities for owning a subpackage/submodule that is included in TensorFlow Addons, as well as the periodic review process for all components of Addons.

In this RFC, we are soliciting discussion regarding Addons maintainership and periodic review. This RFC discussion will help the SIG Addons team determine appropriate roles and responsibilities for proxy maintainers.

### DIFF
--- a/rfcs/20190308-addons-proxy-maintainership.md
+++ b/rfcs/20190308-addons-proxy-maintainership.md
@@ -129,9 +129,23 @@ review will be:
 1. Number of active maintainers
 2. Amount of OSS use
 3. Amount of issues or bugs attributed to the code
+4. A better solution becomes available
 
-Because we encourage TF-Addons to be used in production, we'll plan on a 
-graceful deprecation process, instead of a swift removal.
+Functionality within TensorFlow Addons can be categorized into three groups:
+
+* **Suggested**: well-maintained API; use is encouraged.
+* **Discouraged**: a better alternative is available; the API is kept for historic reasons; or the API requires maintenance and is the waiting period to be deprecated.
+* **Deprecated**: use at your own risk; subject to be deleted.
+
+The status change between these three groups is: Suggested <-> Discouraged -> Deprecated.
+
+The period between an API being marked as deprecated and being deleted will be 90 days.
+
+1. In the event that TensorFlow Addons releases monthly, there will be 2-3 releases before an API is deleted. The release notes could give user enough warning.
+
+2. 90 days gives maintainers ample time to fix their code.
+
+3. Google often gives a 90 day window for security fixes, which is a trade off between risk and time to respond. For more comparison of different time window, please see [this link](https://googleprojectzero.blogspot.com/search?q=90+days).
 
 ### Adding a New Subpackage
 
--- a/rfcs/20190308-addons-proxy-maintainership.md
+++ b/rfcs/20190308-addons-proxy-maintainership.md
@@ -1,6 +1,6 @@
 # Sustainability of TensorFlow Addons
 
-| Status        | Proposed       |
+| Status        | Accepted       |
 :-------------- |:---------------------------------------------------- |
 | **Author(s)** | Paige Bailey (webpaige@google.com), Sean Morgan (seanmorgan@outlook.com), Yan Facai (facai.yan@gmail.com) |
 | **Sponsor**   | Karmel Allison (karmel@google.com)                 |
--- a/rfcs/20190308-addons-proxy-maintainership.md
+++ b/rfcs/20190308-addons-proxy-maintainership.md
@@ -1,0 +1,174 @@
+# Sustainability of TensorFlow Addons
+
+| Status        | Proposed       |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | Paige Bailey (webpaige@google.com), Sean Morgan (seanmorgan@outlook.com), Yan Facai (facai.yan@gmail.com) |
+| **Sponsor**   | Karmel Allison (karmel@google.com)                 |
+| **Updated**   | 2019-03-23                                         |
+
+## Objective
+ 
+[TensorFlow Addons](https://github.com/tensorflow/addons) is a repository of 
+contributions that conform to well-established API patterns, but implement new 
+functionality not available in core TensorFlow. For TensorFlow Addons to 
+support [code that is moved from `tf.contrib` to `addons`](https://github.com/tensorflow/community/blob/ef626896f30130dfc3b5e75126c94624b689a943/rfcs/20181214-move-to-addons.md#code-to-be-moved-from-tfcontrib-to-addons),
+as well as new functionality, requirements for maintaining and retiring those 
+submodules must be defined and enforced.
+
+This document details requirements and responsibilities for owning a 
+subpackage/submodule that is included in TensorFlow Addons, as well as the 
+periodic review process for all components of Addons.
+
+## Motivation
+
+In this RFC, we are soliciting discussion regarding Addons maintainership 
+and periodic review. This RFC discussion will help the SIG Addons team 
+determine appropriate roles and responsibilities for proxy maintainers.
+
+## Design Proposal 
+
+### Repository Structure
+
+TensorFlow Addons is structured in a into a heirarchy of subpackages 
+and submodules as described in the [python docs](https://docs.python.org/3/tutorial/modules.html#packages).
+
+A newly proposed subpackage will be required to have at least one active proxy 
+maintainer or organization; and the inclusion must be 
+[discussed with the community at large](https://github.com/tensorflow/addons/issues/58) 
+to see if its matches the addons paradigm. A newly proposed submodule or code 
+addition is required to get approval from the governing subpackage maintainers. 
+If a contributor is not willing to maintain a piece of code, 
+subpackage owners will decide if they are willing to maintain 
+the code going forward and if it should be accepted into future addons releases.
+
+TensorFlow Addons would like to build as inclusive of a community as 
+possible. More than one maintainer is encouraged for subpackages and submodules. 
+The Addons core team will send out monthly emails to the addons [Google Group](https://groups.google.com/a/tensorflow.org/forum/#!forum/addons) detailing which subpackages/submodules are light on maintainers and encouraging 
+community support.
+
+### Subpackages 
+
+To date, the `tensorflow/addons` [subpackages](https://github.com/tensorflow/addons#maintainers) include:
+
+| Subpackage    | Maintainers  | Contact Info                        |
+|:----------------------- |:----------- |:----------------------------|
+| tfa.activations | SIG-Addons | addons@tensorflow.org    |
+| tfa.image |  |                                   |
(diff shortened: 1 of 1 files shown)


rfcs/20190308-addons-proxy-maintainership.md
```diff
@@ -0,0 +1,174 @@
...
+
+In this RFC, we are soliciting discussion regarding Addons maintainership 
+and periodic review. This RFC discussion will help the SIG Addons team 
+determine appropriate roles and responsibilities for proxy maintainers.
+
+## Design Proposal 
+
+### Repository Structure
+
+TensorFlow Addons is structured in a into a heirarchy of subpackages
```
alextp: s/heirarchy/hierarchy

rfcs/20190308-addons-proxy-maintainership.md
```diff
@@ -0,0 +1,174 @@
...
+
+
+## Maintainership 
+
+### How to Become a Proxied Maintainer
+
+The easiest way to become a proxy maintainer is to respond to our monthly 
+"request for help" emails. Subpackage maintainership will only be 
+granted after substantial contribution has been made in order to limit the 
+number of users with write perission. Contributions can come in the form of
```
alextp: s/perission/permission

rfcs/20190308-addons-proxy-maintainership.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Maintainership 
+
+### How to Become a Proxied Maintainer
+
+The easiest way to become a proxy maintainer is to respond to our monthly 
+"request for help" emails. Subpackage maintainership will only be 
+granted after substantial contribution has been made in order to limit the 
+number of users with write perission. Contributions can come in the form of 
+issue closings, bug fixes, new code, or optimizing existing code. Submodule
```
alextp: or documentation?

bhack:
One of the main issue about addons/plugins systems is how you can quickly be aware on what addon you could rely expecially in produciton (without manually checking ticket lag, commit activity etc..).
Can we have a quick reference in the `Readme.md` about "quality/support"  discrete levels of different submodules?
We can never have all the submodules with the same homogeneous level of maturity/support over the time. Popularity could be a good proxy but it doesn't work so well for MIA.

rfcs/20190308-addons-proxy-maintainership.md
```diff
@@ -0,0 +1,174 @@
...
+submission. A potential maintainer would be required to:
+
+* Demonstrate that the code is useful to the community at large.
+* Modify the code to utilize TF 2.x functionality
+
+**Useful Links:**
+* [SIG Addons RFC](https://github.com/tensorflow/community/blob/ef626896f30130dfc3b5e75126c94624b689a943/rfcs/20181214-move-to-addons.md).
+* [SIG Addons Charter](https://github.com/tensorflow/community/blob/master/sigs/addons/CHARTER.md).
+
+## Questions and Discussion Topics
```
lamberta: How are the docs surfaced? (if at all?)
Seems like guides/tutorials/narrative should not surface on tensorflow.org. What about API docs?

facaiy: initial thoughts, if I'm not wrong, we prefer to host it on tensorflow.org, please see the discussion in https://github.com/tensorflow/addons/issues/46

rfcs/20190308-addons-proxy-maintainership.md
facaiy: +1

facaiy:
@bhack 

> Can we have a quick reference in the Readme.md about "quality/support" discrete levels of different submodules?

Sounds great. That's why we propose a periodic review process for all components of Addons. @seanpmorgan @dynamicwebpaige What do you think, Sean, Paige?

rfcs/20190308-addons-proxy-maintainership.md
```diff
@@ -0,0 +1,174 @@
...
+
+Given the nature of this repository, subpackages and submodules may become less 
+and less useful to the community as time goes on. In order to keep the 
+repository sustainable, we'll be performing bi-annual reviews of our code to 
+ensure everything still belongs within the repo. Contributing factors to this 
+review will be:
+
+1. Number of active maintainers
+2. Amount of OSS use
+3. Amount of issues or bugs attributed to the code
```
lc0: would it make sense to aggregate this info into a single place? Could potentially be a good proxy of maturity that @bhack was referring to in https://github.com/tensorflow/community/pull/84#issuecomment-476353018

rfcs/20190308-addons-proxy-maintainership.md
```diff
@@ -0,0 +1,174 @@
...
+
+Because we encourage TF-Addons to be used in production, we'll plan on a 
+graceful deprecation process, instead of a swift removal.
```
lc0: similar here, should we define what is `graceful` in this case. Could be too early, but potentially very valuable promises for people introducing addons in production

rfcs/20190308-addons-proxy-maintainership.md
seanpmorgan: I like this a lot as a way to combine two things we need to keep track of. I worry that it'll be difficult to monitor all sub-packages effectively so this would help out.

rfcs/20190308-addons-proxy-maintainership.md
seanpmorgan: Yeah, this is probably better done sooner rather than later. I do think we'll reserve deprecating things unless they're truly unused in the community at least for a while.

lc0: explicit is better than implicit ;)

rfcs/20190308-addons-proxy-maintainership.md
dynamicwebpaige: Thank you!

rfcs/20190308-addons-proxy-maintainership.md
dynamicwebpaige: +1, as well! Thanks for recommending the add. :slightly_smiling_face:

rfcs/20190308-addons-proxy-maintainership.md
dynamicwebpaige: Great comment! As discussed last week - API documentation for `addons` should conform to the existing documentation standards for TensorFlow, and ideally be centralized with core, `tfp`, `tfds`, `tfdv`, etc.

rfcs/20190308-addons-proxy-maintainership.md
Smokrow: I am wondering if the amount of issues and active maintainers are not kind of the same thing. A module with a lot of issues will have more maintainers as well. 

Another point to consider is that the code quality of a module will increase over time which will lead to a decrease of issues and by this to less active maintainers.

rfcs/20190308-addons-proxy-maintainership.md
```diff
@@ -0,0 +1,174 @@
...
```
Smokrow: We discussed the idea of a 2 phase deprecation process during the monthly meeting. We mark modules as "Going to be removed if nobody responds" or something similar. After half a year we do a "spring cleanup" and code that is not needed anymore will be removed.

rfcs/20190308-addons-proxy-maintainership.md
qlzh727: Agreed that "explicit is better than implicit". 

IMHO, "deprecation" can come from different reasons, eg if the lib is short of maintenance/full of bug, or a better solution is available to solve the same problem, which makes the current API less useful.

I think API can be categorized into 3 groups: Suggested/Discouraged/Deprecated.
Suggested: Well maintained API and encourage user to use.
Discouraged: Better alternative is available, the API is kept for historical reason; or just in the waiting period to be deprecated; or the API is short of maintenance, which might cause issue in prod.
Deprecated: Use at your own risk, subject to be deleted after certain peroid. 

The status change between these three is: Suggested <-> Discouraged -> Deprecated

The period between an API being marked as deprecated and being deleted can be 90 days.
1. If addons does release monthly, there will be 2-3 releases before an API is deleted. The release notes could give user enough warning.
2. It is not too long comparing to 180 days, nor too short as 30 days, so that people can fix their code.
3. Google project zero uses 90 days window for security fix, which is a trade off between risk and time to respond. For more comparison of different time window, please see https://googleprojectzero.blogspot.com/search?q=90+days.

Also we should be mindful the amount of Discouraged API. They should either be fixed, or move to deprecated. Keeping Discouraged API will make long tail and cause the repo hard to maintain.

seanpmorgan: I think this is a solid plan... would you like to submit a modification to this PR? Not sure if you have access to the  branch, but @dynamicwebpaige should be able to help.

rfcs/20190308-addons-proxy-maintainership.md
seanpmorgan: Agree. @qlzh727 described a 3 phase system after this post that I think gives us even more flexibility.

rfcs/20190308-addons-proxy-maintainership.md
qlzh727: I think its fine for either you or Paige to just update the PR.

ewilderj:
In terms of process, shall we consider the next SIG Addons meeting as the final review for this RFC and have it finalized and moved to "Accepted" after that meeting? @seanpmorgan ?

dynamicwebpaige:
> In terms of process, shall we consider the next SIG Addons meeting as the final review for this RFC and have it finalized and moved to "Accepted" after that meeting? @seanpmorgan ?

SGTM! 👍

rfcs/20190308-addons-proxy-maintainership.md
dynamicwebpaige: Updating now! And very much like the plan for Suggested -> Discouraged -> Deprecated. :+1:

ewilderj:
Per today's team call, please push an update changing status to Accepted, and I will merge

ewilderj:
Thanks everyone, congrats on the finalized RFC!

PR #90, opened 2019-04-11
RFC: TensorBoard - Improved Plugin Extensibility

stephanwlee:
Comment period ends 2019-05-19

# TensorBoard: Improved Plugin Extensibility

| **Author(s)** | Stephan Lee (Google), William Chargin (Google) |
| **Sponsor**   | Mani Varadarajan (Google)                 |
| **Updated**   | 2019-05-16                                           |

This RFC proposes a friction-free approach to enabling TensorBoard contributors to create and share TensorBoard plugins as installable extensions.

TensorBoard is composed of various plugins that show up as separate dashboards to users (e.g., scalars, graphs, hparams, histograms, etc.). Today. anyone can contribute a new plugin by forking TensorBoard, adding their plugin, and getting the PR merged. However, this can require a lengthy review to make sure it is architected correctly and is aligned with TensorBoard’s goals.

This RFC covers two changes that will make this process easier:
Plugins can be distributed as Python packages. Users can install them as extensions to TensorBoard.
Plugins are no longer limited to use Polymer but can instead use any frontend library or framework (e.g., React)

The intention is to enable any prospective contributor to easily be able to create or reuse a visualization with tools familiar to them, and enable users to try it out quickly. Plugins that are useful to many users should still be added to TensorBoard, but this removes the friction needed to start getting feedback or trying out something new.

### DIFF
--- a/rfcs/20190411-tensorboard-improved-plugin-ext.md
+++ b/rfcs/20190411-tensorboard-improved-plugin-ext.md
@@ -1,0 +1,132 @@
+# TensorBoard: Improved Plugin Extensibility
+
+| Status        | Accepted       |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | Stephan Lee (Google), William Chargin (Google) |
+| **Sponsor**   | Mani Varadarajan (Google)                 |
+| **Updated**   | 2019-05-16                                           |
+
+## Overview
+
+This RFC proposes a friction-free approach to enabling TensorBoard contributors to create and share TensorBoard plugins as installable extensions.
+
+TensorBoard is composed of various plugins that show up as separate dashboards to users (e.g., scalars, graphs, hparams, histograms, etc.). Today. anyone can contribute a new plugin by forking TensorBoard, adding their plugin, and getting the PR merged. However, this can require a lengthy review to make sure it is architected correctly and is aligned with TensorBoard’s goals.
+
+This RFC covers two changes that will make this process easier:
+1. Plugins can be distributed as Python packages. Users can install them as extensions to TensorBoard.
+2. Plugins are no longer limited to use Polymer but can instead use any frontend library or framework (e.g., React)
+
+The intention is to enable any prospective contributor to easily create or reuse a visualization with tools familiar to them, and enable users to try it out quickly. Plugins that are useful to many users should be able to be packaged as a "first party" plugin, but this removes the friction needed to start getting feedback or trying out something new.
+
+## Motivation
+
+Making the end-to-end plugin development process easier will have the following benefits:
+- Contributors can develop new plugins more quickly/easily as they can use the tools that are familiar to them, while maintaining a consistent user experience within TensorBoard
+- Contributors can test out their plugins without being gated on adding the plugin directly into TensorBoard. This is useful for tools that aren’t necessarily widely used or for quick testing with early users.
+
+## Status quo
+TensorBoard builds JavaScript and TypeScript using a process called Vulcanization -- the process can be described as a language transpilation and a bundler that optionally create a single HTML file with all JS, HTML, and CSS dependencies inlined. All plugin code today lives in the TensorBoard repository and naturally gets Vulcanized, and eventually gets bundled into a TensorBoard pip package and distributed with Python code. When launched, TensorBoard starts a web server that serves the Vulcanized HTML file.
+
+The TensorBoard frontend largely uses Polymer components (while more advanced components use D3, canvas, and WebGL; we do not have technical limitations in supporting a library like React today). TensorBoard repository comes with many shareable components (in `tf_dashboard_common`) that are essential building blocks most plugins and they promote consistency in UI and UX.
+
+## Design Proposal
+### Plugin distribution
+A plugin will be distributed as a single Pip distribution package that includes both a Python backend and a web frontend.
+
+The Python backend must export an object conforming to a standard interface that describes the plugin’s metadata and provides an option to load the plugin’s implementation. The exact shape is not yet specified, but it will be conceptually similar to a function returning a tuple containing a name, a description, and a [`tensorboard.plugins.base_plugin.TBLoader`][tb-loader] object.
+
+In order to be discovered by TensorBoard, the distribution should include an [`entry_points` entry][entry-points] declaring itself as a TensorBoard plugin. The entry point group is `tensorboard_plugins`. The entry point name is arbitrary, and ignored. The entry point object reference should refer to the standard interface described in the previous paragraph. As an example, the setup.py file for a hypothetical “widgets” plugin might look like:
+
+```python
+import setuptools
+setuptools.setup(
+    name="tensorboard_plugin_widgets",
+    description="Widget chart support for TensorBoard",
+    packages=setuptools.find_packages(),
+    # more metadata (version, license, authors…)
+    entry_points={
+        "tensorboard_plugins": [
+            "widgets = tensorboard_plugin_widgets.plugin:get_plugin",
+        ],
+    },
+)
+```
+
+assuming that the `tensorboard_plugin_widgets.plugin` module exposes a function `get_plugin` conforming to the interface described above.
+
+To facilitate plugin development without requiring plugins to be packaged into a wheel and installed into a virtualenv, TensorBoard will learn an `--extra_plugins` flag, whose value will be a list of comma-separated object references. For instance, for testing the widgets plugin, one might invoke
+
+```
+tensorboard --logdir /tmp/whatever \
+    --extra_plugins tensorboard_plugin_widgets.plugin:get_plugin
+```
+
+with the `tensorboard_plugin_widgets.plugin` module on the Python path.
+
+[entry-points]: https://packaging.python.org/specifications/entry-points/
+[tb-loader]: https://github.com/tensorflow/tensorboard/blob/dc71eeea403eb027612422ab6cd3061d4f9a10f3/tensorboard/plugins/base_plugin.py#L149-L162
+
+### Frontend binary loading
+One of our soft goals is to enable existing visualization suite to integrate with TensorBoard with ease without having to change, for instance, module system. Of course, a plugin has to procure data by writing a backend and some glue code to integrate with our backend, but we are hoping that most of the visualization part can remain untouched.
+
+#### Requirement: Notebook Integration
+TensorBoard recently added support for Jupyter and Colab and the feature is well received by the community. The new plugin frontend design should be compatible with the notebook integration.
+
+Of two prominent notebooks, Colab imposes greater limitations to our designs due to its security model: running in google.com domain, Colab sandbox and isolate an output cell from the main process using various techniques. Colab, specifically, employs ServiceWorker to proxy network request originating from output cell to a server running on the kernel but this imposes restriction the iframe usage; network request to fetch the document of the iframe is not intercepted by the ServiceWorker.
+
+#### Iframe based binary loading
+When a plugin is activated by a user, TensorBoard will create an iframe and point the iframe to the endpoint that plugin defined as an entry point.
+
+Iframe sandboxes JavaScript context and it gives plugin developers a lot of freedom in terms of frameworks -- with context disjoint from others, each plugin binary can incorporate as many dependencies as it would like to use without worrying about the correctness. This property is especially useful for Polymer-based applications because Polymer uses a global state, `CustomElementRegistry` (in v1, `document.registerElement`), and is not possible to load components with the same component name more than once. Loading plugins in its own separate frame will let plugin authors develop with ease and potentially have a standalone widgets for use outside of TensorBoard.
+
+It must be clearly stated that the use of global state is not unique to Polymer. Although JavaScript bundles use Immediately Invoked Function Expression (IIFE) and use closure to encapsulate dependencies, a bundler like Browserify or Webpack _can_ uses global states when shimming CommonJS `window.require` and use non-unique identifier for a global module.
+
+Despite the benefits and its simplicity, Colab makes this solution more complex -- because frames with different origins cannot be intercepted by Colab's ServiceWorker, we need some mechanism to route request to the non-publicly visible server running on the kernel. The TensorBoard team is working with Colab to introduce new Colab API that allows a port to be "exposed" to a stable URL (actual behavior is a lot more complex and is subject to changes) and the team is confident that it will make the iframe solution compatible with Colab.
+
+### Frontend binary
+Vulcanization today does not support CommonJS or packages from NPM. The team recognizes it to be an impediment for a plugin authorship and is unnatural for developing a frontend. TensorBoard team will provide a canonical build configuration that uses bazelbuild/rules_nodejs and Rollup but will not inhibit one from using Webpack and other bundlers.
+
+The frontend binary will be a collection JavaScript modules that implements certain interface -- one for configuring the dashboard (registerDashboard) and another for render. With exceptions of Polymer1 and Polymer2, modern frameworks and libraries either use JavaScript or a JavaScript based template system to create a DOM or a WebComponent and, thus, the requirement should be easy to fulfill.
+
+### Core module and UI/UX Consistency
+One of the most important key concepts when using TensorBoard is the data selection. TensorBoard uses runs and tags to filter potentially a large set of event logs to visualize. Most plugins implemented this today and we expect third party plugins to do the same.
+
+For ease and to promote maximal interaction and UI consistency, TensorBoard will provide a library that bundles Polymer components and encourage plugin authors to use the building blocks to build out a plugin. One of the components will be tf-runs-selector which will not only provide necessary controls but also tie with the TensorBoard core and make the selection consistent across all plugins. Note that the component can be used via Polymer or by using `document.createElement('tf-runs-selector')` if using other view libraries.
+
+### Plugins Misc
+#### Accessibility
+N/A: Using an iframe does not change behavior screen reader
+#### Instrumentation Considerations
+A plugin may decide to put instrumentations and there is no way for TensorBoard to enforce/disallow that.
+#### Docs and Tutorial
+How can a plugin show a helpful manual without asking user to visit the repository to review the README? Can TensorBoard assist by showing HTMLfied README in the browser?
+
+## Alternatives considered
+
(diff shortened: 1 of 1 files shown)


rfcs/20190411-tensorboard-improved-plugin-ext.md
```diff
@@ -0,0 +1,151 @@
...
+
+This RFC proposes a friction-free approach to enabling TensorBoard contributors to create and share TensorBoard plugins as installable extensions.
+
+TensorBoard is composed of various plugins that show up as separate dashboards to users (e.g., scalars, graphs, hparams, histograms, etc.). Today. anyone can contribute a new plugin by forking TensorBoard, adding their plugin, and getting the PR merged. However, this can require a lengthy review to make sure it is architected correctly and is aligned with TensorBoard’s goals.
+
+This RFC covers two changes that will make this process easier:
+1. Plugins can be distributed as Python packages. Users can install them as extensions to TensorBoard.
+2. Plugins are no longer limited to use Polymer but can instead use any frontend library or framework (e.g., React)
+
+The intention is to enable any prospective contributor to easily be able to create or reuse a visualization with tools familiar to them, and enable users to try it out quickly. Plugins that are useful to many users should still be added to TensorBoard, but this removes the friction needed to start getting feedback or trying out something new.
```
caisq: Might be good to clarify "added to TensorBoard" means "added to the codebase of TensorBoard and the tensorboard package directly" in this case, to avoid confusion with "added to TensorBoard via the 3p plugin route".

caisq: Grammar nit:
"to enable any prospective contributor to easily be able to"
-->
"to make it easier for any prospective contributor to" or
"to enable any prospective contributor to easily"

rfcs/20190411-tensorboard-improved-plugin-ext.md
```diff
@@ -0,0 +1,151 @@
...
+
+The TensorBoard frontend largely uses Polymer components (while more advanced components use D3, canvas, and WebGL; we do not have technical limitations in supporting a library like React today). TensorBoard repository comes with many shareable components (in `tf_dashboard_common`) that are essential building blocks most plugins and they promote consistency in UI and UX.
+
+## Design Proposal
+### Plugin distribution
+A plugin will be distributed as a single Pip distribution package that includes both a Python backend and a web frontend.
+
+The Python backend must export an object conforming to a standard interface that describes the plugin’s metadata and provides an option to load the plugin’s implementation. The exact shape is not yet specified, but it will be conceptually similar to a function returning a tuple containing a name, a description, and a [`tensorboard.plugins.base_plugin.TBLoader`][tb-loader] object.
```
caisq: > ... an object conforming to ...

Should it be an object or a function? The `get_plugin` example below is a function.

rfcs/20190411-tensorboard-improved-plugin-ext.md
```diff
@@ -0,0 +1,151 @@
...
+One of our soft goals is to enable existing visualization suite to integrate with TensorBoard with ease without having to change, for instance, module system. Of course, a plugin has to procure data by writing a backend and some glue code to integrate with our backend, but we are hoping that most of the visualization part can remain untouched.
+
+#### Requirement: Colab Integration
+TensorBoard recently added support for Jupyter and Colab and the feature is well received by the community. The new plugin frontend design should be compatible with Colab integration.
+
+One of the critical limitation Colab imposes TensorBoard is the usage of iframe and it is related to how it routes requests to the backend. To make it transparent to TensorBoard, Colab uses a ServiceWorker to intercept and reroute requests to localhost via its internal API. However, due to the implementation of Colab's ServiceWorker that is designed to intercept only requests to localhost (cannot use a relative path), one cannot render an iframe in an output cell.
+
+This restriction is relevant for *all solutions* below since a plugin can decide to render an iframe that makes a request anywhere.
+
+#### Non-iframe
```
caisq: > Non-iframe

Should this be listed as "Option 0: Non-iframe", so it can be clearer what the "Pro" and "Con" sections are talking about below?

rfcs/20190411-tensorboard-improved-plugin-ext.md
```diff
@@ -0,0 +1,151 @@
...
+Generally speaking, there are largely two types of solutions: one using iframe and one without using an iframe. Within the non-iframe based solution, there are several ways to achieve the goal and they are depicted below.
+
+One critical drawback of all non-iframe approaches is that TensorBoard cannot enforce plugin authors from inadvertently mutating the global object. For instance, if a library that a plugin transitively depends polyfills and monkey-patches a global object, it can influence how other plugins behave.
+
+**Pro**
```
caisq: See my comment above. The structure of the text around this point is somewhat hard to follow. It's unclear 
1. what the Pro and Con sections refer to
2. why these sections exist for some of the options but not others

rfcs/20190411-tensorboard-improved-plugin-ext.md
```diff
@@ -0,0 +1,151 @@
...
+
+A difficulty arises with how ServiceWorker behave with respect to an iframe. A ServiceWorker can only intercept requests that originates from a document within its scope. A ServiceWorker, derived from its module source, is associated to a scope (for instance, if the ServiceWorker module originates from tensorflow.org/foo/bar/sw.js, it will have a scope of "tensorflow.org/foo/bar" and it can only intercept request from a document whose host starts with "tensorflow.org/foo/bar/"). Now, to render an iframe whose content source is backed by an endpoint running in a VM, we need to go through the Colab proxy by setting the source of the iframe as "https://localhost/*" (untrue; will be discussed below). This, though, is a problem to a request that originates within the iframe -- since iframe's origin is localhost, only ServiceWorkers with scope of "localhost" will be active and Colab's proxy will not be effective.
```
caisq: > "https://localhost/*"

nit: use backtics to surround this. Or all the font becomes italic after this point (thanks to markdown renderer....)

rfcs/20190411-tensorboard-improved-plugin-ext.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Design Proposal
+### Plugin distribution
+A plugin will be distributed as a single Pip distribution package that includes both a Python backend and a web frontend.
+
+The Python backend must export an object conforming to a standard interface that describes the plugin’s metadata and provides an option to load the plugin’s implementation. The exact shape is not yet specified, but it will be conceptually similar to a function returning a tuple containing a name, a description, and a [`tensorboard.plugins.base_plugin.TBLoader`][tb-loader] object.
+
+In order to be discovered by TensorBoard, the distribution should include an [`entry_points` entry][entry-points] declaring itself as a TensorBoard plugin. The entry point group is `tensorboard_plugins`. The entry point name is arbitrary, and ignored. The entry point object reference should refer to the standard interface described in the previous paragraph. As an example, the setup.py file for a hypothetical “widgets” plugin might look like:
```
caisq: > standard interface describe in the previous paragraph

I can't find a detailed description of any interface in the preceding paragraph. Am I missing something? I see you said "The exact shape is not yet specified." But can you at least come up with an outline? There is already such a plugin registration method in the existing TensorBoard plugin system. Is the new interface expected to be similar or significantly different. I feel this RfC should at least include some rudimentary information on this topic.

rfcs/20190411-tensorboard-improved-plugin-ext.md
```diff
@@ -0,0 +1,151 @@
...
+#### Docs and Tutorial
+How can a plugin show a helpful manual without asking user to visit the repository to review the README? Can TensorBoard assist by showing HTMLfied README in the browser?
+
+## Alternatives considered
+### Iframe and Inter-frame communication
+Using XMLHttpRequest, TensorBoard can fetch the content of an iframe as text and use `iframe.srcdoc` to render the content. This iframe, however, will have origin of `null` and its requests will not be intercepted by any ServiceWorker.
+
+Instead, TensorBoard can provide a library one of which will be a `request`. The method will use `window.postMessage` to route the network request to the main frame and hide away the fact that it is using inter-frame communication at all. This solution, though easy for XMLHttpRequest based request, is difficult to apply for, for example, image, audio, video, CSS, web-fonts, EsModule, and WebSocket requests. To illustrate the problem more concretely, when `<img src="foo.png">` is created, a browser initiates the requests and it is not possible to route that request via the library. This will make a plugin like Images plugin difficult to write.
+
+## Questions and Discussion Topics
```
caisq: Big missing pieces I see in this draft

1. What is the signature of `get_plugin()`, i.e., how tensorboard discovers the 3p plugin? See my comment about the "standard interface" above.
2. Apart from the standard interface, are there going to be any new protocols for any of the following?
  - data storage in the backend (e.g., tensorboard core can provide a unified database write/read API)
  - communication protocol between the plugin's frontend component and the backends (e.g., tensorboard can provide an NPM package that helps the frontend component manage the requests to the backend).

bhack:
There were some general comments related to plugins/modularities in the Tensorflow addons RFC starting from https://github.com/tensorflow/community/pull/84#issuecomment-476353018.
I hope that we can include a solution to handle quality fragmentation also here.

rfcs/20190411-tensorboard-improved-plugin-ext.md
bileschi: +1

Building on the discussion of the communication protocol, Does this illustration capture how you expect the pattern of communication to work?  Or do you rather expect the TensorBoard FE to pass an address to the Plugin FE so that the Plugin FE can talk to the backend components directly?


image

rfcs/20190411-tensorboard-improved-plugin-ext.md
stephanwlee: @GalOshri did we mean that it should be part of the "first-party plugin" if it is good?

rfcs/20190411-tensorboard-improved-plugin-ext.md
stephanwlee: I am not sure what you mean here. Non-iframe is a class of solution which has two specific solutions under `option 1` and `option 2`.

rfcs/20190411-tensorboard-improved-plugin-ext.md
stephanwlee: I am failing to follow :(

image

rfcs/20190411-tensorboard-improved-plugin-ext.md
GalOshri: Yes, @caisq's clarification is useful. I think we'd need to think through these details, but the idea is that if some functionality is super useful to most users, it should be a part of the default experience.

rfcs/20190411-tensorboard-improved-plugin-ext.md
```diff
@@ -0,0 +1,151 @@
...
+with the `tensorboard_plugin_widgets.plugin` module on the Python path.
+
+[entry-points]: https://packaging.python.org/specifications/entry-points/
+[tb-loader]: https://github.com/tensorflow/tensorboard/blob/dc71eeea403eb027612422ab6cd3061d4f9a10f3/tensorboard/plugins/base_plugin.py#L149-L162
+
+### Frontend binary loading
+One of our soft goals is to enable existing visualization suite to integrate with TensorBoard with ease without having to change, for instance, module system. Of course, a plugin has to procure data by writing a backend and some glue code to integrate with our backend, but we are hoping that most of the visualization part can remain untouched.
+
+#### Requirement: Colab Integration
+TensorBoard recently added support for Jupyter and Colab and the feature is well received by the community. The new plugin frontend design should be compatible with Colab integration.
```
ewilderj: To be explicit, you also intend to preserve Jupyter compatibility too? If so, maybe change the heading to "Requirement: Colab and Jupyter Integration"

rfcs/20190411-tensorboard-improved-plugin-ext.md
wchargin: Functions are objects in Python.\* We use the term “object” here
generically (as Python does).

\* `isinstance(lambda: None, object)  # True`

rfcs/20190411-tensorboard-improved-plugin-ext.md
wchargin: Correct, the shape is not yet specified; the outline is, “conceptually
similar to a function returning a tuple containing a name, a
description, and a loader object”. Maybe it’ll just end up being a
function that returns a plugin loader, if we determine that we don’t
need any extra metadata.

The existing plugin system requires you to write your own `main` module
to register plugins, whereas this new system will not.

We haven’t given a ton of thought to the details of this interface,
because we expect it to be simple for anyone to implement. We’ll pick
something and specify it clearly, and people can just use it.

rfcs/20190411-tensorboard-improved-plugin-ext.md
stephanwlee: Thanks for the feedback. I changed it to "notebook integration" and explained how Colab differs from others.

rfcs/20190411-tensorboard-improved-plugin-ext.md
stephanwlee: @bileschi thanks for the feedback. At this moment, we would like to confine the scope of design to redo the distribution strategy. The architecture you are describing has more profound impact than depicted in the picture and should be redesigned holistically (including the summary system and storage system. Note that the storage should also be pluggable: data format, storage mechanism, and etc...).

Before closing it, I would also like to note that close coupling between frontend and backend (and distributed as one unit) alleviates the [concern of fragmentation](https://github.com/tensorflow/community/pull/90#issuecomment-482559006) raised below. Though hard to avoid duplication of features and varying level of quality, current plugin system allow us to do without API versioning and conflict resolutions.

bileschi: Thanks for the feedback!  Can you provide a substitute diagram which describes the proposed communication strategy?

rfcs/20190411-tensorboard-improved-plugin-ext.md
davidsoergel: +1.  What's confusing here is that the options 1, 2, and 3 are at different levels of the hierarchy.  IIUC, the intent is:

- Option A: Non-iFrame
    - Pro, con
    - Option A1
    - Option A2
- Option B: iFrame
    - Pro, con

rfcs/20190411-tensorboard-improved-plugin-ext.md
```diff
@@ -0,0 +1,153 @@
...
+
+Within the non-iframe based solution, there are about two options to achieve load plugins. Two solutions share strengths and weaknesses and they are:
+
+**Pro**
+- easy to support
+- supports Colab as long as plugins do not use iframes
+
+**Con**
+- global mutating libraries like Polymer app will not work
+- can lead to an awkward UX: e.g., different version of a visualization library have different UI/UX
```
davidsoergel: I'm not sure I follow-- can you explain this concern further or give an example?

rfcs/20190411-tensorboard-improved-plugin-ext.md
```diff
@@ -0,0 +1,153 @@
...
+#### Option 3: iframe
+**Pro**
+- ease of development. Can bring your own version of Polymer, React, and etc...
+- encapsulation and containment of plugins accidentally doing wrong things
+
+**Con**
+- does not work in Colab
+- generally larger binaries
+- Cannot reuse components across plugins
+- can lead to an awkward UX: e.g., different version of a visualization library have different UI/UX
```
davidsoergel: This a Con of both options?  Is there any alternative that avoids it, then?

rfcs/20190411-tensorboard-improved-plugin-ext.md
davidsoergel: I also strongly believe that we should reconsider the architecture of providing plugins as coupled (Backend, Frontend) pairs.

At the same time, I think that problem is somewhat separable from the current proposal.  That is, this RFC goes one step in the right direction; we'd like to discuss additional related steps soon, but they don't need to block this one.

In particular I think we can easily get consensus on these bits:

* Plugins providing BE Python code are dynamically loaded at runtime.
* BE plugins are distributed via PIP.

While these aspects of the current architecture (which this RFC retains) should remain open topics:

* Plugins bundle FE code.
* The TensorBoard backend is responsible for assembling the FE app from the available plugins and serving it.
* Plugin network APIs are not formalized or versioned.

The reason this is coming up now is that most of this doc is specifically about how the plugin FEs should be assembled into the FE app that TB serves.  Under a decoupled architecture in which the FE is built and served as an independent entity (presumably, from components distributed via npm, bundled with webpack or similar, etc.), the discussion around whether or not to use iFrames etc. would be very different.

The upshot is I think it's OK to go ahead with this proposal, because the dynamic BE plugin loading (and PIP distribution of those) is great in any case.  It may turn out that the FE-assembly questions here will be obviated by a later design of broader scope; let's give that the attention it deserves in a separate RFC.

davidsoergel: @bileschi As I understand it, the current proposal does not change the existing communication strategy, which is that a plugin FE makes requests only to routes that its matched BE exposes (or to standard routes that the TB core provides).  See https://github.com/tensorflow/tensorboard/blob/master/tensorboard/http_api.md.  There is no client-side TB code mediating those requests, and no client-side state shared between FE plugins.  (Those are certainly good things to discuss; they're just outside the present scope.)

stephanwlee:
Due to uncertainty with the Colab + iframe story, we are going to push out the deadline few days til May 6th. Thanks for all the comments so far!

stephanwlee:
Sorry for the delay in closing the RFC. Prototype with Colab took longer than expected but everything is more or less looks good now. I have updated the proposal accordingly (the iframe based solution wins).

Since there was no major disagreement generally and especially since there wasn't much disagreement on how plugins are loaded, where we spent extra time prototyping with Colab, we will bake the proposal few more days and try to close it on May 19th. Thanks!

stephanwlee:
Alright, we have let it bake for awhile and I think it is good to close this.
@ewilderj can you help us with merging this? Thanks!

ewilderj:
Please push an update changing the status from Proposed to Accepted, and I will merge. Many thanks!

stephanwlee:
@ewilderj done! Thanks a lot!

PR #95, opened 2019-04-25
RFC: Keras Pre-made Models

tanzhenyu:
**Feedback period will be open until 2019-05-10**

# Keras pre-made models

| **Author(s)** | Zhenyu Tan (tanzheny@google.com)|
| **Sponsor**   | Francois Chollet (fchollet@google.com), Alexandre Passos (apassos@google.com) |
| **Updated**   | 2019-04-29                                           |

## Objective

This document proposes several pre-made Keras models that would allow users to:
* build basic machine learning models easily
* compose them with other keras Layers
* replace [Canned Estimators](https://www.tensorflow.org/guide/premade_estimators) in TF 2.0.

### DIFF
--- a/20190425-keras-premade-models.md
+++ b/20190425-keras-premade-models.md
@@ -1,0 +1,380 @@
+# Keras pre-made models
+
+| Status        | Proposed                                             |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | Zhenyu Tan (tanzheny@google.org)|
+| **Sponsor**   | Francois Chollet (fchollet@google.org), Alexandre Passos (apassos@google.com) |
+| **Updated**   | 2019-04-25                                           |
+
+## Objective
+
+This document proposes several pre-made Keras models that would allow users to:
+* build basic machine learning models easily
+* compose them with other keras Layers
+* replace [Canned Estimators](https://www.tensorflow.org/guide/premade_estimators) in TF 2.0.
+
+## Motivation
+
+Scikit-learn is the de facto framework for creating basic machine learning models. From tensorflow perspective,
+it is beneficial to provide users several important basic models to possibly support the following features:
+* distributed mini-batch training
+* accelerated training (GPU & TPU)
+* enable AutoML to select different pre-made models
+* enable users to freely compose pre-made models with other Keras layers and models.
+
+Currently most pre-made models from tensorflow have been exposed through [CannedEstimator](https://www.tensorflow.org/guide/premade_estimators). However estimator
+* relies on continuous graph rebuilding and checkpoints reloading which slows down training
+* relies on global collections and not TF 2.0 friendly.
+* makes many advanced features such as meta/transfer learning difficult
+* enforces user to create input functions when not necessary
+
+Given the status, we propose the following pre-made models as a start:
+1. Linear models including Linear Regression and Logistic Regression
+2. DNN models including DNN Regression and DNN Classification
+3. Wide-Deep models including Wide-Deep Regression and Wide-Deep Classification
+4. BoostedTree models including BoostedTree Regression and BoostedTree Classification
+
+Users can use [feature_columns](https://www.tensorflow.org/guide/feature_columns) with pre-made models for feature engineering. We use **pre-made** and **canned** interchangeably in this doc.
+
+Several side notes: This document should serve as a foundation for other pre-made models that are not included in this proposal, such as Kmeans, random forest, collaborate filtering, which are yet to come; Detailed design regarding implementation details for BoostedTree, such as distributed training, saving & restoring, acceleration, supported feature columns should live in separate docs. 
+
+## Design Proposal
+
+### Challenges
+Many canned models, including BoostedTrees, KMeans (as well as WideDeep and many others more in the future) are highly complicated and do not follow the simple foward & backprop workflow, an assumption heavily relied on by Keras. Building such models, while not compromising the basic composability of Keras (i.e., compose layers on top of layers), is the main challenge we need to resolve in this document. Another challenge is that these models can have multiple training phases (such as collaborative filtering models). We propose the following two approaches to address them:
+
+### Proposal 1: Customized training function & composability
+We propose to let each subclassed pre-made model override the training function. It is optional to provide a special subclass `CannedModel` if other methods such as `compile` and `fit` needs to be overriden as well. In traditional Keras models, such training function is dominated by autodiff - given the forward pass of the model, the gradients for each Operation is generated and backprop computation is automatically laid out for the entire model. However such assumption is only valid for neural network based supervised learning architecture. For many other scenarios, we would need to break this assumption:
+1. gradients may not be used, e.g., any un-supervised learning tasks
+2. gradients may only be used partially, e.g., tree based models
+3. both gradients and higher order statistics are used for training, e.g., linear model with BGFS algorithm that approximates hessian
+4. only gradients are used, but the backprop path is composed of customized training path, e.g., wide-deep model
+5. primal-dual formulation, e.g., SDCA linear model, and constrained optimization algorithms in general.
+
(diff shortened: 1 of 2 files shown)


20190425-keras-premade-models.md
```diff
@@ -0,0 +1,380 @@
...
+
+### Proposal 1: Customized training function & composability
+We propose to let each subclassed pre-made model override the training function. It is optional to provide a special subclass `CannedModel` if other methods such as `compile` and `fit` needs to be overriden as well. In traditional Keras models, such training function is dominated by autodiff - given the forward pass of the model, the gradients for each Operation is generated and backprop computation is automatically laid out for the entire model. However such assumption is only valid for neural network based supervised learning architecture. For many other scenarios, we would need to break this assumption:
+1. gradients may not be used, e.g., any un-supervised learning tasks
```
seanpmorgan: The subclass `CannedModel` seems a like a half measure. If the training function is heavily modified it seems strange not to have an object type that describes that. Perhaps `IrregularModel` which would specify models that do not follow the typical Keras paradigm. 

The canned DNN model is a `CannedModel` imo, but not an irregular model.

seanpmorgan: The type `CannedModel` also makes it sound as if users are not supposed to be building their own unique model types. To me canned mean pre-made.

AakashKumarNain:
Because the RFC is about `Pre-made` models, I think this should also be considered in the RFC:
https://groups.google.com/a/tensorflow.org/forum/#!topic/developers/QMCbkS6uZSM

unrahul:
Could you give a clear example of what you mean by pre-made models, I am not sure if i clearly understand the difference between pre-made and extensible. Does this proposal mean, we can have `estimator` like models that are extensible for users to modify if needed otherwise use as is ?

20190425-keras-premade-models.md
```diff
@@ -0,0 +1,380 @@
...
+
+Several side notes: This document should serve as a foundation for other pre-made models that are not included in this proposal, such as Kmeans, random forest, collaborate filtering, which are yet to come; Detailed design regarding implementation details for BoostedTree, such as distributed training, saving & restoring, acceleration, supported feature columns should live in separate docs. 
+
+## Design Proposal
+
+### Challenges
+Many canned models, including BoostedTrees, KMeans (as well as WideDeep and many others more in the future) are highly complicated and do not follow the simple foward & backprop workflow, an assumption heavily relied on by Keras. Building such models, while not compromising the basic composability of Keras (i.e., compose layers on top of layers), is the main challenge we need to resolve in this document. Another challenge is that these models can have multiple training phases (such as collaborative filtering models). We propose the following two approaches to address them:
+
+### Proposal 1: Customized training function & composability
```
terrytangyuan: What about other phases like evaluation and prediction? What's the difference between this and `tf.estimator`'s `model_fn`?

20190425-keras-premade-models.md
tanzhenyu: It's probably better not to have `CannedModel` for now, and let's just override train_function (or train_on_batch) -- WDYT?

20190425-keras-premade-models.md
tanzhenyu: evaluation and prediction are specified in Mode in Estimator world. In Keras model will by default create train_function, evaluate_function and predict_function and use the corresponding one during fit/eval/predict.

20190425-keras-premade-models.md
```diff
@@ -0,0 +1,380 @@
...
+* distributed mini-batch training
+* accelerated training (GPU & TPU)
+* enable AutoML to select different pre-made models
+* enable users to freely compose pre-made models with other Keras layers and models.
+
+Currently most pre-made models from tensorflow have been exposed through [CannedEstimator](https://www.tensorflow.org/guide/premade_estimators). However estimator
+* relies on continuous graph rebuilding and checkpoints reloading which slows down training
+* relies on global collections and not TF 2.0 friendly.
+* makes many advanced features such as meta/transfer learning difficult
+* enforces user to create input functions when not necessary
```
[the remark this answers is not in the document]
tanzhenyu: We have two sets of high level APIs and Keras is the one that we choose to proceed because of its simplicity and pythonic implementation.

tanzhenyu:
> Could you give a clear example of what you mean by pre-made models, I am not sure if i clearly understand the difference between pre-made and extensible. Does this proposal mean, we can have `estimator` like models that are extensible for users to modify if needed otherwise use as is ?

@unrahul I'm not sure why do you mean by `estimator` like models. But pre-made in this case is similar to keras-application models, you can compose them with any other layers and models. You can modify it by inheritance yes.

tanzhenyu:
> Because the RFC is about `Pre-made` models, I think this should also be considered in the RFC:
> https://groups.google.com/a/tensorflow.org/forum/#!topic/developers/QMCbkS6uZSM

@AakashKumarNain Not sure I understand the correlation -- from the discussion it looks like that is related to training/testing for variables. The models in this proposal do not have any variables in that property.

AakashKumarNain:
@tanzhenyu Yeah I got it but then I think the name for RFC should have been something else. Anyways, thank you for clarification again

20190425-keras-premade-models.md
seanpmorgan: Agree that's better than an optional subclass. I would probably still prefer a new Model type so that it's understood how different these are than typical Keras Models, but something like `IrregularModel` may just be more bloat to the API than its worth.

tanzhenyu: I'm keeping this as a design question to be addressed through design reviews.

KesterTong:
Can we add some discussion on integration with ProcessingLayers, which should also help clarify what the inputs to a CannedModel are?

@fchollet and I have brainstormed the following two options:

Option 1: Models accept a list of tensors as inputs, or pair of lists for DNNLinearModel, where the pair is the list of DNN and linear inputs respectively.

```
age = tf.keras.Input(shape=(1,), dtype=tf.float32, name='age')
bucketized_age = tf.keras.Discretize(age, bins=4)(age)
occupation = tf.keras.Input(shape=(None,), dtype=tf.string, name='occupation')
occupation_id = tf.keras.VocabLookup(vocabulary_list)(occupation)
occupation_embed = tf.keras.layers.Embedding(output_dim=10)(occupation_id)
processing_stage = tf.keras.ProcessingStage(
    inputs=[age, occupation], outputs=[bucketized_age, occupation_embed])
# for DNNLinearModel:
# processing_stage = tf.keras.ProcessingStage(
#     inputs=[age, occupation],
#     outputs=([bucketized_age, occupation_embed], [occupation_id]))

canned_model = tf.keras.canned.LinearClassifier()
# for DNNLinearModel:
# canned_model = tf.keras.canned.DNNLinearClassifier()
output = canned_model(processing_stage.outputs)
model = tf.keras.Model(inputs=processing_stage.inputs, outputs=[output])

dftrain = pd.read_csv('storage.googleapis.com/tf-datasets/titanic/train.csv')
y_train = dftrain.pop('survived')
ds = tf.data.Dataset.from_tensor_slices(dict(dftrain), y_train)
processing_stage.update(ds)
model.train(ds, epochs=10)
```

Option 2: Instead of accepting a list of tensors (or pair of lists for DNNLinearModel), models accept a single tensor (or pair of tensors for DNNLinearModel).  The input represents the concatenated features.  

```
age = tf.keras.Input(shape=(1,), dtype=tf.float32, name='age')
bucketized_age = tf.keras.Discretize(age, bins=4)(age)
occupation = tf.keras.Input(shape=(None,), dtype=tf.string, name='occupation')
occupation_id = tf.keras.VocabLookup(vocabulary_list)(occupation)
occupation_embed = tf.keras.layers.Embedding(output_dim=10)(occupation_id)
processing_stage = tf.keras.ProcessingStage(
    inputs=[age, occupation],
    outputs=tf.keras.layers.concatenate([bucketized_age, occupation_embed]))
# for DNNLinearModel:
# processing_stage = tf.keras.ProcessingStage(
#     inputs=[age, occupation],
#     outputs=(tf.keras.layers.concatenate([bucketized_age, occupation_embed]), occupation_id))

canned_model = tf.keras.canned.LinearClassifier()
# for DNNLinearModel:
# canned_model = tf.keras.canned.DNNLinearClassifier()
output = canned_model(processing_stage.outputs)
model = tf.keras.Model(inputs=processing_stage.inputs, outputs=[output])

dftrain = pd.read_csv('storage.googleapis.com/tf-datasets/titanic/train.csv')
y_train = dftrain.pop('survived')
ds = tf.data.Dataset.from_tensor_slices(dict(dftrain), y_train)
processing_stage.update(ds)
model.train(ds, epochs=10)
```

PR #100, opened 2019-05-01
RFC: TensorCord Variant Object

ebrevdo:
**Comment period is open until 2019-05-15**

# RFC: TensorCord Variant Object

| **Author(s)** | Eugene Brevdo (ebrevdo@google.com) |
| **Sponsor**   | Alex Passos (apassos@google.com)                 |
| **Updated**   | 2019-05-01

## Objective

This document proposes a new Variant object called TensorCord.  A TensorCord
contains a list of `absl::string_view` and release functors / tensor references; and
can be used to store references to other Tensors or other memory locations.  The
destructor of a TensorCord reduces the count on its referenced releasers; once
the reference count of a releaser reaches 0, it is executed.

### DIFF
--- a/rfcs/20190501-tf-tensor-cord.md
+++ b/rfcs/20190501-tf-tensor-cord.md
@@ -1,6 +1,6 @@
 # TensorCord Variant Object
 
-| Status        | (Proposed / Accepted / Implemented / Obsolete)       |
+| Status        | Proposed       |
 :-------------- |:---------------------------------------------------- |
 | **Author(s)** | Eugene Brevdo (ebrevdo@google.com) |
 | **Sponsor**   | Alex Passos (apassos@google.com)                 |
@@ -39,8 +39,8 @@ transparent manner).
 ## Design Proposal
 
 The TensorCord object itself can be implemented using RefCounted objects with a
-constructor that takes an absl::string_view and either a releaser callback or
-a pointer to a Tensor or RefCounted object.
+constructor that takes an `absl::string_view` and either a releaser callback or
+a pointer to a `Tensor` or `RefCounted`.
 
 Below is an example of use:
 
@@ -67,10 +67,11 @@ constructor so users don't accidentally create a `DT_CORD` tensor without
 a releaser.
 
 | Alternatives | TensorCord DT_VARIANT | DT_CORD |
+:------------- |:--------------------- |:------- |
 | Separate releasers per element | Yes | No |
 | Overhead | Higher (each Tensor element keeps a reference; Variant & RunVariantDtor overhead is more costly) | Lower (can have onereleaser per tensor) |
 | Intrusiveness | Lower (use DT_VARIANT) | Higher (add new TF type) |
-| Flexibility | High (elements can point to locations backed by different owners) | Lower (all elements must be backed by data whose lifetime depends a shared set of releasers) |
+| Flexibility | High (elements can point to locations backed by different owners) | Lower (all elements must be backed by data whose lifetime depends a shared set of releasers)
 
 ## Detailed Design
 
@@ -108,17 +109,21 @@ public:
   // is small enough, no reference is created on `tensor`; instead the memory is
   // stored inline.
   explicit TensorCord(absl::string_view view, Tensor* tensor);
+  explicit TensorCord(absl::string_view view, RefCounted* ref_counted);
 
   void Append(const TensorCord& other);
   void Append(absl::string_view view, CordRep::Releaser releaser,
                         void* memory = nullptr);
   void Append(absl::string_view view, Tensor* tensor);
+  void Append(absl::string_view view, RefCounted* ref_counted);
 
   size_t size() const;
   bool empty() const;
 
   explicit operator string() const;
 
+  // Usage example:
+  //  for (absl::string_view s : cord.Chunks()) { ... }
   ChunkRange Chunks() const;
   ChunkIterator chunk_begin() const;
   ChunkIterator chunk_end() const;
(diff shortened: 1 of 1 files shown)


byronyi:
This would benefit a lot for messaging heavy distributed jobs, in particular CollectiveAllReduce dist-strat.

Can’t wait to see it implemented in TF.

rfcs/20190501-tf-tensor-cord.md
```diff
@@ -0,0 +1,244 @@
...
+
+## Design Proposal
+
+The TensorCord object itself can be implemented using RefCounted objects with a
+constructor that takes an `absl::string_view` and either a releaser callback or
+a pointer to a `Tensor` or `RefCounted`.
+
+Below is an example of use:
+
+```python
```
kirillbobyrev: nit: this should probably be "c++" instead of "python"
```suggestion
```с++
```

ebrevdo: looks like you may need to sign a CLA to get this change through... sorry!

rfcs/20190501-tf-tensor-cord.md
```diff
@@ -0,0 +1,244 @@
...
+
+  // Copy and move constructor, copy and move assignment operators.
+  // And all the associated Variant-stored object methods
+  // (Encode, Decode, DebugString, etc).
+};
+```
+
+### Ops and Op extensions supporting TensorCord:
+
+The following ops would be extended to support TensorCord:
```
jharmsen: Could you give an idea of what it would take to extend an Op to support TensorCord?
E.g., is extending `join` as simple as templatizing the Op for both string and TensorCord?

rfcs/20190501-tf-tensor-cord.md
kirillbobyrev: Uh, my bad, I had the CLA signed and I was certain I had it but now I realized it was for before I changed my username.

gregbillock:
We're excited about the prospects this has to speed up text handling for text models. There are lots of situations where being able to manage strings using string-view type constructs with the occasional owned string (e.g. a normalized version) will be a big performance boost.

rfcs/20190501-tf-tensor-cord.md
ebrevdo: I'll update the doc with a corresponding code example.  It's pretty similar to what you said: you first templatize the Op for both DT_STRING and DT_VARIANT.  Then in the kernel you check the input dtype; if it's Variant, you access an Eigen tensor of Variant objects.  From each variant in the Tensor, you access the underlying TensorCord via `get<TensorCord>()`.  If the pointer is nullptr, error out (because the stored object is not a TensorCord); otherwise access it and perform the join.

For the specific case case of the join op, you may actually prefer to emit a new TensorCord Variant Tensor containing the joined strings as sub-Cords (converting to string would perform the concatenation).

yongtang:
This could really help input parsing for different file formats. We very much look forward to seeing it in tensorflow repo soon 👍

thuang513:
Design Review Notes:

Q: What has changed since it has sent out?
Nothing, other than changes have been submitted to test out/prototype the proposal.

Proposal summary: variant object for strings, string ops also add support for new variant 

Q: Why do we need this?
Some string ops (e.g. tokenization, sentence breaking, string_split) are expensive because they make many copies. Their operation can be more efficient if they are able to emit string_views (pointer, data_size) instead of copying and release new strings.

In many of the applications, we only care about the string views up to the point that we do an embedding lookup in the input_fn then we can get out.

We did an analysis and found that emitting thousands of ref objects is not that expensive.

Q: Who handles the memory management?
TensorBuffer can have global memory management,  where we can handle the ownership/keep track of the ref counts of strings.

Changes are needed to work with Eigen, but since Eigen does nothing for string, we’re okay with dropping it.

Some changes needed to the allocate_output() API to propgate refcounts/dependencies when allocating new tensors.

Q: What’s in a reference?
a pointer to a destructor
a pointer to a memory location.
Need destructor and pointer per tensor (at least, likely per string).

Q: The design has similar goals to tf-string RPC, can we try to merge and consolidate these two proposals/efforts in one refactor?
Yes. The main roadblock was get a main use case. Design is done, but downstream changes and backward compatible is not there yet.

We can replace kPreallocType, which has a limited use case, with kStringView, which would denote a tstring scalar that does not own the underlying memory for the string array. 


With the tstring proposal---like in std::string with SSO---a kSmallType converts into a kLargeType when assigned with a string > (sizeof(tstring) - 1).  Similarly, for kOffsetType, any writes larger than its original capacity would result in a conversion to a kLargeType type and a heap allocation.

Q: What if we have access to the original string (likely as a Cord)? Can we do a zero copy import? (e.g. from an RPC, record from disk, etc)

Current proposal has to import/copy it, but if this is an important use case we should reconsider this implementation and come up with other options (e.g. more space to store pointers) that would allow zero copy on import. STRING_JOIN will also need to be copied, which is no different than today.

We could potentially increase the sizeof(tstring), in order to allow for more pointers, and allow for tstrings to store a Cord, but it is not clear that the benefits are worth the added complexity.

Q: Can we do a zero copy from python to TF?
We could do this as long as the returned memory is aligned via tcmalloc. Externally, numpy has to compile with an allocator, maybe ask if numpy can do this? Otherwise a copy on import is still needed.

Q: next steps?
Make a strawman CL that modifies tensor allocator
changes to new tstring design
changes to TensorBuffer stub that manages ownership
Need to rewrite kernels to allocate and emit Tensors with TensorBuffers that propagate refcounts/ownership.

ebrevdo:
As a followup: the proposal to merge tstring with TensorCord was reversed because of the problems this would create for users of tstring:

* tstring would no longer represent a single contiguous string; so operations like array-type access and cast to string_view would no longer be possible.  similarly, the "fast path" that most programs rely on when dealing with string buffers, by accessing buffer.data(), would on longer be available.

* the idea of separating the view (a tstring with multiple pointers to memory it has no control over) from the deallocation operation (which would sit elsewhere in a TensorBuffer) was problematic.

PR #101, opened 2019-05-06
RFC: Modular Filesystems C API

mihaimaruseac:
**Feedback period will be open until 2019-05-27**

# Modular Filesystems C API

 | Status        | Accepted                                             |
| **Author(s)** | Mihai Maruseac (mihaimaruseac@google.com)            |
| **Sponsor**   | Gunhan Gulsoy (gunan@google.com)                     |
| **Updated**   | 2019-05-06                                           |

## Summary

Define the API for the plugin-based implementation of filesystems under modular TensorFlow (as proposed in #77)

A major module of the modular TensorFlow involves all operations regarding files and the filesystem. Saving models, checkpointing, reading files as input to models (i.e. images, audio, etc.), and many other file manipulations are all tasks that TensorFlow has to support.

While users can use the underlying operating system’s API (such as calling fopen and friends on Posix environments, for example), it is definitely better to have a cross-platform, cross-environment API that can be used everywhere. TensorFlow provides such an API at the moment but it results in compiling code for all known filesystems in the final binary. We propose a new design where all filesystems are implemented as plugins and only the required ones are added at runtime. To do so, we will first need to design a C API for the filesystems and then integrate this with the rest of the modularization effort.


rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+1. Interface between TensorFlow and plugins must be as simple as possible as
+   well as thoroughly tested. It is better to localize errors in either
+   TensorFlow core or the plugin providing the filesystem functionality than in
+   the glue code between the two.
+
+1. Plugin registration should support mandatory metadata such as version
+   numbers, checks for backwards and forwards compatibility as well as optional
+   fields such as author information, bug reporting URL, etc.
+
+1. The mandatory metadata must guarantee _binary compatibility_: allow plugins
```
alextp: I think this needs to be clearer about the amount of binary and source compatibility we provide.

I also don't think we should break source compatibility outside of major versions, which is fairly easy to do if we always add new API entry points instead of editing / removing existing ones.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+on ensuring a better development workflow by a team-wide effort designed to make
+TensorFlow more modular according to the [current RFC][modular_rfc].
+
+A major module of the modular TensorFlow involves all operations regarding files
+and the filesystem. Saving models, checkpointing, reading files as input to
+models (i.e. images, audio, etc.), and many other file manipulations are all
+tasks that TensorFlow has to support.
+
+While users can use the underlying operating system’s API (such as calling fopen
+and friends on Posix environments, for example), it is definitely better to have
```
alextp: nit: I'd mention cloud environments here as many people's thoughts when reading about filesystems will be around things like ext4 and xfs for which we do not need a layer

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+
+To allow both the core and the plugin to know which methods from the tables are
+supported, we also add the `num_ops` field, counting how many methods are
+defined in the table. This field can be used to determine the amount of
+available API calls, as it is guaranteed to only increase during minor releases.
+We provide a `NUM_OPS_IN_TABLE` macro to allow the compiler to set the value of
+`num_ops` automatically.
+
+Alternatively, instead of `num_ops` field we can have a size field which is the
+size of the entire structure and can be obtained via `sizeof`. The benefits of
```
alextp: The size field makes it easier to be memory-safe, if you do choose to go with exposing the struct to users.

20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+
+Instead, we provide 3 different data structures to act as function tables, to
+hold function pointers for files and the filesystem. Currently, these structures
+are filled with the functionality already present in the existing filesystem
+API:
+
+```cpp
+// Operations on a TF_RandomAccessFile
+typedef struct TF_RandomAccessFileOps {
+  size_t (*const Read)(const TF_RandomAccessFile*, uint64, size_t, char*, TF_Status*);
```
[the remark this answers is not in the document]
mihaimaruseac: I don't think so, the filesystem state is internal to the filesystem, which will be retrieved by the filesystem implementation from the file structure itself. I'll clarify this too

20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+TF_CAPI_EXPORT extern void TF_InitPlugin(TF_Status* status);
+```
+
+This function must allocate all of the above tables and then call a function
+provided by the core TensorFlow to register plugin filesystems:
+
+```cpp
+TF_CAPI_EXPORT extern void RegisterFileSystemPlugin(
+    const string& plugin, const string& scheme,
+    const TF_Filesystem_Plugin_Metadata* metadata,
```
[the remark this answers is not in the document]
mihaimaruseac: I'll need to test first that this won't cause compatibility issues. I also think that passing them by value results in extra copies, but we need at least one copy anyway so probably this doesn't matter

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+Similarly, we will have to implement a subclass of the `FileSystem` class,
+`ModularFileSystem`, in the same vein.
+
+Another change we have to do is to `FileSystemRegistry`: the `Register()` call
+now uses a `unique_ptr<FileSystem>` argument instead of the existing
+`Factory`-based implementation since the constructor of `ModularFileSystem`
+needs to take as argument the function tables for the operations supported by
+the filesystem.
+
+Because of this change, we also need a new subclass of `Env`, as each instance
```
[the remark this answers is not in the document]
mihaimaruseac: Both are equivalent, probably it's easier to just change existing one instead. Will update in doc.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+  well as providing support for compressed data and for buffered I/O;
+* High level API which offers functionality needed by the kernels, ops and other
+  utilities.
+
+Furthermore, the codebase has some additional methods to access the filesystem.
+These are used sporadically in some kernels and ops or for special cases such as
+very special filesystems or test environments. We can translate these additional
+APIs to use the design presented here, unifying API all across the codebase.
+
+![Filesystem layers](20190502-filesystem-plugin-modular-tensorflow/big_picture.png "Overview of filesystem support")
```
ewilderj: Image path wrong, the date in the folder has changed.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+
+If a filesystem doesn’t support one of the file types, then we don’t need to
+implement the corresponding interface. Instead, the method creating the file of
+that type can just error, as in this
+[HadoopFileSystem][HadoopFileSystem] example.
+
+There are multiple filesystems supported by TensorFlow, some of which are
+present in the next diagram:
+
+![Example filesystems](20190502-filesystem-plugin-modular-tensorflow/filesystem.png "Example filesystems")
```
ewilderj: https://github.com/tensorflow/community/pull/101

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+`FileSystemCopyFile()` helper introduced in the previous section.
+
+The `GetMatchingPaths()` call can be used to identify all paths in the current
+filesystem that match a specific pattern. The default implementation uses
+[`GetMatchingPaths()`][GetMatchingPaths]:
+
+The following diagram shows several of the implementations of the `Env`
+interface:
+
+![Example environments](20190502-filesystem-plugin-modular-tensorflow/env.png "Example environments")
```
ewilderj: Image path wrong.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
ewilderj: sorry, copypasta. I meant to comment that the image path is wrong.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
mihaimaruseac: Fixed now, thanks

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
mihaimaruseac: Fixed now, thanks

20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+on ensuring a better development workflow by a team-wide effort designed to make
+TensorFlow more modular according to the [current RFC][modular_rfc].
+
+A major module of the modular TensorFlow involves all operations regarding files
+and the filesystem. Saving models, checkpointing, reading files as input to
+models (i.e. images, audio, etc.), and many other file manipulations are all
+tasks that TensorFlow has to support.
+
+While users can use the underlying operating system’s API (such as calling fopen
+and friends on Posix environments, for example), it is definitely better to have
```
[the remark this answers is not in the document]
mihaimaruseac: GitHub made a duplicate comment, resolving the duplicates but will handle the originals

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+
+Instead, we provide 3 different data structures to act as function tables, to
+hold function pointers for files and the filesystem. Currently, these structures
+are filled with the functionality already present in the existing filesystem
+API:
+
+```cpp
+// Operations on a TF_RandomAccessFile
+typedef struct TF_RandomAccessFileOps {
+  size_t (*const Read)(const TF_RandomAccessFile*, uint64, size_t, char*, TF_Status*);
```
yongtang: @mihaimaruseac One very useful method for random access file could be `Size` or `GetSize` which returns the size of the file. At the moment, File Size is not visible from RandomAccessFile which is really inconvenient in many situations.

The reason being that many file format actually has the meta data at the end of the file (not at the beginning of the file). One good (and important) example is the zip file. Zip file's meta data is stored at the end of the file (central directory). So any parser to open a zip file will have to find the file size first before processing.

With a `GetSize` (or `Size`) API, RandomAccessFile could be greatly optimized.

Note technically it is possible to find the file size with `Read` to probe through until the EOF is reached, but this is really not efficient.
 
Of course, some file systems may not be able to provide a `Size` before hand (e.g., in stream fashion). I think by default `GetSize` could simply return `NotImpemented`. So any file size based optimization will only be "best effort".

For optimization, if a `GetSize`/`Size` API is exposed, then reading files could try `GetSize` first. If `NotImplemented` is returned, then a normal approach of probe through until the end is used. Otherwise the size of the file could be directly used.

Note: ZIP is the format mentioned but a lot of the files are actually zip files with `[Content_Types].xml` or `META-INFO` file inside.

For example, Windows office files (pptx, docx, etc) are all zip files. Java jar files are zip files. And python whl files are zip files as well.

You could just use unzip to decompress pptx, jar, whl actually and see each individual content files.

20190506-filesystem-plugin-modular-tensorflow.md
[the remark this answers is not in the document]
mihaimaruseac: Done.

20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+1. Interface between TensorFlow and plugins must be as simple as possible as
+   well as thoroughly tested. It is better to localize errors in either
+   TensorFlow core or the plugin providing the filesystem functionality than in
+   the glue code between the two.
+
+1. Plugin registration should support mandatory metadata such as version
+   numbers, checks for backwards and forwards compatibility as well as optional
+   fields such as author information, bug reporting URL, etc.
+
+1. The mandatory metadata must guarantee _binary compatibility_: allow plugins
```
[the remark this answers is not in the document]
mihaimaruseac: Done.

20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+The low-level filesystem API in TensorFlow guarantees the same interface both
+across multiple operating systems as well as across multiple filesystems hiding
+the complexities arising from the differences between these platforms. Thus,
+there are two different things that need consideration: implementing a new
+filesystem and making it accessible for each operating system (i.e.
+registering).
+
+#### The filesystem implementation
+
+A new filesystem is currently implemented by subclassing 4 interfaces: one for
```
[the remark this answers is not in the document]
mihaimaruseac: Done.

20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+  virtual Status GetMatchingPaths(const string& pattern, std::vector<string>* results) = 0;
+
+  // Misc
+  virtual void FlushCaches();
+  virtual string TranslateName(const string& name) const;
+};
+```
+
+In TensorFlow, files are identified by a URI of form
+`[<scheme>://[<host>]]<filename>`. Because sometimes we only need the filename
```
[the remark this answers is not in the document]
mihaimaruseac: Done.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+  virtual Status GetMatchingPaths(const string& pattern, std::vector<string>* results) = 0;
+
+  // Misc
+  virtual void FlushCaches();
+  virtual string TranslateName(const string& name) const;
+};
+```
+
+In TensorFlow, files are identified by a URI of form
+`[<scheme>://[<host>]]<filename>`. Because sometimes we only need the filename
```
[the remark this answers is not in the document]
mihaimaruseac: Done.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+The low-level filesystem API in TensorFlow guarantees the same interface both
+across multiple operating systems as well as across multiple filesystems hiding
+the complexities arising from the differences between these platforms. Thus,
+there are two different things that need consideration: implementing a new
+filesystem and making it accessible for each operating system (i.e.
+registering).
+
+#### The filesystem implementation
+
+A new filesystem is currently implemented by subclassing 4 interfaces: one for
```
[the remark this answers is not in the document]
mihaimaruseac: Done.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
mihaimaruseac: Done.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
mihaimaruseac: Done.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+
+Instead, we provide 3 different data structures to act as function tables, to
+hold function pointers for files and the filesystem. Currently, these structures
+are filled with the functionality already present in the existing filesystem
+API:
+
+```cpp
+// Operations on a TF_RandomAccessFile
+typedef struct TF_RandomAccessFileOps {
+  size_t (*const Read)(const TF_RandomAccessFile*, uint64, size_t, char*, TF_Status*);
```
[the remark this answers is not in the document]
mihaimaruseac: I don't think so, the filesystem state is internal to the filesystem, which will be retrieved by the filesystem implementation from the file structure itself. I'll clarify this too

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+TF_CAPI_EXPORT extern void TF_InitPlugin(TF_Status* status);
+```
+
+This function must allocate all of the above tables and then call a function
+provided by the core TensorFlow to register plugin filesystems:
+
+```cpp
+TF_CAPI_EXPORT extern void RegisterFileSystemPlugin(
+    const string& plugin, const string& scheme,
+    const TF_Filesystem_Plugin_Metadata* metadata,
```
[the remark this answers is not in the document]
mihaimaruseac: I'll need to test first that this won't cause compatibility issues. I also think that passing them by value results in extra copies, but we need at least one copy anyway so probably this doesn't matter

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+```cpp
+class FileSystemRegistry {
+  typedef std::function<FileSystem*()> Factory;
+  virtual Status Register(const string& scheme, Factory factory) = 0;
+  virtual FileSystem* Lookup(const string& scheme) = 0;
+  virtual Status GetRegisteredFileSystemSchemes(std::vector<string>* schemes) = 0;
+};
+```
+
+### Convenience API
```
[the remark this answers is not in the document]
mihaimaruseac: I'm separating this into an appendix

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+
+We will allow each structure to evolve independently. That is, the `version`
+field in each one can be increased independently of the value of the field in
+the other structures.
+
+To increase the range of compatibility, we will give a guarantee that, except on
+new major versions, new methods will only be added to the end of each of these
+structures and that no method will be deleted from the table. Changing the
+signature of a method will be handled by adding the method with the new
+signature to the end of the table and marking the old method as deprecated.
```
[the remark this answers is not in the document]
mihaimaruseac: Added that to the doc

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+
+Deprecation of a method will be done in two parts. First, we will add compiler
+pragmas to warn the users compiling against the interface and relying on
+deprecated methods. Second, at plugin’s registration, TensorFlow code can check
+the function pointers sent by the plugin to see which ones are for deprecated
```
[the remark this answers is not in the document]
mihaimaruseac: I added a section of questions and comments and raised this point here with a short analysis on benefits/costs for both cases. I'm leaning towards private structs, as mentioned in the section

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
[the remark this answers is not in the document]
mihaimaruseac: Added this to the open questions too

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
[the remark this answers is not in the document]
mihaimaruseac: I added a section of questions and comments and raised this point here with a short analysis on benefits/costs for both cases

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,1082 @@
...
+which is something we don’t want.
+
+Instead, we provide 3 different data structures to act as function tables, to
+hold function pointers for files and the filesystem. Currently, these structures
+are filled with the functionality already present in the existing filesystem
+API:
+
+```cpp
+// Operations on a TF_RandomAccessFile
+typedef struct TF_RandomAccessFileOps {
```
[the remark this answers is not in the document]
mihaimaruseac: I changed the wording so first time I'm presenting the structs the version info is elided and only presented in the proper section

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
mihaimaruseac: I agree. Added this to the questions too but I totally agree now.

mihaimaruseac:
Thank you @alextp for all the review. I incorporated all of your changes, either by inline editing, extracting to appendix or deferring to an open questions section (including the way I'm leaning to) at the end of the document.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
mihaimaruseac: I agree with this request. I only included the API calls that are currently included in the existing classes but I guess we can expand these. I'm going to add a new section on proposed additions to the op tables.

mihaimaruseac: Actually one of the fields of `Read` is the offset from where you want to read and the other is the number of bytes to read. So, if reading a zip file, you can go directly to the end of the file and read the size section from there.

Sure, we can add it as convenience function but we only identify the filesystem plugin based on the scheme part of the URI, not the extension (`hdfs://path/to/file.zip` and `/path/to/file.zip` will use different implementations but `/path/to/file.zip` and /path/to/another/file.txt` will use the same ops)

yongtang: @mihaimaruseac while it is possible to read the content with offset, the `Read` API does not give you the option to specify `offset = -10` to read from 10 bytes before the end of the file EOF. In oder to achieve that, we will need to find out the length of the file (e.g., say 100), and subtract by 10 (100 - 10 = 90) to read from offset `90` (vs. `-10`).

The filesystem plugin will be consumed by a client calling it (a function inside tensorflow, or another plugin/customer ops outside of tensorflow repo).

For example, inside tensorflow one function might want to parse a file (not necessary zip file) with file name `hdfs://path/to/special/file`. Outside of tensorflow another plugin want to access Tensorflow's file system schema URI `gcs://path/to/another/file`.

Eventually the  function (inside or outside tensorflow) will need to access `Read` (or proposed `GetSize/GetFileSize`) through the plugin.

If `GetSize/GetFileSize` is not exposed, then size is unknown to this function (function inside tensorflow, or another plugin using tensorflow). This function will have to probe through to find the EOF.

mihaimaruseac: I wasn't clear enough, sorry. Let me try with an example.

Suppose we live in a world where there are only 3 types of files: `*.start` which have the size metadata at the beginning, `*.end` which have it at the end and `*.none` which have no size metadata.

Suppose we also have only two filesystems: A (`a:///path/to/file`) and B (`b:///path/to/file`). Let's assume that A allows rewinds (so you can read the file and then come back to its start) while B is only streaming (once you read some part of the file, you can't go back, unless you open the file again).

Suppose we have a piece of user code which takes as input a filepath and returns the size of the file, let's call this `ExternalFileSize(const string& path)`.

There are 4 posibilities:

1. We keep the API as it is. In this case, `ExternalFileSize(path)` will call `Env::Default()->NewRandomAccessFile(...)` to get a new `RandomAccessFile` object and will have to call `Read` on it until the end increasing a size variable and then returning it. This, I agree with you, is not that good.

    ```cpp
    // This code is outside TensorFlow
    size_t ExternalFileSize(const string& path) {
      RandomAccessFile file;
      Env::Default()->NewRandomAccessFile(path, &file);
      return ReadFromStartToEndAndGetSize(file);
   }

   // Extracting this as separate function as I refer to it below
   // This is still outside TensorFlow
   ReadFromStartToEndAndGetSize(RandomAccessFile file) {
      size_t size = 0;
      // ignore casts for size_t to uint64 and from Status to bool
      while (file->Read(size, amount_to_read, output, result)) {
        // assume no partial reads for now as it doesn't matter here
        size += amount_to_read;
      }
      return size;
    }
    ```

2. We add a `GetSize` method in the plugin API and in `RandomAccessFile` class. To implement `ExternalFileSize(path)` we now call `NewRandomAccessFile` and `GetSize` on the returned object. However, `NewRandomAccessFile` dispatches on the scheme, so both plugins need to handle `*.start`, `*.none` and `*.end` files inside the `GetSize` equivalent, resulting in code duplication as well as behavior that is not common across plugins. This is what I was mentioning above, in the second paragraph.

    ```cpp
    // This code is outside TensorFlow
    size_t ExternalFileSize(const string& path) {
      RandomAccessFile file;
      Env::Default()->NewRandomAccessFile(path, &file);
      return file->GetSize();
   }

   // This code is in plugin (both A and B)
   size_t GetFileSize(const TF_RandomAccessFile *file) {
      const char *filename = file->filename;

      // SizeAtStart, SizeAtEnd, ReadFirstBlock, ReadLastBlock, ReadSizeFromMetadata, ReadFromStartToEndAndGetSize
      // are all functions implemented in each plugin, code duplication
      if (SizeAtStart(filename)) {
        char* metadata = ReadFirstBlock(file);
        return ReadSizeFromMetadata(metadata);
      } else if (SizeAtEnd(filename)) {
        char* metadata = ReadLastBlock(file);
        return ReadSizeFromMetadata(metadata);
      } else {
        // This calls the plugin function, not the external one!
        return ReadFromStartToEndAndGetSize(file);
      }
    }
    ```

3. We allow `Read` to receive a negative argument for the offset. In this case, `ExternalFileSize(path)` first looks at the path extension. If it is `none` it calls `Read` and increments a size variable until end of file, returning this size. If it is `end`, it calls `Read` with the proper negative offset, reads until EOF and returns the size from the metadata. If it is `start`, it reads from start, only the metadata and returns the size field. Since `Read` now is on core tensorflow, each plugin only needs to implement it with support for negative offsets. If negative offsets are not supported (i.e., streaming filesystems) then `Read` with negative offset fails and code falls back to the same case as `none` extension above.

    ```cpp
    // This code is outside TensorFlow
    size_t ExternalFileSize(const string& path) {
      RandomAccessFile file;
      Env::Default()->NewRandomAccessFile(path, &file);

      // SizeAtStart, SizeAtEnd, ReadSizeFromMetadata, ReadFromStartToEndAndGetSize
      // implemented only once in the external code
      if (SizeAtStart(filename)) {
        file->Read(0, amount_to_read, output, metadata);
        return ReadSizeFromMetadata(metadata);
      } else if (SizeAtEnd(filename)) {
        if (file->Read(-amount_to_read, amount_to_read, output, metadata)) {
          return ReadSizeFromMetadata(metadata);
        } else {
          // filesystem doesn't support negative offsets (filesystem B)
          // so, fallback to default
          return ReadFromStartToEndAndGetSize(file);
        }
      } else {
        return ReadFromStartToEndAndGetSize(file);
      }
      return file->GetSize();
   }

   // This code is in plugin A
   size_t Read(const TF_RandomAccessFile *file, uint64 offset, ...) {
      if (offset < 0) { 
        // position from end and read ...
      } else {
        // read directly
        ...
      }
   }

   // This code is in plugin B
   size_t Read(const TF_RandomAccessFile *file, uint64 offset, ...) {
      if (offset < 0) { 
        // error as seek is not supported
      } else {
        // read directly
        ...
      }
    }
    ```

4. We change `Read` to only receive the number of bytes to read and add a `Seek` method to specify where the read should happen from. `Seek` will have an argument for the relative position (negative offset, positive offset) and another enum argument to specify if the position change should be from current position, start of file or end of file. 

    ```cpp
    // This code is outside TensorFlow
    size_t ExternalFileSize(const string& path) {
      RandomAccessFile file;
      Env::Default()->NewRandomAccessFile(path, &file);

      // SizeAtStart, SizeAtEnd, ReadSizeFromMetadata, ReadFromStartToEndAndGetSize
      // implemented only once in the external code
      if (SizeAtStart(filename)) {
        file->Read(amount_to_read, output, metadata);
        return ReadSizeFromMetadata(metadata);
      } else if (SizeAtEnd(filename) && file->Seek(-amount_to_read, FILE_END)) {
        file->Read(amount_to_read, output, metadata);
        return ReadSizeFromMetadata(metadata);
      } else {
        return ReadFromStartToEndAndGetSize(file);
      }
      return file->GetSize();
   }
    ```
   This is similar to 3. above, but has a clearer API.

I would go with 4 or 3 instead of 2 as you propose.

yongtang: @mihaimaruseac Thanks. I think 3 or 4 makes sense 👍 . For 3, maybe the offset will be `int64` instead of `uint64` (to allow negative numbers)?

mihaimaruseac: It has to be `int64`, indeed

suphoff: Option (**4**) would prevent concurrent reads from the RandomAccessFile (and require additional locking ...) - while this functionality is currently underused I expect this to change.

I would really like to see a GetSize() (**2**) method for RandomAccessFile -  currently GetFileSize() is used as a substituted either before or after NewRandomAccessFile().
GetSize() could ensure on most file systems that both operations actually refer to the same file object in the presence of rename/delete file operations.

mihaimaruseac: So that leaves option 3 to read the size after decompression.

Indeed, `GetFileSize()` returns the size of the file on disk. Probably we can also implement `GetSize()` which does the same thing, returning the size of the file at the moment it was read (and filesystems can decide to not implement it in which case we'll default to calling `GetFileSize()` from the filesystem object itself)

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,829 @@
...
+the filesystem plugin interface. Then, the outer level C API can just include
+this header and the structure will still be visible there.
+
+Once these changes are in place, the rest of the TensorFlow code can use the
+modular filesystem API transparently, with no change.
+
+### Optional additional changes to TensorFlow C/C++ core
+
+Something we might elect to do is to force all future filesystem implementations
+to use the plugin design suggested above. To do that, we will have to convert
```
alextp: +1 let's definitely aim to have only one way of specifying filesystems in the future

mihaimaruseac:
Review notes:

1. CI tests to ensure that version number changes when structure layout changes (this also keeps us safe from compile flag changes which could cause issues)
1. During implementation we need to decide between the following methods to deal with the function tables for the operations:

    1. Structures should be private, size field and `memcpy`-ed upon registration
    1. Structures can be public
    1. Functions in a header, all set to `nullptr`, provide API to set them upon registration

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,829 @@
...
+   pointer? Having a direct pointer to `TF_Filesystem` allows storing the state
+   of the filesystem but we argue that this is not needed for all filesystem
+   implementations and in the cases where is needed the plugin can store a
+   pointer to the filesystem inside of the memory pointed to by the
+   `TF_RandomAccessFile` pointer. Leaning towards **No**.
+
+1. **Should we pass plugin structs by value during registration?** Doing so
+   makes ownership clear and prevents memory allocation bugs. Passing them by
+   value results in a copy but we need a copy for security reasons anyway. We're
+   leaning towards **Yes**.
```
r4nt: Is this about the structs that just wrap the void*? In that case, compilers will handle exactly like they would a void* (so no extra copy).

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,829 @@
...
+   core versions. If integration tests for API compatibility between version A
+   of the plugin and version Z of TensorFlow core pass, as detailed on [modular
+   TensorFlow proposal][modular_rfc], then version A of the plugin should be
+   allowed to work with version Z of TensorFlow core and mandatory metadata
+   should allow this. If integration tests between version B of plugin and
+   version Y of core fail then the metadata must forbid this combination.
+
+1. Furthermore, it is ideal to have compilation failures when trying to compile
+   plugins against an old API interface which will surely fail the version
+   metadata tests. For example, if a method's signature changes this will break
```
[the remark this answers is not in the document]
mihaimaruseac: Yes, that's exactly the case.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,829 @@
...
+
+  void (*const FlushCaches)(const TF_Filesystem*);
+  const char* (*const TranslateName)(const TF_Filesystem*, const char*);
+
+  void (*const Init)(TF_Filesystem*);
+  void (*const Cleanup)(TF_Filesystem*);
+} TF_FilesystemOps;
+```
+
+We also provide some convenience calls that plugins can use to fill in these
```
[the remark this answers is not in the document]
mihaimaruseac: "we" here is more like "we, the authors of the plugin interface".

I'll have to reconsider the Windows issue, thanks for pointing it out

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
mihaimaruseac: This is about the structs containing the function pointers.

mihaimaruseac:
Thanks @r4nt for the extra review and suggestions. Will consider these too while implementing.

rfcs/20190506-filesystem-plugin-modular-tensorflow.md
```diff
@@ -0,0 +1,829 @@
...
+
+ private:
+  string filename_;
+  const TF_RandomAccessFile* file_;
+  const TF_RandomAccessFileOps* ops_;
+  TF_DISALLOW_COPY_AND_ASSIGN(ModularRandomAccessFile);
+};
+
+Status ModularRandomAccessFile::Read(uint64 offset, size_t n, StringPiece* result, char* scratch) const {
+  TF_Status* plugin_status = TF_NewStatus();
```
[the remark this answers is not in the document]
mihaimaruseac: I guess it can also go on the stack. Will have to see how it is used elsewhere and also will have to learn modern C++ :)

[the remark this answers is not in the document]
mihaimaruseac: Actually, it cannot go on the stack as it needs to be private to plugins for ABI compatibility reasons.

ziliangpeng:
Is there a place to track the implementation status of this RFC?

byronyi:
> Is there a place to track the implementation status of this RFC?

https://github.com/tensorflow/tensorflow/search?q=modular+filesystem&type=Commits

mihaimaruseac:
For now all the implementation will reside in [`tensorflow/c/experimental/filesystem`](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/c/experimental/filesystem).

POSIX support should be completed. I'm working on windows now then I will update this and send an email with instructions on how to test the experimental filesystems instead of the non-modular ones. At that time, we will start working on converting all other filesystems and then we will switch TF to use the modular ones (by moving the code to the proper locations)

PR #104, opened 2019-05-14
Update addons maintainership RFC

seanpmorgan:
PR to update the maintainers of addons. I don't think we need to keep this constantly up to date but the prior listing was pretty incomplete and outdated.


ewilderj:
Thanks Sean. If this doc does get out of date, is there a place we have the current list, and you could update this RFC to point to that?

PR #112, opened 2019-06-10
Document concurrency semantics for resource variables

sanjoy:
No implementation change is needed, this is the behavior TensorFlow currently implements.


sanjoy:
@alextp PTAL.

ewilderj:
To clarify, are we publishing this purely informationally?

sanjoy:
> To clarify, are we publishing this purely informationally?

Not sure what you mean by "informationally", but this RFC does not propose any implementation changes, but only documents what TensorFlow does today already.

ewilderj:
> Not sure what you mean by "informationally", but this RFC does not propose any implementation changes, but only documents what TensorFlow does today already.

By informationally, I mean please confirm we are publishing it for documentation reasons, rather than inviting any comment and review.

sanjoy:
> By informationally, I mean please confirm we are publishing it for documentation reasons, rather than inviting any comment and review.

Yes, this is only documenting behavior already part of TensorFlow.  If there are comments or suggestions (beyond typographical errors) they should be addressed as separate PRs that fix the implementation and the documentation.

ewilderj:
Thanks @sanjoy! I'll wait for an LGTM from @alextp and merge.

PR #115, opened 2019-06-13
RFC: MLIR Dialects for TensorFlow

joker-eph:
**This proposal is open for review until 2019-06-27**

# TensorFlow MLIR Dialects

 |Status        | Proposed                                  |
|:------------ | :-----------------------------------------|
|**Author(s)** | Mehdi Amini (aminim@google.com)           |
|              | Tatiana Schpeisman (shpeisman@google.com) |
|              | Chris Lattner (clattner@google.com)       |
|**Sponsor**   | Alexandre Passos (apassos@google.com)     |
|              | Jacques Pienaar (jpienaar@google.com)     |
|**Updated**   | 2019-06-10                                |

 ## Objective

 [MLIR](https://medium.com/tensorflow/mlir-a-new-intermediate-representation-and-compiler-framework-beba999ed18d) is the intermediate representation and compiler framework we are investing in to build the compiler infrastructure for TensorFlow. The representation for TensorFlow exposed in this document will be what future high-level transformations will operate on.

We make use of two different dialects to model TensorFlow graphs in MLIR: first the tf_executor dialect that represents the execution model of the TensorFlow executor (e.g. control dependencies, deadness propagation) and the tf dialect which represent the regular operations in a TensorFlow graph (the ones that don’t have special contract with the executor).

One intent of this design is that TensorFlow 2.x features can choose to target just the tf dialect, allowing us to phase out the tf_executor dialect in subsequent TensorFlow releases. The combination of the two dialects allows to represent arbitrary existing TensorFlow graphs.

The representation in this document does not address the specific needs of accelerators or "custom backends" for TensorFlow. We plan to provide a generic infrastructure for replacing the TF/XLA bridge with a more flexible and reusable system across targets. A later design proposal will address these aspects. Also this representation does not address shape inference, an independent design exploration is being conducted separately at the moment.

### DIFF
--- a/rfcs/20190612-mlir-dialect.md
+++ b/rfcs/20190612-mlir-dialect.md
@@ -1,0 +1,335 @@
+# TensorFlow MLIR Dialects
+
+|Status        | Proposed                                  |
+|:------------ | :-----------------------------------------|
+|**Author(s)** | Mehdi Amini (aminim@google.com)           |
+|              | Tatiana Schpeisman (shpeisman@google.com) |
+|              | Chris Lattner (clattner@google.com)       |
+|**Sponsor**   | Alexandre Passos (apassos@google.com)     |
+|              | Jacques Pienaar (jpienaar@google.com)     |
+|**Updated**   | 2019-06-10                                |
+
+## Objective
+
+[MLIR](https://medium.com/tensorflow/mlir-a-new-intermediate-representation-and-compiler-framework-beba999ed18d)
+is the intermediate representation and compiler framework we are investing in to
+build the compiler infrastructure for TensorFlow. The representation for
+TensorFlow exposed in this document will be what future high-level
+transformations will operate on.
+
+We make use of two different dialects to model TensorFlow graphs in MLIR: first
+the `tf_executor` dialect that represents the execution model of the TensorFlow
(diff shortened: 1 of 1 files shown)


ewilderj:
@joker-eph thanks for the RFC. I note it doesn't have the end of a review period annotated on it. Usually this would be two weeks, ie. 6/27. Are you good with that? If so, I'll add in.

jpienaar:
@ewilderj That sounds fine yes.

byronyi:
I see SendRecvOps mentioned with Eager, but not later-added CollectiveOps. Ping @dubey and @mrry; is this somewhat related to what you've mentioned in the meeting last week?

mrry:
@byronyi The collective ops don't (as far as I know) require any special treatment by the executor, so I think they're handled by the TensorFlow dialect.

ewilderj:
@joker-eph was this RFC accepted by review? If so can you copy any notes from the review meeting here? Then I'll be able to merge it.

joker-eph:
@ewilderj it was accepted! (and implemented)
If I remember correctly, @jpienaar took some notes at the time of some clarifications that were asked to be added in the doc.

theadactyl:
@ewilderj can you add your review so this can be merged?

PR #124, opened 2019-07-23
RFC: On-Device Training with TensorFlow Lite

miaout17:
**Review period open until 2019-08-06**

# On-Device Training with TensorFlow Lite

Status        | Proposed
:------------ | :---------------------------------------------------------------
**Author(s)** | Yu-Cheng Ling (ycling@google.com)
**Sponsor**   | Andrew Selle (aselle@google.com), Jared Duke (jdduke@google.com)
**Updated**   | 2019-07-22

## Overview & Roadmap

TensorFlow Lite is TensorFlow's recommended solution for on-device machine learning. Initially the project focuses on inference, but more and more users are asking for on-device training recently.

The doc scopes a multi-quarter effort to get generalized & optimized on-device training working with TensorFlow Lite


rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+  gradients = tape.gradient(loss, dense_layers.trainable_variables)
+  _OPTIMIZER.apply_gradients(zip(gradients, dense_layers.trainable_variables))
+```
+
+Note that `tf.GradientTape` is calculated over only the dense layers, and
+`tape.gradients` are only applied to `dense_layers.trainable_variables` too.
+
+Though this example is simple, it's easy to extend it to support complex use
+cases. For example:
+
```
karmel: Given that we're focusing on Keras models, it would be nice if there were a way to simply pass the model and use the built-in methods by default, rather than redefining loops. In other words, is it worth having a special happy path for Keras models?

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+**Milestone 1: Working prototype for basic training (e.g. fully connected / conv layers only)**<br/>
+Goal: Have a working prototype to train over fully connected & convolutional layers.
+
+**Milestone 2A: Optimized basic training**<br/>
+Goal: Make inference & training performance comparable with TFMobile.
+
+**Milestone 2B: Enable on-device training loop with TensorFlow functionality**<br/>
+Goal: Encode the on-device training loop inside a TensorFlow model, with TensorFlow functionality like tf.Example and tf.DataSet.
+
```
karmel: What is in scope here? Eg... Preprocessing? Control flow? There is a whole support matrix here to consider-- how much is covered?

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+For the variable requirement of TensorFlow Lite training, the high level goals
+are:
+
+*   Convert TensorFlow Resource Variables to TensorFlow Lite, and preserve the
+    same semantic
+*   Able to share variables between multiple TensorFlow Lite subgraphs
+
+We propose to define variables in TFLite that have similar semantics to resource
+variables in TensorFlow:
```
karmel: Why do we need a special variable type? Can you explain further what the difference is?

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+*   Option 2: Write the variable value back to TensorFlow Lite model.
+    *   Pros: Only having a single file & single file format.
+    *   Cons: Have to ship FlatBuffer writer in the binary. Have to write the
+        entire model (which may contain other huge frozen constants) back to
+        storage.
+*   Option 3: Use TensorFlow checkpoint format
+    *   3A: Implement checkpoint reading / writing functionality in TensorFlow
+        Lite
+    *   3B: Reading / writing through Flex runtime (so it only works when Flex
+        runtime is used)
```
karmel: +1 to any option that doesn't involve inventing a new serialization scheme.

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+
+### Feeding data with tf.Example / tf.DataSet
+
+`tf.Example` and `tf.DataSet` are technically not training-specific
```
karmel: nit: tf.data.Dataset is I presume what you mean?

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+requirements. These features can also be used in inference.
+
+The basic training can work by feeding the raw data (e.g. float values) into
+TensorFlow Lite interpreter. On top of this, we can support tf.Example and
+tf.DataSet to make this easier to use (likely via Flex runtime). There may be a
```
karmel: Given the multiple references, it would be useful to define/give background on the flex runtime somewhere in this doc.

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+performance drawback using these features.
+
+### Selective Registration
+
+Initially we will need to rely on Flex runtime to execute some of the fused
+gradient ops. This requires to add the gradient op kernels to Flex runtime, and
+it will bloat up the Flex binary size.
```
karmel: How much bloat are we talking? What percent increase is expected over the existing binary size?

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+## Overview & Roadmap
+
+TensorFlow Lite is TensorFlow's recommended solution for on-device machine
+learning. Initially the project focuses on inference, but more and more users
+are asking for on-device training recently.
+
+The doc scopes a multi-quarter effort to get generalized & optimized on-device
+training working with TensorFlow Lite. The project can be broken down into a few
+milestones:
+
```
karmel: Are there any new APIs being proposed in this doc? I don't see any new concrete APIs so much as ways to extend existing ones to training-- is that an accurate reading?

rfcs/20190722-tflite-training.md
miaout17: Thanks for the comment. 

Special happy Keras path sounds great. I'd love to continue this conversation. I don't fully understand "use the built-in methods by default". Do you have concrete recommendation for the design?

On the other hand, we're not only focusing on Keras model. The example is written with Keras API because it's more concise, but we're also aiming to support models constructed with lower-level TF APIs.

rfcs/20190722-tflite-training.md
miaout17: Currently TensorFlow Lite doesn't really(*) support variables. Essentially there's no way to update the trainable weights. That's why we need a new design. 

(*) In TensorFlow Lite codebase, there's a concept called "variable", but it's only works with fused RNN ops now and can't be shared across subgraphs. It doesn't fulfill the requirement for training. In the long term we want to deprecate the old variable and migrate to the new resource-variable like semantic.

rfcs/20190722-tflite-training.md
miaout17: Thanks. I'm starting looking into just using TF checkpoint. Hopefully it also helps the interoperability between TensorFlow and TensorFlow Lite (there's a chance that we can take TensorFlow Lite trained checkpoint and load into TensorFlow directly, and vice versa). 

I'm hoping I can update this section by 8/5.

rfcs/20190722-tflite-training.md
miaout17: Sure. Leaving this comment open and will do it soon.

rfcs/20190722-tflite-training.md
miaout17: As far as I remember, TFLite interpreter (with no kernel) is around 80K, and it's around ~700K with all the builtin ops. The Flex runtime is at least 2MB. This are binary library size for one architecture (e.g. arm64). 

These numbers may be a few months old and TFLite is changing fast. I can find out the newest number if necessary.

rfcs/20190722-tflite-training.md
miaout17: Yes (leaving this open as a TODO)

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+@tf.function(input_signature=[
+    tf.TensorSpec(shape=[None, 64, 64, 3], dtype=tf.float32),
+    tf.TensorSpec(shape=[None, 10], dtype=tf.float32),
+])
+def train(x, y):
+  with tf.GradientTape() as tape:
+    prediction = model(x)
+    loss = _LOSS_FN(prediction, y)
+  gradients = tape.gradient(loss, model.trainable_variables)
+  _OPTIMIZER.apply_gradients(zip(gradients, model.trainable_variables))
```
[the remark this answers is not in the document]
miaout17: The training loop can be in the model or out of the model. Either should work and we want to support both. 

I can foresee that people may want to do both in different cases. When batch training in background, it might be desired to encode the loop into the model to simplify the logic. When training in a foreground app, application developers may want to have a custom loop to get finer grained control.

ewilderj:
Reminder for future: please ping me after submitting this PR so I can publicize to ensure the broader community reads the RFC. I'll announce it right now so the community can read it.

bhack:
What is the main use case of this? Cause I see in mentioned examples is mainly oriented to few layers fine-tuning on device data.
Is there any continual learning scenario on edge devices? What we suppose to do in this case? The model could be often in inference and in training mode on the device in this case or adapt its resources with model evolution when will be available in some low resource incremental techniques for edge devices. 
Or is this just for fine-tuning few layers or enable federated learning?

miaout17:
> Reminder for future: please ping me after submitting this PR so I can publicize to ensure the broader community reads the RFC. I'll announce it right now so the community can read it.

Acknowledged.

bhack:
Wth the big effort of the Tensorflow team to unify over keras API  will this new training capabilities be exposed as not Keras API? /cc @fchollet
As currently you can interact with the interpreter in python, c++ and Java is it not a problem that Keras is currently a Python only API and so it is hard to unify over that in other context like this one?

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+*   The user want to retrain the 2 dense layers using personalized data on
+    device
+
+### Authoring TensorFlow graph
+
+**Defining the model using Keras Layers**
+
+The code defines the entire model using Keras API. Since we want to be retrain
+only dense layers on device, the model is defined in 2 parts (`conv_layers` and
+`dens_layer`).
```
lc0:
```suggestion
`dense_layers`).
```

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+     for func in [inference, train_dense_layers]])
+```
+
+In this case, it's only required to run inference or train dense layers, so only
+`inference` and `train_dense_layers` functions are exported. If it's required to
+train the whole model on device, add the `train` function when converting.
+
+### Using the Model in TensorFlow Lite runtime
+
+There's an ongoing design effort to add enable users to call different subgraphs
```
lc0:
```suggestion
There's an ongoing design effort to enable users to call different subgraphs
```

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+The `train` function can be used to train the model offline. For retraining only
+dense layers, define another `train_dense_layers` TensorFlow function:
+
+```python
+
+@tf.function(input_signature=[
+    tf.TensorSpec(shape=[None, 64, 64, 3], dtype=tf.float32),
+    tf.TensorSpec(shape=[None, 10], dtype=tf.float32),
+])
+def train_dense_layers(x, y):
```
lc0: would it make sense to add more verbosity and return loss / metrics? How would you see the process of actual training blindly on device?

lc0: I mean, do we want to share these later via `Interpreter`?

miaout17:
> What is the main use case of this? Cause I see in mentioned examples is mainly oriented to few layers fine-tuning on device data.
> Is there any continual learning scenario on edge devices? What we suppose to do in this case? The model could be often in inference and in training mode on the device in this case or adapt its resources with model evolution when will be available in some low resource incremental techniques for edge devices.
> Or is this just for fine-tuning few layers or enable federated learning?

Yes this should enable federated learning, fine-tuning / transfer learning etc. For small models it's also possible to train from scratch on device. 

The RFC is mainly an infrastructure change to make this possible with TensorFlow Lite, and application developers can use their creativity to use this feature.

miaout17:
> Wth the big effort of the Tensorflow team to unify over keras API will this new training capabilities be exposed as not Keras API? /cc @fchollet
> As currently you can interact with the interpreter in python, c++ and Java is it not a problem that Keras is currently a Python only API and so it is hard to unify over that in other context like this one?

Developers can use Keras API to define a model in TensorFlow, and convert it to TensorFlow Lite format to use it efficiently on device. This is how TensorFlow and TensorFlow Lite works. We don't really have Keras API or similar things in TensorFlow Lite, but we do get benefits of Keras for model construction. 

Let me know if I understand your question correctly. Thanks!

rfcs/20190722-tflite-training.md
miaout17: Yes. This example is overly simplified to be concise. In practice I think people will want to return loss in training function, and define an additional validation function.

lc0:
@googlebot I fixed it.

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,427 @@
...
+converter will convert each TensorFlow function to a TensorFlow Lite subgraph,
+and users can choose subgraphs to invoke.
+
+The following code defines `inference` and `train` TensorFlow functions.
+
+```python
+@tf.function(input_signature=[
+    tf.TensorSpec(shape=[None, 64, 64, 3], dtype=tf.float32)])
+def inference(x):
+  return model(x)
```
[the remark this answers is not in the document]
miaout17: In TF 2.0, the TensorFlow Lite converter takes a `ConcreteFunction` as an input, so wrapping to-be-converted things inside `tf.function` becomes a canonical way. See also [this doc](https://www.tensorflow.org/lite/r2/convert/concrete_function). 

LMK if I understand your question correctly. 

Related your other comment, indeed I think it will be better if we can provide an easier to use interface for Keras.

rfcs/20190722-tflite-training.md
miaout17: Updated this section. Making TF checkpoint recommended solution for now.

rfcs/20190722-tflite-training.md
miaout17: Fixed

rfcs/20190722-tflite-training.md
miaout17: Done

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,444 @@
...
+tf.data.Dataset to make this easier to use (likely via Flex runtime). There may be a
+performance drawback using these features.
+
+### Selective Registration
+
+Initially we will need to rely on Flex runtime to execute some of the fused
+gradient ops. This requires to add the gradient op kernels to Flex runtime, and
+it will bloat up the Flex binary size.
+
+We also plan to implement fused training ops as TensorFlow Lite builtin ops.
```
jackwish: Will there be *building configuration*, which controls whether to build the training-dedicated operators (or functionality) into binary? To me, there will be plenty of code to be added, resulting a larger tflite runtime.

rfcs/20190722-tflite-training.md
karmel: Eg, model.fit/test/predict: https://www.tensorflow.org/beta/guide/keras/overview#train_from_numpy_data . I thought @gargn had added direct support to the converter as well, with from_keras_model or similar. An analogous API would be nice here-- given a keras model, assume we want to use .fit/.test/.predict for train/eval/inference.

rfcs/20190722-tflite-training.md
ychen404: Would it be better if user can select what are the operators they need for their model? If users understand what they need they can better trim down the tflite binary size.

ychen404: How about the support of the Pixel Visual core during training? 
It seems that currently, Android Neural Network only has inference APIs.

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,444 @@
...
+*   The whole model is trained offline
+*   The user want to retrain the 2 dense layers using personalized data on
+    device
+
+### Authoring TensorFlow graph
+
+**Defining the model using Keras Layers**
+
+The code defines the entire model using Keras API. Since we want to be retrain
+only dense layers on device, the model is defined in 2 parts (`conv_layers` and
```
ychen404: Can you also train other layers of the model?
It seems that only for the last fully connected layer is a little limited. In federated learning setting, each device may need to train a whole model with its local data.

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,444 @@
...
+### Gradient and optimizer operations
+
+To train a model, it's required to compute **gradients**, and apply the gradient
+to trainable variables with **optimizers** (e.g. SGD, ADAM...etc). Technically,
+gradients and optimizers are just some mathematical computation and variable
+read/write operations. When constructing training graphs in TensorFlow,
+sometimes a single fused op is used (e.g. `ReluGrad`), and sometimes it produces
+multiple regular ops in unfused form (e.g. `Mul`, `MatMul`, `Add`) to compute
+the gradient of one op.
+
```
ychen404: Is there any optimization plan on the storage of intermediate results when training on-device?

rfcs/20190722-tflite-training.md
```diff
@@ -0,0 +1,444 @@
...
+[Select TensorFlow operators to use in TensorFlow Lite](https://www.tensorflow.org/lite/guide/ops_select)
+project, which enables using TensorFlow kernels in TensorFlow Lite directly.
+Throughout the document, it will be referred as "Flex" to be concise.
+
+This is the easiest approach: Just whitelist these training ops to Flex whitelist. However this means Flex runtime is required for training (bloating binary size), and Flex kernels are not optimized for mobile.
+
+**Implement fused ops as TensorFlow Lite builtin ops (optimization)**<br/>
+We can implement these fused ops as TensorFlow Lite builtin ops. This further
+enables optimizing for mobile devices (e.g. writing optimized SIMD CPU kernels,
+writing GPU delegate...etc). This may need a significant amount of work.
```
ychen404: How about the support of the Pixel Visual core during training?
It seems that currently, Android Neural Network only has inference APIs.

rfcs/20190722-tflite-training.md
miaout17: I see what you meant. I'm working closely with @gargn and I'm familiar with the TFLiteConverter code. I can prototype around this idea, but probably need to do that after the review period. 

One extra thought: We often want to train only a few layers (but not the entire model) on-device. I'm unsure if we can use the existing `Layer.trainable` flag to do this (training the entire model offline, turn off `trainable` for most layers, then convert the model.

rfcs/20190722-tflite-training.md
miaout17: > Can you also train other layers of the model?

Yes. 

> Does this mean that tflite can only support dense layer, but not convolution?

No. This is just an example. 

For example, if we change the model conversion code to export 3 functions:
```
tf.lite.TFLiteConverter.from_concrete_functions(
    [func.get_concrete_function()
     for func in [inference, train_dense_layers, train]])
```
then you can use `train` function to train the whole model on the device

rfcs/20190722-tflite-training.md
miaout17: Could you elaborate more? Are you referring to the intermediate computation result (e.g. intermediate tensors those are neither inputs nor outputs) in memory? or the trained weights?

rfcs/20190722-tflite-training.md
miaout17: The software architecture should allow delegating the training computation to hardwares (though the TensorFlow Lite [Delegate API](https://www.tensorflow.org/lite/performance/delegates)). So far we didn't look into specific delegate yet (e.g. NNAPI, GPU) but it should be doable.

rfcs/20190722-tflite-training.md
miaout17: I'm assuming that you're commenting on this line. 

> We also plan to implement fused training ops as TensorFlow Lite builtin ops.

Yes. I think we need better selective registration mechanism on TensorFlow Lite builtin ops too, so the binary size stay small when training isn't used. 

However for the short term, we're just going to use the Flex runtime instead of defining builtin gradient ops.

rfcs/20190722-tflite-training.md
ychen404: Right, i am referring to the intermediate computation results in memory.

rfcs/20190722-tflite-training.md
miaout17: Control flow is definitely in the roadmap. 

Also, be able to leverage the Flex runtime to use TensorFlow kernels for TF Example and DataSet etc. This for users who are requesting to port the TFMobile flow as is (but using TF Example and DataSet isn't necessarily optimized)

bhack:
> Developers can use Keras API to define a model in TensorFlow, and convert it to TensorFlow Lite format to use it efficiently on device. This is how TensorFlow and TensorFlow Lite works. We don't really have Keras API or similar things in TensorFlow Lite, but we do get benefits of Keras for model construction.
> 
Can we not subset some keras API in c so that we can have a minimal Keras subset available in Tflite where the python interpreter could be not available?
What Is the strategical point of view of forking training API between TF and Tflite? It Is better to have a TF.keras C  subset for Tflite so that end users will not need to learn/switch API betwen devices type.

miaout17:
@bhack 

Thanks a lot for the feedback. Your point is well received. 
I think it requires a bigger design (essentially the same point can be applied to inference too), and we can consider to have a higher-level API on top of the TFLite interpreter.

miaout17:
Pasting the summary of the design review notes

* User-friendly route for Keras
  * Takeaway: Don’t operate for Keras at first because that won’t make it flexible to support non-Keras model
  * Target for the layer below Keras (e.g. SavedModel). Theoretically Keras model can be converted to SavedModel so we should have well support for both. 
* Getting gradient and optimizer operations
  * Takeaway: Use TF fused gradient ops initially. Do not introduce fused ops into TFLite. 
* Defining (resource) variables in TensorFlow Lite
  * Takeaway: Define the TFLite resource variable in a way that’s very similar to TF. Using integer as a variable id instead of a top level container name and variable name and type as a key. This makes the conversion very simple and allows the TF ops to 1:1 map to the TFLite ops.
* (Not) Supporting TensorFlow 1.x
  * Takeaway: We will not support TensorFlow 1.x for on-device training in TFLite
  * Some features are supported only with the new MLIR converter (i.e. control flow and training). There are a few things that are required to make this work including needing a way to handle variables. Currently we are only focused on TensorFlow v2 variables and only control flow v2 is supported and only v2 tensor list ops are supported.
* Weight saving format
   * Takeaway: TFLite interpreter should be agnostic to the saving format
   * It should be possible to support multiple saving format
* Having separate training&inference models, or have one model that contains training&inference functions?
  * Takeaway: Ideally we need flexibility to support both options - joined and separate cases
  * Many existing use cases in TF 1.x is the former. The latter is recommended in TF 2.x

miaout17:
@ewilderj Could you help to merge this? Thanks a lot!

bhack:
> @bhack
> 
> Thanks a lot for the feedback. Your point is well received.
> I think it requires a bigger design (essentially the same point can be applied to inference too), and we can consider to have a higher-level API on top of the TFLite interpreter.

@miaout17 Do you mean that there will be a follow-up RFC on this topic?

ewilderj:
> @ewilderj Could you help to merge this? Thanks a lot!

sorry for the delay. Thanks for your work on this RFC! Merged.

PR #126, opened 2019-07-26
RFC: Best practices for custom operations in TensorFlow

alextp:
**Review period is open through 2019-08-09**

# Best practices for custom operations in TensorFlow

| Status        | Proposed                                          |
| **Author(s)** | Alexandre Passos (apassos@google.com |
| **Sponsor** | Karmel Allison (karmel@google.com) |
| **Updated**   | 2019-06-10                                           |

For most of TF’s history, it was very expensive for third-party packages or libraries to release their own tf operations. This created pressure to put ops in tf core or in tf contrib, which created some uncertainty around support stories and backwards compatibility.

Around (but technically not a part of) TF 2.0, however, TensorFlow supports [a straightforward way for third-party packages to build and deploy their own custom TF ops](https://github.com/tensorflow/custom-op/blob/master/README.md). To maintain a healthy ecosystem, we recommend best practices in this RFC.


awav:
That's fantastic! Is there a plan to add XLA op as an example as well?

rfcs/20190726-custom-ops.md
```diff
@@ -0,0 +1,56 @@
...
+ - each downstream package can choose its own backward and forward compatibility
+   processes, allowing fine-grained trade-offs between velocity and stability
+
+## Out-of-tree ops must be namespaced
+
+Since an op’s name uniquely identifies it, different TF packages should ensure
+their op names are globally unique across the entire TF ecosystem. To do so,
+prepend the package’s name to the op’s name and separate with a ‘>’. An op named
+“MatMul” inside the “tensorflow_addons” package should be named “Addons>MatMul”,
+for example.
```
facaiy: @tensorflow/sig-addons-maintainers Seems that we need to rename all C++ ops? cc @seanpmorgan

rfcs/20190726-custom-ops.md
```diff
@@ -0,0 +1,56 @@
...
+This has many advantages:
+ - downstream packages often have a faster release cadence than core TensorFlow
+ - each downstream package can choose its own backward and forward compatibility
+   processes, allowing fine-grained trade-offs between velocity and stability
+
+## Out-of-tree ops must be namespaced
+
+Since an op’s name uniquely identifies it, different TF packages should ensure
+their op names are globally unique across the entire TF ecosystem. To do so,
+prepend the package’s name to the op’s name and separate with a ‘>’. An op named
```
seanpmorgan: Can we make `tf.load_op_library` throw a warning or even an error if there is no package name specified? AFAIK only custom-op repos utilize this function

rfcs/20190726-custom-ops.md
seanpmorgan: https://github.com/tensorflow/addons/issues/379

seanpmorgan:
So I took a stab at implementing this for Addons... but currently it'll throw an "Invalid Op Name" because of the `>`: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/op_def_util.cc#L254

What will the behavior be once the package separating character is accepted? At the moment the camel case op name is lower-cased and separated by `_` in the python object's method. For example `SkipGramGenerateCandidates` becomes `skip_gram_generate_candidates`.

`obj.addons_skip_gram_generate_candidates` seems a little verbose seeing as the obj is already loaded using `tf.load_library` from a known shared_object.

rfcs/20190726-custom-ops.md
```diff
@@ -0,0 +1,56 @@
...
+“MatMul” inside the “tensorflow_addons” package should be named “Addons>MatMul”,
+for example.
+
+The string used for a package’s component name is any valid op name, but should
```
caisq: It would be convenient to developers if you can include a link to the documentation of what constitutes a valid op name (a regex?)

alextp: The regex is in https://github.com/tensorflow/tensorflow/blob/876f07dbc27ce790b076291c74550538ad183c45/tensorflow/core/framework/op_def.proto#L17

caisq: In that case, isn't the suggestion above "To do so, prepend the package’s name to the op’s name and separate with a ‘>’."  not going to work, given that the regex appears to be "[A-Z][a-zA-Z0-9_]*"?

alextp: Clearly this proposal should be read as implying a change to this regex to be something like ([A-Z][a-zA-Z0-9_]>)$EXISTING_REGEX

caisq: OK. Thanks for the clarification.

rfcs/20190726-custom-ops.md
```diff
@@ -0,0 +1,56 @@
...
+it is ok with taking responsibility for the code, it’s fine to propose adding a
+new version with the final intended interface and implementation to core
+TensorFlow. The intermediate states are best explored in another package.
+
+This has many advantages:
+ - downstream packages often have a faster release cadence than core TensorFlow
+ - each downstream package can choose its own backward and forward compatibility
+   processes, allowing fine-grained trade-offs between velocity and stability
+
+## Out-of-tree ops must be namespaced
```
miaout17: The namespace problem is similar to the naming discussion in the [Standardizing Composite Operations In TensorFlow](https://github.com/tensorflow/community/pull/113/files#diff-9076661f98cc4fc83f8f434c42a5c687R309) RFC. 

Any thought around making the naming scheme more unified?

ewilderj:
Ready to merge the RFC but I never heard back after the design meeting. Are there any notes to share? Was this design accepted?

martinwicke:
Yes, this can be merged. Comments have already been integrated as far as I can tell.

PR #127, opened 2019-08-02
RFC: TensorFlow Debugger v2 -- Callbacks for Eager Execution and tf.function

caisq:
**Comment period open through 2019-08-15**

# TensorFlow Debugger v2: Callbacks for Eager Execution and `tf.function`s

| **Author(s)** | Shanqing Cai (cais@google.com)                       |
| **Sponsor**   | Alexandre Passos (apassos@google.com)                |

## Objective

This RFC presents an API-level design for how to achieve debugging instrumentation of [eager execution](https://www.tensorflow.org/guide/eager) and [`tf.function`](https://www.tensorflow.org/beta/guide/autograph) in TensorFlow 2 ("TF2" hereafter). This will enable users to "hook into" the following types of events in TF2 with a unified API:

  - Eager execution of an operation ("op" hereafter) at runtime
  - Creation of an symbolic op at graph-construction time, i.e., when a user-defined Python functions is transformed into a graph (FuncGraph) with the [tf.function](https://www.tensorflow.org/beta/tutorials/eager/tf_function) API.
  - Runtime execution of FuncGraphs.


caisq:
cc @mdanatg

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,349 @@
...
+
+### Design: Debug Callbacks for Op Instrumentation
+
+The following API meets the requirements listed above. The code below shows the
+API and the detailed signature and semantics of the callbacks that can be passed
+to the API.
+
+```python
+# Exported publicly as: tf.debugging.op_callback()
+def op_callabck(callback_fn):
```
lc0:
```suggestion
def op_callback(callback_fn):
```

lc0:
@googlebot I fixed it.

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+Such "hooks" will allow both observation and overriding of TF op's outgoing
+tensors, including concrete `EagerTensor`s and symbolic `Tensor`s (see details
+below.) This is a foundational part of the effort to bring [TensorFlow Debugger
+(tfdbg)](https://www.tensorflow.org/guide/debugger) up-to-date with TF2's
+execution paradigm.
+
+Currently, tfdbg is compatible with only the `tf.Session` API of TensorFlow 1.x.
+However, users of TF2 have raised questions and issues that indicate needs for
+a dedicated debugger, ones that cannot be easily met by a generic Python
+debugger such as pdb.  The most common examples of such needs involve
```
karmel: Can you clarify what is in scope given the existence of pdb? Is this designed to be a supplement that handles the specific use-cases given in this paragraph, or is it a replacement for pdb?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+To expand on the aims listed in the Object section, the proposed API will enable
+three key capabilities:
+
+  - **Capability A**: The ability to intercept eagerly-executing TF ops.
+    Specifically, we want a way to register one or more callbacks that are
+    invoked immediately after an `EagerTensor` has been computed through the
+    eager execution of an op. The callback should provide visibility into the
+    input and output `EagerTensor`s of the execution. If so desired, the
+    callback may _override_ the output `EagerTensor`s and thereby transparently
+    affect downstream eager execution.
```
karmel: Is overriding input also feasible?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+    assisted by [AutoGraph](https://www.tensorflow.org/guide/autograph). This
+    will form the basis for simulated stepping through lines of
+    `tf.function`-decorated Python functions to assist debugging of graph
+    construction in TF2. Like in Capability A, the callbacks should be able to
+    override the output symbolic tensors of the op in order to affect all
+    downstream graph ops.
+  - **Capability C**: Similar to Capability A above, we want the ability
+    to intercept the runtime execution of FuncGraphs. Note that this
+    requirement could be folded into Capability A, if FuncGraphs are
+    regarded as a special type of op.
```
karmel: To clarify, this means that for a FuncGraph generated by a tf.function, for example, you could inspect as if it were a single op, eg override the output tensors? And this would happen for the FuncGraph itself, rather than the call sites independently? Or am I misunderstanding the mode of operation here?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+
+  Args:
+    A callback_fn that has the following signature:
+      def callback_fn(op_type,
+                      inputs,
+                      attrs,
+                      outputs,
+                      op_name=None,
+                      graph=None):
+        # op_type: The type of the op. E.g., "MatMul".
```
karmel: As a string name? Or some sort of handle? Does this work for externally built ops as well (eg, from TF-Addons)?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+        #          For the special case of FuncGraph execution, op_type
+        #          takes the special value `tf.Graph`.
+        # inputs: (`tuple` of `Tensor`s) Input tensors to the op or the
+        #         FuncGraph.
+        #         In eager execution, these are `EagerTensor`s.
+        #         In graph construction, these are non-eager `Tensor`s
+        #         that form the inputs to the just-created op.
+        # attrs: The attributes of the op or FuncGraph of which the execution
+        #        or creation caused the current invocation of the callback.
```
karmel: I don't understand this-- can you explain further?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+        #          internally-generated name of the FuncGraph.
+        #          In graph construction, this is the name of the op.
+        # graph: The graph that the op belongs to (if any).
+        #        In eager execution of an op, this is `None`.
+        #        In eager execution of a FuncGraph, the is the FuncGraph
+        #        object per se.
+        #        In graph construction, this is the op's containing graph.
+        #
+        # Return values:
+        #   This callback function is expected to return a `list` or `tuple`
```
karmel: Presumably in the same order? Is there value in allowing different return types? Eg, nested things?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+      def callback_fn(op_type,
+                      inputs,
+                      attrs,
+                      outputs,
+                      op_name=None,
+                      graph=None):
+        # op_type: The type of the op. E.g., "MatMul".
+        #          For the special case of FuncGraph execution, op_type
+        #          takes the special value `tf.Graph`.
+        # inputs: (`tuple` of `Tensor`s) Input tensors to the op or the
```
karmel: How are composite tensors intended to be handled?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+Capability C is also met by this API, due to the fact that the callbacks are
+invoked not only during the eager execution of ops, but also the execution
+of FuncGraphs. In this paradigm, when a FuncGraph has been constructed and
+undergoes execution by TF2, it is simply treated as a special type of op.
+
+The example code snippet below shows how Capability A is met, i.e., how eager
+execution of ops is intercepted by the callback mechanism.
+
+```python
+with tf.debugging.op_callback(my_callback):
```
karmel: Can you include what my_callback is in this case? This is written by the user?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+  #      EagerTensor with value 1.386 (≈log(4.0)).
+```
+
+The code below illustrates how Capability B is met, i.e., how creation of
+ops inside a user-defined Python function decorated by `tf-function`:
+
+```python
+with tf.debugging.op_callback(my_callback):
+
+  @def_function
```
karmel: tf.function outside of TF?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+  #   arg being something like `_inference_log_1plusp_30`, and the `outputs` arg
+  #   being an EagerTensor of value 1.386 (≈log(4.0)).
+```
+
+The example above dealt with a relatively simple FuncGraph. But the proposed API
+applies to more complex FuncGraphs, including those that involve control flow
+such as if-else and while loops. For instance, see the code example below.
+
+```python
+with tf.debugging.op_callback(my_callback):
```
karmel: Is it worth considering more generic names given that this operates over non-ops? tf.debugging.callback() ?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+# Listing 2. Define a FuncGraph with `tf.debugging.op_callback()` and
+#            by passing a callback that overrides the symbolic output
+#            tensors of the just-created graph ops.
+
+def debugger_callback(op_type, inputs, attrs, outputs,
+                      op_name=None, graph=None):
+  instrumented_outputs = []
+  for output_slot, output in enumerate(outputs):
+    # Construct overriding output tensor.
+    instrumented_outputs.append(gen_debug_ops.debug_identity_v2(
```
karmel: Is gen_debug_ops.debug_identity_v2 a publicly available symbol? What will this look like for a user?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+  return tf.math.log(1 + x)
+```
+
+```python
+# Listing 2. Define a FuncGraph with `tf.debugging.op_callback()` and
+#            by passing a callback that overrides the symbolic output
+#            tensors of the just-created graph ops.
+
+def debugger_callback(op_type, inputs, attrs, outputs,
+                      op_name=None, graph=None):
```
karmel: Is there a set of built-in callbacks that would be useful to provide?

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+to the API.
+
+```python
+# Exported publicly as: tf.debugging.op_callback()
+def op_callback(callback_fn):
+  """Intercepts op execution and op creation.
+
+  The `callback_fn` will be invoked immediately after any of the three types
+  of events:
+    - The execution of an op under eager mode,
```
reedwm: Perhaps `callback_fn` should also be called if an op fails with a TypeError, e.g. when trying to add a float16 and a float32 tensor. In this case, there would be no outputs, and `attrs` would be partially unspecified.

The reason I mention this is that I am considering implementing a tool to help debug TypeErrors, by detecting where the float16 and float32 tensors originated, and recommending an appropriate place to insert a cast. This is currently impossible, as when a TypeError occurs there is no way of determining the input tensors to the op that failed with a TypeError. I am working on mixed precision, which is using a mix of float16 and float32, so TypeErrors occur fairly frequently.

/CC @alextp @azaks3

rfcs/20190815-tfdbg-v2-callbacks.md
```diff
@@ -0,0 +1,356 @@
...
+        #
+        # Return values:
+        #   This callback function is expected to return a `list` or `tuple`
+        #   of `Tensor`s, with its length matching `len(outputs)`.
+        #   In eager execution, these returned `Tensor`s should be
+        #   `EagerTensor`s. Their values will replace the original values of
+        #   `outputs` during downstream eager execution.
+        #   In graph construction, these returned `Tensor`s should be
+        #   non-eager `Tensor`s. Their values will replace the original
+        #   `outputs` during downstream graph construction.
```
reedwm: Can the output dtypes be different?

rfcs/20190815-tfdbg-v2-callbacks.md
caisq: Yes. Composite tensors including SparseTensors and RagggedTensors will be supported. That'll work under both eager execution and in tf.functions. In fact, given this is a low-level API, it doesn't "know" or "care" those are composite tensors. As a result, the op_types that show up in the callback will be all low-level ops such as "SparseTensorDenseMatMul".

I added a bullet point to the "Considerations for Various Cases" to reflect this.

rfcs/20190815-tfdbg-v2-callbacks.md
caisq: These are the attributes of the op. It is a tuple of alternating attribute key and attribute value, e.g., 

```
('adjoint_a', False, 'adjoint_b', False, 'Tindices', 9L, 'T', 1L, 'T', 1L)
```

I beefed up the doc string here to clarify it.

rfcs/20190815-tfdbg-v2-callbacks.md
caisq: Yes, e.g., 
1. a callback that crashes with a helpful error message and stack trace when any output tensor contains NaN or Infinity, including during eager execution and execution of intra-graph ops.
2. a callback that dumps numeric summary of all tenors (or a subset of them) to the filesystem during model training, to facilitate post hoc analysis of failed training jobs
3. a callback that sends debug info to a debug server via a grpc stream, to facilitate interactive debugging

But the detailed design is outside the scope of this doc. This is briefly mentioned in the newly added "What this callback API offers beyond pdb" section.

rfcs/20190815-tfdbg-v2-callbacks.md
caisq: `gen_debug_ops.debug_identity_v2` is not and won't be a public API that users can use. This is just an example. But the user can do something similar to this code example, using a user-defined op (see above) or built-in op. For instance, the following approach will negate all outputs in the instrumented graph.

```py
instrumented_outputs.append(tf.math.negative(output))
```

I added comments to the code snippet here to clarify that.

rfcs/20190815-tfdbg-v2-callbacks.md
caisq: Good question. I added a section to address this question. It is titled "What This Proposal Offers Beyond pdb". PTAL.

Briefly, this is not a replacement for pdb. Instead, it is a TensorFlow-specific debugging instrumentation that can supplement pdb.

rfcs/20190815-tfdbg-v2-callbacks.md
caisq: I don't see why this API should prohibit changing the dtype. But beware that if you override the dtype (e.g., float32 to int64), downstream graph construction or eager execution may likely error out with a dtype-not-supported error.

rfcs/20190815-tfdbg-v2-callbacks.md
caisq: Interesting use case. IIUC, the API will be able to help you debug that. The callback will be invoked during upstream graph construction. The callback can store the graph-construction history somewhere (e.g., on disk), so you know how the graph is built, op by op. When the program errors out, you can load the history and reason, op by op, why the dtype is wrong.

There is no particular need to call the callback when the error happens. The information is there in the error itself.

How does that sound?

reedwm: The issue is that when the TypeError occurs, you don't know what inputs caused the TypeError, even if you record the graph-construction history from past successful ops. For example, the following programs raises a TypeError

```
a = tf.constant(1., dtype='float16')
b = tf.constant(1., dtype='float32')
c = a + b
```

I could record the graph-construction history, so I could know how `a` and `b` were created. But when the TypeError occurs, I don't know that the inputs that caused the TypeError were `a` and `b`: They could have been some other tensors that were previously created instead. The TypeError itself does not state the inputs, as in Eager mode, tensors do not have names so there would be no way of identifying them.

If this cannot be resolved, it's no big deal. This is a fairly minor use case.

caisq: @reedwm I see. It seems to me that this issue can be addressed by printing these two extra pieces of info in the error message for dtype error:
1. The name(s) of all input tensors to the op that failed to be created.
2. The name of the graph being constructed (if any).

reedwm: That works in graph mode, but in Eager mode, tensors do not have names.

Still, this doesn't have to be resolved. I could always make the tool require tf.functions to be used.

rfcs/20190815-tfdbg-v2-callbacks.md
caisq: Added further clarification in the code comment.

rfcs/20190815-tfdbg-v2-callbacks.md
caisq: During offline discussion, we agreed that an callback that errors out when an op's execution or construction fails is outside the scope of the design, and so is a "pre-op" callback. Those features make sense and may be added under the tf.debugging.* namespace in the future.

rfcs/20190815-tfdbg-v2-callbacks.md
caisq: (See the discussion with reedwm@ below about pre-op callback.)

caisq:
@ewilderj @goldiegadde @martinwicke @theadactyl Following today's design review, I've resolved all the questions and tackled a small number of TODO items. Per the process guideline, I've flipped the status of the RFC to "Accepted". PTAL.

ewilderj:
@casiq thank you. are there notes from the review?

caisq:
@ewilderj Yes, thanks to @alextp for taking the notes.

- Changes since sending it out?
  - [cais] comments by karmel and reed; responses to those comments. added section comparing this proposal and pdb
- Major questions
  - Do we need to change the inputs?
    - [karmel] user might want to patch the data before executing ops to debugging
    - [cais] that would be a pre-op callback
    - [apassos] do we need to separately provide a pre-op callback api? it’d most naturally fit with a slightly different signature to this so separate
    - [cais,karmel] let’s agree about pre-op callbacks being out of scope
    - [reedwm] if we had callbacks before resolving attributes it’d solve some issues for me
    - [apassos] that would be harder than this because of how the generated code works
    - [reedwm] all we’d need is a callback when the error occurs or a way to identify a tensor when an error occurs
    - [cais] we could do something like that in the tfdbg namespace
    -[apassos] worry about scope creep, let’s have a separate conversation
  - Short discussion about user-defined ops: they are covered as long as the user uses the python op binding code generator
  - Short discussion on composite tensor support. It’s implied by composites not existing at the graph level. Possible to debug those, but need to be aware of how the components map to actual ops.
  - [cais] naming question
    - [cais] op_callback more appropriate than callback
    - [karmel] op_callback might not make it clear functions are covered; careful docstrings are fine to address this
    - [cais] This API covers eager ops and graphs and graphs contains ops. The concern can be addressed with good documentation and examples, which we plan to include.
  - [karmel] unclear how users use these callbacks in practice. maybe more examples is enough
    - [cais] Will add more to the examples [DONE]
    - [cais] ideally you just return the original outputs and accumulate some information about them, but you can also override them if you want
    - [karmel] docs / a guide for this would be helpful
  - [reedwm] will we support tfdbg v1 mode which dumps all the tensors?
    - [cais] this will be supported but it’s very expensive in practice

ewilderj:
Thank @caisq and everyone who left feedback and participated in the design. Merging this now. Congratulations on the RFC!

PR #130, opened 2019-08-03
RFC: TensorFlow Official Model Garden Redesign

rachellj218:
This RFC will be open for comment until Friday, August 16th, 2019.

# TensorFlow Official Model Garden Redesign

|**RFC#**| 130|
| **Author(s)** | Jing Li (jingli@google.com), Hongkun Yu (hongkuny@google.com), Xiaodan Song (xiaodansong@google.com) |
| **Sponsor**   | Edd Wilder-James (ewj@google.com)                 |

## Objective

This document presents a proposal to redesign TensorFlow official model garden.
We aim to provide a central and reliable place to contain popular examples,
state-of-the-art models and tutorials to demonstrate the best practice in TF2.0
and illustrate real-world use cases.

### DIFF
--- a/rfcs/20190802-model-garden-redesign.md
+++ b/rfcs/20190802-model-garden-redesign.md
@@ -58,7 +58,8 @@ We are going to reorganize the official model directory to provide:
         layer/network/model to define and utilize model building blocks.
         
         **NOTE:** we are still figuring out what level of building block extraction would be the most useful and sharable
-        during refactoring. Once we confirm the implementation is really useful, we will move it tensorflow/addons. 
+        during refactoring. Once we confirm the implementation is really useful, we will move it tensorflow/addons and/or tf.text.
+
 *   popular state-of-the-art (SOTA) models for end users as a product.
 *   reference models for performance benchmark testing.
     *   For models provided as SOTA models, we will share the network and
@@ -72,7 +73,7 @@ structure. The SOTA model list will be updated to cover more categories.
 | Directory     | Subdirectories       |   | Explainations                   |
 :-------------- |:---------------------|:--|:------------------------------ |
 | modeling      |                      |   | Common modeling libraries       |
-|               | layers               |   | Temporary folder to hold common modules/layers, not built-in tensorflow layers yet during refactoring. Will move to tensorflow/addons finally.|
+|               | layers               |   | Temporary folder to hold common modules/layers, not built-in tensorflow layers yet during refactoring. Broadly used layers will be graduated to tensorflow/addons and/or tf.text.|
 |               | networks             |   | Well-known networks built on top of layers, e.g. transformer |
 |               | training             |   | Training utils, e.g. example custom training loop |
 | utils         |                      |   | Miscellaneous Utilities         |
--- a/rfcs/20190802-model-garden-redesign.md
+++ b/rfcs/20190802-model-garden-redesign.md
@@ -72,53 +72,27 @@ structure. The SOTA model list will be updated to cover more categories.
 
 | Directory     | Subdirectories       |   | Explainations                   |
 :-------------- |:---------------------|:--|:------------------------------ |
-| modeling      |                      |   | Common modeling libraries       |
-|               | layers               |   | Temporary folder to hold common modules/layers, not built-in tensorflow layers yet during refactoring. Broadly used layers will be graduated to tensorflow/addons and/or tf.text.|
-|               | networks             |   | Well-known networks built on top of layers, e.g. transformer |
-|               | training             |   | Training utils, e.g. example custom training loop |
-| utils         |                      |   | Miscellaneous Utilities         |
-|               | hyperparameters      |   | Common flags and model parameters. |
-|               | ...                  |   |                                 |
-| benchmarks    |                      |   | benchmark testing and reference models to validate tensorflow |                                |
-|               | utils                |   |                                 |
-|               | examples             |   | reference models for testing/validating end-to-end tensforflow |
-|               |                      | Resnet |                            |
-|               |                      | BERT   |                            |
-|               |                      | Transformer |                       |
-|               |                      | NCF   |                             |
 | nlp           |                      |   | models/tasks for Natural Language Processing |
-|               | utils                |   | NLP specific utils, e.g. input dataset |
+|               | modeling             |   | NLP modeling library |
 |               | BERT                 |   | NLP specific utils, e.g. input dataset |
-|               |                      | BERT core modeling |  |
-|               |                      | tasks | specific tasks on open public datasets, e.g Squad, MNLI |
+|               | ALBERT                    | BERT core modeling |  |
 |               | XLNET                |    |   |
-|               | GPT                  |    |   |
 |               | Transformer          |    |   |
-|               | GNMT                 |    |   |
 |               | ...                  |    |   |
 | vision        |                      |    | models/tasks for Computer Vision |
-|               | image_classification |    |   |
-|               |                      | resnet |   |
-|               |                      | EfficientNet   |   |
-|               |                      | MnasNet   |   |
-|               |                      | ...   |   |
-|               | detection            |       |   |
-|               |                      | RetinaNet    |   |
-|               |                      | Mask-RCNN    |   |
-|               |                      | ShapeMask    |   |
-|               |                      | ...    |   |
-|               | segmentation         | UNet     |   |
-|               |                      | DeepLab v3/v3+ |   |
-|               |                      | FastSCNN |   |
+|               | image_classification |    | e.g. resnet, EfficientNet, ... |
+|               | detection            |    | e.g. RetinaNet, Mask-RCNN, ... |
+|               | ...                  |    |  |
 | recommendation|                      |     |   |
 |               | NCF                  |     |   |
-| staging       |                      |     | accepting community contributions  |
-| archive       |                      |     | deprecated models, not officially supported  |
+| utils         |                      |   | Miscellaneous Utilities.          |
+|               | ...                  |   |                                 |
+| benchmarks    |                      |   | benchmark testing and reference models to validate tensorflow |
+| staging       |                      |     | Utilities not in TF core yet, and not suitable for tf addons  |
 | r1            |                      |     | tf1.x models and utils  |
 |               | utils                |     |   |
 |               | resnet50             |     |   |
 |               | transformer          |     |   |
-|               | ncf                  |     |   |
 |               | wide_deep            |     |   |
 |               | boosted_trees        |     |   |
 
--- a/rfcs/20190802-model-garden-redesign.md
+++ b/rfcs/20190802-model-garden-redesign.md
@@ -56,8 +56,9 @@ We are going to reorganize the official model directory to provide:
     *   Model category related common library, e.g. primitives as basic building
         block for NLP models, or common networks like resnet, mobilenet. We will follow the fundamental design of Keras
         layer/network/model to define and utilize model building blocks.
+        
         **NOTE:** we are still figuring out what level of building block extraction would be the most useful and sharable
-        refactoring. Once we confirm the implementation is really useful, we will move it tensorflow/addons. 
+        during refactoring. Once we confirm the implementation is really useful, we will move it tensorflow/addons. 
 *   popular state-of-the-art (SOTA) models for end users as a product.
 *   reference models for performance benchmark testing.
     *   For models provided as SOTA models, we will share the network and
--- a/rfcs/20190802-model-garden-redesign.md
+++ b/rfcs/20190802-model-garden-redesign.md
@@ -1,0 +1,207 @@
+# TensorFlow Official Model Garden Redesign
+
+| Status        | Proposed       |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | Jing Li (jingli@google.com), Hongkun Yu (hongkuny@google.com), Xiaodan Song (xiaodansong@google.com) |
+| **Sponsor**   | Edd Wilder-James (ewj@google.com)                 |
+| **Updated**   | 2019-08-02                                           |
+
+## Objective
+
+This document presents a proposal to redesign TensorFlow official model garden.
+We aim to provide a central and reliable place to contain popular examples,
+state-of-the-art models and tutorials to demonstrate the best practice in TF2.0
+and illustrate real-world use cases.
+
+## Motivation
+
+The current [TF official model garden](https://github.com/tensorflow/models/tree/master/official)
+mainly has ad hoc support. Example models are implemented using mixed TensorFlow
+APIs in different coding styles and some of them have convergence and/or
+performance regression. With TensorFlow 2.0 launch, there’s a great desire to
+provide tensorflow users a clear and central place to showcase reliable TF2.0
+models with the best practices to follow.
+
+We want to take this opportunity to substantially improve the state of the
+official model garden, and provide seamlessly end-to-end training and inference
+user experience on a wide range of accelerators and mobile device chips. We hope
+to encourage community to contribute innovations and improve TensorFlow
+efficiency and usability.
+
+## User Benefit
+
+We aim to provide the best modeling experience via this revamp effort:
+
+*   Usability and reliability
+    *   keep official models well-maintained and tested for both performance and
+        convergence.
+    *   provide accessible model distribution via [TensorFlow Hub](https://www.tensorflow.org/hub) and share state-of-the-art research accomplishments.
+    *   make training on both GPU and TPU an easy switch.
+    *   provide reusable components for research and production.
+*   End-to-end solutions
+    *   provide seamless end-to-end training and inference solutions, where inference covers serving on TPU, GPU, mobile and edge devices.
+    *   provide hyper parameter sets to tune models for various resource constraints.
+    *   provide solutions with hyper parameters to scale model training to TPU pods or multi-worker GPUs.
+    *   provide variants derived from standard models to tackle various practical tasks.
+
+## Design Proposal
+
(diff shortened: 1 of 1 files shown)


rfcs/20190802-model-garden-redesign.md
```diff
@@ -0,0 +1,207 @@
...
+The following details the policy for models in mature and staging phases:
+
+*   Models graduated from staging subdirectory
+
+    The models will be maintained by the model garden team. After we start to
+    accept community contributions, we will put the contributors as model owners.
+
+    These models will have continuous convergence and performance testing to
+    make sure no regression. In general, we won’t deprecate these models unless:
+    *   the model isn’t compatible with the TF APIs any more and have to be replaced by a new version
```
jackwish: Does this imply that the model will only target *latest* version of TF?

rfcs/20190802-model-garden-redesign.md
```diff
@@ -0,0 +1,207 @@
...
+    *   provide solutions with hyper parameters to scale model training to TPU pods or multi-worker GPUs.
+    *   provide variants derived from standard models to tackle various practical tasks.
+
+## Design Proposal
+
+### Official model directory reorgnization
+
+We are going to reorganize the official model directory to provide:
+
+*   common libraries, mainly two types:
```
jayfurmanek: Moving common code into a central location is always a good design practice, but let's be careful not create another SLIM here. There are *still* models (Deeplab anyone) that are chained to SLIM.

Remember, part of what would be valuable here is showing good examples on how to write a *model*, not necessarily good examples on how to write a model *garden* with intertwined and complicated dependencies.

rfcs/20190802-model-garden-redesign.md
```diff
@@ -0,0 +1,207 @@
...
+## Objective
+
+This document presents a proposal to redesign TensorFlow official model garden.
+We aim to provide a central and reliable place to contain popular examples,
+state-of-the-art models and tutorials to demonstrate the best practice in TF2.0
+and illustrate real-world use cases.
+
+## Motivation
+
+The current [TF official model garden](https://github.com/tensorflow/models/tree/master/official)
```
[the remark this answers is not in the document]
jayfurmanek: I have this same question.

rfcs/20190802-model-garden-redesign.md
```diff
@@ -0,0 +1,207 @@
...
+### Official model directory reorgnization
+
+We are going to reorganize the official model directory to provide:
+
+*   common libraries, mainly two types:
+    *   Common training util library in TF2.0, model configuration and
+        hyperparameter definition in a consistent style.
+    *   Model category related common library, e.g. primitives as basic building
+        block for NLP models. We will follow the fundamental design of Keras
+        layer/network/model to define and utilize model building blocks.
```
fchollet: Note: `Network` is a private class, so objects will need to subclass either `Layer` or `Model`.

Also note that whenever possible, we should prefer the pattern of Functional model + custom layers, rather than subclassed models (which have less functionality).

rfcs/20190802-model-garden-redesign.md
```diff
@@ -0,0 +1,207 @@
...
+|               | Transformer          |    |   |
+|               | GNMT                 |    |   |
+|               | ...                  |    |   |
+| vision        |                      |    | models/tasks for Computer Vision |
+|               | image_classification |    |   |
+|               |                      | resnet |   |
+|               |                      | EfficientNet   |   |
+|               |                      | MnasNet   |   |
+|               |                      | ...   |   |
+|               | object_detection     |    |   |
```
AakashKumarNain: Unet is mainly used for segmentation. So instead of `object_detection`, rather make two sub-directories `object_detection` and `segmentation`. `segmentation` will then contain:
* `DeepLab v3/v3+`
* `Unet`
* `FastSCNN`
...

rfcs/20190802-model-garden-redesign.md
AakashKumarNain: I agree on the last part. `models/research` is a typical example of this. The whole garden there is so complex and huge that it is nowhere near usable for a project until unless you:
* Just want the model off the shelf without any modification
* You know in an out of each and every part of it

rfcs/20190802-model-garden-redesign.md
```diff
@@ -0,0 +1,207 @@
...
+|               | ncf                  |     |   |
+|               | wide_deep            |     |   |
+|               | boosted_trees        |     |   |
+
+### Pretrained model repository
+
+We are going to provide the pretrained models for research exploration and
+real-world application development. The plan is to integrate with [TensorFlow Hub](https://www.tensorflow.org/hub),
+where users can access the Hub modules and SavedModel for pretrained checkpoints and links to the code in the model
+garden.
```
AakashKumarNain: Please avoid use of `configs` as much as possible. The code should be flexible enough to change the parameters in the code itself. Changing things though a `config` file makes sense for some people but not for all. It is easy to maintain the changes in the code but it is hard to track the changes in the config file and at the same time, it is hard to validate the changes as well

rfcs/20190802-model-garden-redesign.md
```diff
@@ -0,0 +1,207 @@
...
+
+The following table shows the detailed view of proposed model directory
+structure. The SOTA model list will be updated to cover more categories.
+
+| Directory     | Subdirectories       |   | Explainations                   |
+:-------------- |:---------------------|:--|:------------------------------ |
+| modeling      |                      |   | Common modeling libraries       |
+|               | layers               |   | Common modules/layers, not built-in tensorflow layers yet.|
+|               | networks             |   | Well-known networks built on top of layers, e.g. transformer |
+|               | optimziers           |   | New or customized optimizers    |
```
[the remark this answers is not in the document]
lc0: Same with optimizers https://github.com/tensorflow/addons/tree/master/tensorflow_addons/optimizers

rfcs/20190802-model-garden-redesign.md
jackwish: Same concern. Software engineering design may not be applied to deep learning model garden these days.

Personally, a model *garden* is simply somewhere user can obtain **the whole runnable model** from a very sub directory. Sharing common library, and any similar things, sounds like a SDK built on top of TF to me. Maybe we can have TF SDK, and then the garden.

rfcs/20190802-model-garden-redesign.md
```diff
@@ -0,0 +1,207 @@
...
+        module for benchmark testing will have addtional flags and setups for
+        performance testing.
+
+The following table shows the detailed view of proposed model directory
+structure. The SOTA model list will be updated to cover more categories.
+
+| Directory     | Subdirectories       |   | Explainations                   |
+:-------------- |:---------------------|:--|:------------------------------ |
+| modeling      |                      |   | Common modeling libraries       |
+|               | layers               |   | Common modules/layers, not built-in tensorflow layers yet.|
```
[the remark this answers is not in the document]
seanpmorgan: Splintering the TF ecosystem with multiple repos that contain new layers seems counter-productive. Addons is already going to manage the burden of graduating layers into core so why have 2 repos for the same thing. 

cc @karmel @facaiy for visibility

rfcs/20190802-model-garden-redesign.md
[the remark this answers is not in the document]
seanpmorgan: Same concern, addons has additional optimizers that can be graduated to core. 

Any optimizer/layer that isn't useful as a supplementary package can just be committed to the model garden repository IMO.

rfcs/20190802-model-garden-redesign.md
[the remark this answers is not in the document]
AakashKumarNain: With this, we are going in the direction of TF1.x again.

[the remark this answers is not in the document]
karmel: @rachellj218 , I believe we discussed having a process to ensure any broadly useful layers/etc make it to Addons and/or tf.text. We should update the doc to reflect that, as it seems to be a common concern-- maybe a section describing what you would like the evaluation/graduation process to be?

rfcs/20190802-model-garden-redesign.md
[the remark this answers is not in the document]
rachellj218: The current plan is still to keep the model garden lie in the current directory. But we are going to work with TF Hub to provide a unified UI to provide both pretrained models and links to model codes, hopefully it will make it easier for users to find. Thanks!

rfcs/20190802-model-garden-redesign.md
rachellj218: Thanks for the suggestions! We do plan to follow the pattern of functional model + custom layers.

rfcs/20190802-model-garden-redesign.md
rachellj218: We plan to do model garden release for major TF release, e.g. TF 2.1, 2.2. Thanks!

rfcs/20190802-model-garden-redesign.md
rachellj218: Thanks for your suggestion! We plan to follow the practice in cloud tpu model repo, for example, https://github.com/tensorflow/tpu/blob/master/models/official/mnasnet/mnasnet_main.py#L674. A default parameter dictionary will be provided, but users can overwrite the default value or extend to new key/value with string, dict or yaml.

rfcs/20190802-model-garden-redesign.md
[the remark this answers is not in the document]
rachellj218: Thanks for raising this! As karmel@ mentioned, we do have the plan to graduate the common layers to tensorflow/addons. I clarified in the RFC. Thanks!

I removed optimizers/ subdirectory. I agree it should be good to add to Addons directly.

rfcs/20190802-model-garden-redesign.md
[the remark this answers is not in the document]
rachellj218: Removed optimizers subdirectory.

rfcs/20190802-model-garden-redesign.md
rachellj218: Thanks! Updated to reflect the suggestion. We are still debating the best structure for vision models, this may subject to changes in near future.

rfcs/20190802-model-garden-redesign.md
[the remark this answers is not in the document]
rachellj218: Thanks! We do have the plan to graduate the common layers to tensorflow/addons. I clarified in the RFC. Thanks!

optimizers/ subdirectory was removed from RFC. I agree it should be good to add to Addons directly.

rfcs/20190802-model-garden-redesign.md
rachellj218: Thank you for raising this concern! We definitely need a good balance between implementation duplication and complicated dependencies. The current design is to split the modeling process to two stages: 1)common networks (in modeling directory), such as resnet, transformer, so that these networks can be reused in specific model/task, e.g. resnet50, mask r-cnn, sequence model; 2) specific model/task along with public dataset (e.g. in NLP and vision directory), models will be defined in its own sub directory, data preprocessing and eval and other utils for the same type of tasks/datasets can be shared. 

If you have better suggestions, we are definitely open to them. Thanks again!

rfcs/20190802-model-garden-redesign.md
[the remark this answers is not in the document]
jackwish: I am not sure what others' thoughts, but to me, *model garden directory* is `https://github.com/tensorflow/models/tree/master`, while the *official* means the `official` sub directory...

rfcs/20190802-model-garden-redesign.md
jackwish: Cool! Thanks for your reply.

rfcs/20190802-model-garden-redesign.md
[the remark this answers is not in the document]
jackwish: Thanks for your clarification! In the latest RFC:
> Temporary folder to hold common modules/layers, not built-in tensorflow layers yet during refactoring. Broadly used layers will be graduated to tensorflow/addons and/or tf.text.

It seems to me that *graduation* will be another SLIM. The *graduation* will create different versions of one network which are based on different API, say a network uses 3 layers A, B and C firstly introduced in model garden, these 3 layers move to TF Addons one by one, thus we have four versions of network code. That is pretty confusing. And, as the interfaces of these layers are probably going to be different, it won't be as easy as re-write one line. Eventually, the effort to maintain these code will be huge.

What about implement the layers **directly** in TF Addons, which I think is much more flexible than the TF repo?

Thanks, hoping that my concern won't be too annoying...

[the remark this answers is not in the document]
seanpmorgan: I agree with this concern. IMO it's less of a headache if they are moved from the temporary folder to Addons prior to a model-garden pip release. We can promise a quick review and patch release for official model additions. To ensure this we could have a model-garden team member on addons team.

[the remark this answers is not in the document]
saberkun: I think we need a consolidation stage for modeling components. This is not to develop infra but write models. We will guarantee layers moved to tf-addon must be removed in model garden. It is also up to how do you define "layers" as some of them are not common components as the ones in tf-addon now.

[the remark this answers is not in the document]
facaiy: > To ensure this we could have a model-garden team member on addons team.

+1, good idea. We can work in close coordination. 

Maybe model garden can put its private layers in each model's module. And when you find that some layers need be shared by two models or more, I believe it's a good time to move those layers to tf.addons, rather than modeling.layers. What do you think?

martinwicke:
I like that as a policy.

ewilderj:
@rachellj218 @martinwicke What was the fate of this RFC? Did you have a design review meeting that accepted it? Could you post the notes?

bhack:
It is a little bit of topic but new stuff like [Dectectron2](https://github.com/facebookresearch/detectron2) or High Level task API + models create competitive pressure over Garden/Hub.
Please take care super-care about modernizing old TF Object API and similar high level task approaches in the Garden design.

rfcs/20190802-model-garden-redesign.md
```diff
@@ -0,0 +1,212 @@
...
+We aim to provide the best modeling experience via this revamp effort:
+
+*   Usability and reliability
+    *   keep official models well-maintained and tested for both performance and
+        convergence.
+    *   provide accessible model distribution via [TensorFlow Hub](https://www.tensorflow.org/hub) and share state-of-the-art research accomplishments.
+    *   make training on both GPU and TPU an easy switch.
+    *   provide reusable components for research and production.
+*   End-to-end solutions
+    *   provide seamless end-to-end training and inference solutions, where inference covers serving on TPU, GPU, mobile and edge devices.
```
bhack: Will this include Google Coral devices?

theadactyl:
@rachellj218 @martinwicke pinging again for notes and/or resolution for this RFC.

bhack:
Can we try to extend the deadline if needed? It Is still set on August 16th, 2019.

bhack:
The garden needs gardeners.. It is turning into a jungle again. 
Not much different from the current zoo :smile:

rfcs/20190802-model-garden-redesign.md
pavanky: > The current design is to split the modeling process to two stages: 1)common networks (in modeling directory), such as resnet, transformer, so that these networks can be reused in specific model/task, e.g. resnet50, mask r-cnn, sequence model

How will these be different from https://www.tensorflow.org/api_docs/python/tf/keras/applications ?

bhack: Also, Keras applications are going on an external repo x Keras migration RFC at https://github.com/tensorflow/community/pull/202. 
Right?

bhack:
We really need a more close coordination between HUB and the Garden. Models are spread in too much repositories/places... we really need an Italian style Garden :smile: 
Many non google model/papers reference or third party implementation are published in pytorch on github so this is a super strategic topic to push Tensorflow.
Honestly it is incomprehensible how it is not treated with a TOP priority

rachellj218:
Thank you for providing feedback, @bhack! Here're the differences between TF Hub and Model Garden:
- TensorFlow Hub (https://tfhub.dev) is the main entry point (a search engine for all TensorFlow pre-trained models) for TF users who wants to find a pre-trained model, and it provides a pathway to (1) code examples and (2) tutorials (e.g., Colab notebooks) for all TensorFlow 1 & 2 models (TF, TFLite, TF.js, Coral).
- Model Garden (https://github.com/tensorflow/models) provides TensorFlow 2 model implementations and E2E modeling solutions with code examples. The pretrained checkpoints are exported to TFHub for user access. 
The TFHub team is actively working on improving user experience.

We do realize that the current model garden design isn't good enough. We are actively working on redesigning the model garden. We will make an announcement on our new model garden redesign once ready.

bhack:
I see code example in both the tw bullet points and models are in other repos. I think the entry point need to be unfied. 
I don't think that we need any duplicate elementi between Garden and Hub but coordination between teams for a unified goal.
See "Pre-made Models with Explicit Support for Modification" in https://ai.googleblog.com/2020/03/toward-human-centered-design-for-ml.html

saberkun:
> I see code example in both the two bullet points and models are in other repos. I think the entry point need to be unfied.
> I don't think that we need any duplicate elementi between Garden and Hub but coordination between teams for a unified goal.
> See "Pre-made Models with Explicit Support for Modification" in https://ai.googleblog.com/2020/03/toward-human-centered-design-for-ml.html

There should not have duplicated elements. Hub is the entry point offering links to the code examples and Garden is the place holding solid modeling code for a selected collection of models.

PR #131, opened 2019-08-07
RFC: Java Repositories

karllessard:
**Comment period is open through 2019-08-30**

# RFC: Java Repositories

| **Author**    | Karl Lessard (karl.lessard@gmail.com) |
| **Sponsor**   | James Ring (Google) |

## Objective

Create new repositories under the `github.com/tensorflow` organization to host the code supported by SIG JVM, including the actual Java client found in TensorFlow core repository.

### DIFF
--- a/sigs/jvm/rfcs/20190802-java-repositories.md
+++ b/sigs/jvm/rfcs/20190802-java-repositories.md
@@ -1,5 +1,5 @@
 # Java Repositories
-| Status        | Proposed       |
+| Status        | Accepted       |
 :-------------- |:---------------------------------------------------- |
 | **Author**    | Karl Lessard (karl.lessard@gmail.com) |
 | **Sponsor**   | James Ring (Google) |
--- a/sigs/jvm/rfcs/20190802-java-repositories.md
+++ b/sigs/jvm/rfcs/20190802-java-repositories.md
@@ -1,0 +1,76 @@
+# Java Repositories
+| Status        | Proposed       |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | Karl Lessard (karl.lessard@gmail.com) |
+| **Updated**   | 2019-08-02                                           |
+
+## Objective
+
+Create new repositories under the `github.com/tensorflow` organization to host the code supported by SIG JVM, including the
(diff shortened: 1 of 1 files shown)


karllessard:
@googlebot I signed it!

karllessard:
I just asked the SIG JVM group if they would prefer to rename the repositories to "/tensorflow/jvm" and "/tensorflow/jvm-models", instead of their "java" counterpart.

Sounds like "jvm" makes more sense if we intend to host code and samples for JVM languages other than Java in those repos as well. Waiting to be confirmed.

CC: @ewilderj

ewilderj:
This might add a couple more days as I need to revisit the required repo names, but shouldn't be too much of a delay. Is it the consensus of the group that jvm is preferred to 'java'? If so, we ought to push an update to this RFC as well.

karllessard:
Yeah, sorry for the trouble, I just poked the SIG this morning to see if
they have a preference, I'll keep you posted shortly of our final decision
(I'll try to know before for Wednesday).

Thanks,
Karl

saudet:
@ewilderj What do the teams at Google usually do for polls like this?

ewilderj:
> @ewilderj What do the teams at Google usually do for polls like this?

Google Forms!

btw we now have approval for making the java and java-models repos. I suggest we make these and treat any renaming as a second step.

karllessard:
@ewilderj : ok, you know better what it implies to apply those changes than we do so if you think it would be easier to rename them after, then let’s go for it!

karllessard:
@ewilderj, like any good ol' survey, the result is a tie: half wants to preserve "java" and half wants to rename it to "jvm".

Therefore, I guess the status quo wins, which is to keep the original "/java" and "/java-models" names.

ewilderj:
The repositories are now made. @karllessard @sjamesr and @tzolov you have invitations to the @tensorflow/sig-jvm-maintainers team, which will give you write perms on the repo. Work can commence! 

@karllessard please push an update to this PR changing Status to "Accepted", then I will merge this RFC as part of the record.

karllessard:
Wonderful, that’s very good news, thanks Edd!!

PR #133, opened 2019-08-14
RFC: Kernel and Op Implementation and Registration API

sjamesr:
*Comment period is open until 2019-08-28*

# Kernel and Op Implementation and Registration API

| **Author(s)** | James Ring (sjr@google.com), Anna Revinskaya (annarev@google.com) |
| **Sponsor**   | Günhan Gülsoy (gunan@google.com)                                  |
| **Updated**   | 2019-08-14                                                        |

## Objective

Tensorflow (TF) currently provides a C++ API for implementing kernels and ops. The Voltron project aims to create a modular/plugin-based TF implementation with API and ABI surfaces. Plugins will be able to create and register custom kernel and op implementations.

In order to provide a stable ABI, the Voltron team has chosen to provide C APIs to plugin authors. This document introduces the C API for op and kernel registration. For authors who wish to continue using C++ to interface with TensorFlow, an ABI-stable C++ header-only API is provided.


rfcs/20190814-kernel-and-op-registration.md
```diff
@@ -0,0 +1,658 @@
...
+the bitcast kernel using the C API:
+
+```c++
+#include "tensorflow/c/kernels.h"
+
+typedef struct bitcast_kernel { … } bitcast_kernel;
+
+// Bitcast_Create, Bitcast_Compute and Bitcast_Delete actually implement the
+// kernel. See the section below for discussion on kernel implementation.
+static void* Bitcast_Create(TF_OpKernelConstruction* context) {
```
alextp: In TF today op kernel construction can fail (and if often does, as it validates attributes etc). This should have a status somewhere.

rfcs/20190814-kernel-and-op-registration.md
```diff
@@ -0,0 +1,658 @@
...
+  bitcast_kernel* k = (bitcast_kernel*) calloc(1, sizeof(bitcast_kernel));
+  /* initialize the fields of k as needed */
+  return (void*) k;
+}
+
+static void* Bitcast_Compute(void* k, TF_OpKernelContext* context) {
```
alextp: What is the return value here? Outputs are not supposed to be allocated and then returned (they are allocated via the context) so I don't know why this is not void.

rfcs/20190814-kernel-and-op-registration.md
```diff
@@ -0,0 +1,658 @@
...
+occur.
+
+```c++
+TF_KernelBuilder* TF_NewKernelBuilder(
+  const char* op_name, const char* device_name,
+  void* (*create_func)(TF_OpKernelConstruction*),
+  void (*compute_func)(void*, TF_OpKernelContext*),
+  void (*delete_func)(void*));
+
+void TF_RegisterKernelBuilder(const char* name, TF_KernelBuilder* builder,
```
alextp: Why two functions instead of one function?

rfcs/20190814-kernel-and-op-registration.md
```diff
@@ -0,0 +1,658 @@
...
+    int start,
+    int limit);
+
+TF_CAPI_EXPORT extern void TF_ThreadPool_ParallelFor(
+    TF_OpKernelContext* context,
+    int64_t total,
+    int64_t cost_per_unit,
+    void (*fn)(int64_t, int64_t));
+
+TF_CAPI_EXPORT extern void TF_ThreadPool_ParallelForWithWorkerId(
```
alextp: One thing which is not obvious when reading this proposal is that in C++ we need to wrap almost every call to a method in the context with the op_requires_ok macro to terminate execution in case of an error.

The same will need to be done here as all of these can fail, so I think we need to be explicit about the code the user is expected to write.

rfcs/20190814-kernel-and-op-registration.md
sjamesr: OpKernelConstruction itself contains a status which can be set with TF_OpKernelConstruction_Failure. I'll update the doc.

rfcs/20190814-kernel-and-op-registration.md
sjamesr: This is a typo

rfcs/20190814-kernel-and-op-registration.md
sjamesr: The kernel builder may be customized with other calls, such as `TF_KernelBuilder_TypeConstraint` and `TF_KernelBuilder_HostMemory`, I'll add these to the doc.

rfcs/20190814-kernel-and-op-registration.md
annarev: Context is just used to call accessors, for e.g.:
`auto thread_pool = context->device()->tensorflow_cpu_worker_threads()->workers;`

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/matrix_diag_op.cc#L323

We could modify ThreadPool interface though to also return a Status. Then, we can also take TF_Status output parameter in C API.

annarev: I added a section below called "Getting Status when using device APIs". Basically, since the status is not on the surface of APIs that we are wrapping (we could add it to ThreadPool, but probably not to Eigen::StreamInterface), I propose to have a separate function call to get status.

Let me know if this sounds good.

annarev:
@karllessard sorry I missed your question earlier. See replies inline.
> I don't know if this is in the scope of this RFC but how will be registered ops be exposed through the C API? Right now, both [Ops defs](https://github.com/tensorflow/tensorflow/blob/2725b33c0f86852b5ce33c74ee4aac19932b042d/tensorflow/c/c_api.h#L1452) and [API defs](https://github.com/tensorflow/tensorflow/blob/2725b33c0f86852b5ce33c74ee4aac19932b042d/tensorflow/c/c_api.h#L1492) are returned as proto buffers, will that remain the same?

Yes, the way Op defs and API defs are returned will remain the same. These accessors are already a part of our C API. But let us know if you think some part of the API won't work well for some languages.

> Since at some points we want to move out the different language bindings out of the core repository (e.g. Java), for me it does not make sense to keep the per-language API defs in there. What would be the best strategy then?

I think language bindings can already be removed with current API. You can keep ApiDef files with the language bindings (they are just read when generating language APIs and aren't used anywhere else).  For example, Go API is generated based on a set of directories that contain ApiDef files: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/go/genop/main.go#L39 that are added using calls to TF_ApiDefMapPut (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/go/genop/internal/api_def_map.go#L91). So, I would think we can move language API designed in this way to a separate repo. However, I haven't checked how Java APIs are generated.

karllessard:
> However, I haven't checked how Java APIs are generated.

The Java API works pretty much the same, it has a op generator written in C++ that uses directly TF internal classes to read the API defs found in [this directory](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/core/api_def/java_api) merged with the base one.

But what I'm wondering is that if the op kernels are now implemented and registered outside the core, those API defs cannot reside in the core anymore. Having them in the language bindings would mean that the owners of a plugin must go in each client library (Python, Java, Go...) to add the definition of their op kernels in there as well, which does not sounds right neither.

So then, maybe what would be best is if the API defs are registered by the plugins using similar mechanisms than the ops (e.g. via `TF_NewOpDefinitionBuilder`, or via a new `TF_NewApiDefinitionBuilder` method?), and we should let diligent plugin maintainers the ability to provide a API def per language.

What do you think?

annarev:
Good point. We definitely need a way to register an ApiDef for an op.

theadactyl:
@sjamesr what's the status of this revision?

seanpmorgan:
@sjamesr @gunan @annarev This is the most important change we are looking forward to for TF-Addons as we get a ~few issues per week on ABI issues. Would it be possible to get a status update on this & expected timeline. It would be very beneficial for us to know how much work we should put into workarounds given that this is on the horizon.

byronyi:
What’s the current ABI in TF core wheels?

bhack:
`Comment period is open until 2019-08-28`. /cc @ematejska

annarev:
@theadactyl , @seanpmorgan , @byronyi  Really sorry for not providing any updates earlier. We really haven't had much work on this after initial Kernel API implementation by @sjamesr : https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/kernels.h . Since then, James moved to work on another project and I was focusing on other tasks in TensorFlow.

It seems best to merge this RFC since this doc has been pending for a while and initial implementation was submitted. I removed the device RFC part, since it wasn't finalized. Also, I will keep in mind that we need to support ApiDefs and look into a separate RFC for it when I get time.

For now, this RFC as it stands can be merged.

byronyi:
@annarev Thanks for rolling things forward. 

Btw, would you mind to share with the community the current status of the "Voltron" project?

annarev:
I think @mihaimaruseac completed the filesystem part (https://github.com/tensorflow/community/blob/master/rfcs/20190506-filesystem-plugin-modular-tensorflow.md). Unfortunately, we got caught up in other work and didn't get to spend as much time working on other tasks listed in the doc.
+ @gunan Adding Gunhan in case he has more details

mihaimaruseac:
Only the POSIX part. Windows should come by end of this quarter and GCS and S3 are part of a GSoC project. If there are no more delays (like there have been in the past), by TF 2.4 we should be able to convert filesystems to the Voltron world

PR #136, opened 2019-08-19
RFC: TFX Iterative Notebook Proposal

krazyhaas:
**Comment period is open through 2019-09-02**

# TFX Iterative Notebook Proposal

Status        | Accepted
:------------ | :-------
| **RFC #** | [136](https://github.com/tensorflow/community/pull/136) |
**Author(s)** | Charles Chen (<a href="mailto:ccy@google.com">ccy@google.com</a>), Joe Lee (<a href="mailto:joeyounglee@google.com">joeyounglee@google.com</a>), Kenny Song (<a href="mailto:kennysong@google.com">kennysong@google.com</a>), Kevin Haas (<a href="mailto:khaas@google.com">khaas@google.com</a>), Pushkar Joshi (<a href="mailto:pushkarj@google.com">pushkarj@google.com</a>)
**Updated**   | 2019-08-15

## Objective

We want to build a notebook user experience for modeling / iterative development using TFX Components. This will provide a fast, familiar environment for developing model and pipeline code with standard TensorFlow + TFX utilities, plus automatic notebook → pipeline export:

*   Imperative,     [define-by-run](https://ai.googleblog.com/2017/10/eager-execution-imperative-define-by.html), cell-by-cell workflow
    *   Start directly from Notebook/Colab – no running pipeline needed
    *   Run TFX components as you need them, in separate cells
    *   No explicit DAG definitions or continuous execution
*   Simple Python API per TFX component
    *   ExampleGen, StatsGen, SchemaGen, Transform, Trainer, Evaluator
    *   100% TFX compatible for automatic notebook → pipeline export
*   Analyze artifacts natively in Notebook/Colab
    *   Built-in TensorBoard, Facets, TFMA visualizations
    *   Dataset, stats, eval metrics available in notebook for custom analysis
*   Zero-setup, interactive onboarding tool for new TFX users on
    [tensorflow.org](http://tensorflow.org)


rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+runner which takes in an entire TFX pipeline, an instance of this class allows
+interactive execution of individual components. Here, a user would construct
+components with appropriate parameters and execution properties, and the
+`InteractiveContext.run(component)` method would execute that component, thereby
+materializing any output artifacts of that component.
+
+An advantage of this style is that it does not bifurcate the TFX pipeline runner
+concept into "pipeline runners" and "component runners", and it is very clear
+that this API is only meant for interactive usage (as opposed to the two
+alternatives below). A disadvantage is that we may not want to introduce
```
karmel: What sort of backward compatibility guarantees are you intending to offer on all of these APIs?

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+configuration that users can directly deploy as a scalable TFX pipeline. There
+is no additional modification required.
+
+## Target Users
+
+We target users who want to manually iterate on their models & components, and
+prefer a notebook environment for the benefits outlined above. This is a wide
+range of potential users, and from our user research, spans software engineers
+and data scientists within and outside of Google.
+
```
karmel: It might be useful to think about these users more-- a lot of the below seems to be a replication of the existing APIs for use with colab, which implies the primary users will be going between the two systems/migrating from the existing. Is that the case, or are you imagining that the users of this will be entirely new?

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+This proposal proposes a set of primitives that match concepts in the current
+TFX SDK. We propose to follow the current TFX style of having components
+explicitly take input channels (i.e. streams of artifacts of a specific type)
+and produce output channels (of another specific type). This could look like
+this:
+
+```
+# Here, with an input_base as an execution parameter with a given
+# file path.
+example_gen = CsvExampleGen(input_base=examples)
```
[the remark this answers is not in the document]
karmel: I don't quite follow-- what @alextp is describing seems to more closely align with the description above insofar as a component takes in input and produces an output-- the current example seems to assume the components themselves will have state/methods that need to be accessed separate from the output artifacts. Is that the case?

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+# Next, we use the 'examples' named output of ExampleGen as the
+# input to StatisticsGen.
+statistics_gen = StatisticsGen(input_data=example_gen.outputs['examples'])
+
+# We then similarly use the statsgen output in SchemaGen.
+infer_schema = SchemaGen(stats=statistics_gen.outputs['output'])
+
+# Next, we do example validation.
+validate_stats = ExampleValidator(
+    stats=statistics_gen.outputs['output'],
```
karmel: Needing to call .outputs['output'] seems unfortunate-- is there a convenience method or structuring that will make this less redundant for the default use-case?

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+</table>
+
+<table>
+  <tr>
+   <td><strong>Input[2]:</strong>
+<p>
+<code><em># Next, StatisticsGen with a run / read.</em></code>
+<code>statistics_gen = StatisticsGen(input_data=example_gen.outputs['examples'])</code>
+<p>
+<code>context.run(statistics_gen).outputs['output']</code>
```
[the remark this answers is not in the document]
karmel: If these are pre-determined, can't we provide a convenience that unwraps these for the user?

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+<code>context.show(example_gen.outputs['examples'])</code>
+   </td>
+  </tr>
+  <tr>
+   <td><strong>Output[1]:</strong>
+<p>
+_notebook visualization indicating we have N examples at some temp path_
+   </td>
+  </tr>
+</table>
```
karmel: Nit: I don't know what happened to the styling here, but it makes this section very hard to read. Consider making all of this a code block and separating the "cells" with comments or something?

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+  </tr>
+  <tr>
+   <td><strong>Output[3]:</strong>
+<p>
+_ExecutionResult object for ExampleValidator_
+   </td>
+  </tr>
+</table>
+
+Note that the user may have forgotten to run InteractiveContext.run() on
```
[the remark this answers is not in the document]
karmel: I had the same question as Alex, and the answer has been given in the same way a couple times now, which makes me think: maybe the problem here is that we are trying to "translate" the existing pipeline runner to colab, and Alex and I, as uninitiated users, are confused as to why this isn't more natively colab-like. Specifically, by its nature, a DAG is hard to handle in colab, since the user's expectation is that cells can be run independently to some extent. Plus, colab tends to assume everything is transient, whereas the pipeline relies on artifacts.

Which brings me back to the question at the top-- is the target user an existing TFX user who wants flexibility in notebooks, or is this designed to capture new users who want ML pipelines from notebooks? If the latter, we may want to consider a more notebook-first paradigm, rather than retrofitting the existing pipeline to colab. If, however, the expectation is that the primary use-case will still be the normal pipeline, it's fair to assume the user will end up being familiar with the DAG and prefer uniformity here.

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+upstream components in the dependency graph. Instead of implicitly running these
+upstream components, we remind the user to explicitly run upstream notebook
+cells (with a readable error message). We think this explicit mode of component
+execution is more notebook-friendly, and is easy to use with common notebook
+actions such as “Run All”, “Run Cells Before”, and “Run Cells After”.
```
karmel: Does this imply that the user will have to mentally track the DAG and execute each step in order in the notebook?

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+
+### Export to a selected orchestration engine (v0)
+
+#### Filter out InteractiveContext() objects
+
+##### Option 1: Replace InteractiveContext instances with dummy versions.
+
+1.  Search for possible import alias, e.g. `from
+    tfx.orchestration.interactive.interactive_context import InteractiveContext
```
karmel: Who is doing this search? Can you clarify the context of the process being described here? Does this imply that there would be some process by which we do this translation for users? Seems fragile. We do a lot of Python import magic to make TF exports work, and I wouldn't recommend doing that in arbitrary user code.

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+*   DummyInteractiveContext is now present/clutters the production pipeline code
+    (it's a no-op so mainly affects readability, not execution).
+*   Down the line, converting back to a notebook (replacing
+    DummyInteractiveContext with InteractiveContext) could be fragile.
+
+##### Option 2: Ensure InteractiveContext only runs in notebook context. **[recommended]**
+
+*   If InteractiveContext is run outside of a notebook context, just log a
+    warning and return.
+*   Bi-directional import to notebook from pipeline would "just work".
```
karmel: How is this an export at all? Doesn't this just mean the user has to rewrite the pipeline?

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+...
+%skip_for_export sess.run(example_gen)
+```
+
+Example cell magic:
+
+```
+%%skip_for_export
+# Cell contains scratch work that doesn't need to be exported.
+...
```
karmel: What is supposed to be exported then? The use case being described here is not clear to me.

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+    *   What if user makes a helper function and passes in a context variable?
+        (not likely, but possible)
+
+Note each of these options only filters the InteractiveContext usage in the
+exported python script, and does not prevent the user from adding it back
+afterwards.
+
+#### Export notebook contents to pipeline
+
+1.  Present the user with a Beam pipeline export template cell. Airflow/Kubeflow
```
karmel: What do you mean by "Present the user with..."-- is that in documentation? How is it presented?

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+            but this could make pipeline export more cumbersome for users who
+            have not organized their notebook in such a way. We could also
+            potentially update the notebook example to push users into a
+            particular notebook organization.
+1.  When the user runs the cell, or more specifically, when
+    `context.export_to_pipeline()` is executed, export the notebook code to .py
+    file.
+    1.  It seems beneficial to keep the export_to_pipeline() line in the same
+        cell as the pipeline.Pipeline() declaration so the user can fix any
+        errors before the export happens.
```
karmel: By the nature of notebooks, this is hard to do. Why is it necessary to have these be in the same cell? Should Pipeline() just call export at the end if these are paired processes?

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+
+# Export notebook contents.
+context = InteractiveContext()
+# TODO(USER): Name of the notebook file to be used for retrieving
+# notebook contents. IPython kernels are agnostic to notebook metadata by design,
+# and it seems that existing workarounds to retrieve the notebook filename are not
+# universally robust (https://github.com/jupyter/notebook/issues/1000).
+context.export_to_pipeline(notebook_filename='taxi_pipeline_interactive.ipynb',
+                           type='kubeflow')
+```
```
karmel: What exactly is the benefit of having these templates versus non-notebook code? Can you clarify what the end to end user story is for something like an Airflow pipeline?

rfcs/20190815-tfx-notebook.md
charlesccychen: With the existing TFX API as a baseline, the component DAG configuration aspects are not modified in this proposal, so there are no additional compatibility guarantees.  The only additional API we propose to add here would be the run() functionality, for which the guarantee is that in a notebook, it will continue to provide the functionality of running a specified component.

rfcs/20190815-tfx-notebook.md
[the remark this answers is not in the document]
charlesccychen: We want to target users who want to build TFX pipelines iteratively, so the end goal is that the user is taking what they built and running it on an orchestrator (e.g. Beam, Airflow or Kubeflow) outside of the notebook.

So while we do want to capture users who want to build ML pipelines, we want to do so in a way that (1) does not diverge too much from how such pipelines would be configured in "production TFX" and (2) will expose them gradually to TFX concepts like the component DAG (starting out by running component cells individually to produce outputs, and building the DAG from components that reference previously generated outputs).

rfcs/20190815-tfx-notebook.md
charlesccychen: The proposed notebook API here centers around the TFX Component (https://www.tensorflow.org/tfx/guide#tfx_pipeline_components) which provides a set of functionality that the user may compose or extend (for example, [Kubeflow pipelines](https://www.kubeflow.org/docs/pipelines/overview/pipelines-overview/) is converging to be compatible and use the same interface).

Here, I believe that the TFX abstraction is at a higher level than the TF one, since it is concerned with composing / orchestrating / running these often-prebuilt components, whereas the TF one focuses more closely on the details of model computations.

rfcs/20190815-tfx-notebook.md
```diff
@@ -0,0 +1,563 @@
...
+   <td><strong>Input[1]:</strong>
+<p>
+<code><em># First, ExampleGen with a run / read.</em></code>
+<code>example_gen = CsvExampleGen(input_base=examples)</code>
+<p>
+<code><em># Note that the output will be of type 'ExamplesPath', for which we</em></code>
+<code><em># may have registered a notebook visualization handler.</em></code>
+<code>example_gen_result = context.run(example_gen)</code>
+<p>
+<code>example_gen.outputs['examples']</code>
```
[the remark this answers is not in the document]
charlesccychen: The goal of this proposal is to provide an iterative development experience for building TFX pipelines.  In TFX pipelines, the Component is the unit of composability / extensibility with well-defined input / output typing.

Here, the user may use builtin components like [ExampleGen](https://www.tensorflow.org/tfx/guide/examplegen), [StatisticsGen](https://www.tensorflow.org/tfx/guide/statsgen) or [Transform](https://www.tensorflow.org/tfx/guide/transform) in building this pipeline, or they may define their own.  In a parallel effort, [Kubeflow Pipelines](https://www.kubeflow.org/docs/pipelines/overview/pipelines-overview/) is converging to use a compatible API to develop components, and we expect to have a registry of packaged components available for sharing and reuse.

Some components like [Trainer](https://www.tensorflow.org/tfx/guide/trainer) take in user-defined functions to define particular behavior for their pipeline, as you referenced.

rfcs/20190815-tfx-notebook.md
charlesccychen: I think this comment is a minor point regarding keeping the following calls in the same cell for something like the following, so that they can be modified / executed as one unit.

```python
my_pipeline = pipeline.Pipeline(components=[ExampleGen, Trainer, ...])
context.export_to_pipeline(my_pipeline, 'my_notebook.ipynb', 'exported_pipeline.py')
```

The idea would be that it would be less confusing in a single cell for any Pipeline definitions to always be coupled to the export step.

rfcs/20190815-tfx-notebook.md
charlesccychen: I think the only difference here would be exposure in the tutorial notebook for this use case.

The end-to-end user story would be (1) in a notebook, a user defines, tries out and iterates on a ML pipeline using TFX components; (2) the user decides that the pipeline is in a state where they are comfortable using it in production and runs `export_to_pipeline(..., type='airflow')` to write the output to a Python pipeline script; (3) the user modifies the pipeline script as necessary (e.g. pointing it to their Airflow cluster) and runs the script to submit / trigger their final production pipeline.

karmel: It sounds like the export is a key part of this product, and the primary goal in most cases. In that case, it seems like restructuring this to make sure export is easy and flawless seems wise? Why would a user experiment here if they have to do a lot of manual work to translate for export?

kennysong:
Here are the notes & responses to open questions from the design review on 9/5.

**Who is taking notes?**
kennysong@

**Has anything significant changed since you originally sent out the design proposal?**
No significant changes

**Discussion notes**
- Users & use cases…
- It sounds like exporting is a manual process?
  - There are some fields that require user input
- What is the role of InteractiveContext (esp for export)?
  - We explored multiple options for removing/no-oping/etc the InteractiveContext after export in the RFC
- Reviewed non-interactive TFX DSL script. 
  - The motivation is to keep the configuration as similar as possible
  - In the notebook, the human is the orchestrator
  - InteractiveContext / context.run() in a notebook is analogous to the AirflowDagRunner.run() in a fully orchestrated pipeline.
- We want to preserve components, artifacts as first class citizens, and as much of the pipeline as possible so there’s no cliff between notebook and production pipeline land
- How do we get from InteractiveContext to the AirflowDagRunner.run()?
  - This is the export functionality in the RFC
  - Conceptually, we want to pull out context.run()’s after export
- Is this CsvExampleGen object serializable?
  - Clarification: Since we know what you’re config running based on the context, we can regenerate the config during export rather than copying notebook cells?
    - The downside is that we lose intermediate variable names
  - What happens to .show() and .run()
    - Outside of notebook context, these functions become no-ops
  - Can we automatically remove .show() and .run()?
    - This is brittle. You can alias names, etc and we can’t capture all of these.
  - The end experience we want is to allow the user to specify required configuration through a Colab form, click run, and then the notebook exports to pipeline.
- Components vs functions? Why not provide functions that just do the same thing? i.e. CsvExampleGen can be a function that returns an output rather than storing the result in CsvExampleGen.outputs
  - We want artifacts to be first class citizens and preserve their lineage. This requires components with well defined inputs and outputs.
  - Code is only have the story. The other half is data and artifacts.
- From this notebook, can you iterative develop taxi_utils.py?
  - You can use a Jupyter directive to edit taxi_utils.py on disk
  - The source of truth is the file on disk that is saved as you go. When you export, it references the taxi_utils.py on disk
- component.outputs is a dictionary with sometimes examples, statistics, outputs. 
- We’re not viewing the notebook as a compiler that produces an opaque thing that another opaque runner runs. The exported pipeline is user editable, experimentable, and with as few deltas as possible.
- Can we use something like Keras functional API where components have real/symbolic inputs/outputs? E.g. we have an operation like matmul that has inputs and outputs. You can also have it inside a tf.function() that has a reference to the thing that generated it. So you can run the same code with real inputs to run it interactively, or run it with symbolic inputs to package it up and run it elsewhere.
  - We would need to do more due diligence on technical and UX aspects of this
  - The part that’s missing from this is artifacts that exist on disk and provenance tracked in MLMD
    - We’re aware that it’s harder to pass in inputs on disk rather than in memory
  - Also, when exporting, the exported code would look significantly different than the interactive code. We would be defining the DAG in different way than we execute. 
  - A similar comment about: Can you use existing patterns from asynchronous programming / RPCs where things run on external machines?
- How about CsvExampleGen().call()?
  - We considered that, but this entire line would need to be rewritten on export. It’s easier to export when .run() is a separate line.

**Open Questions & Responses:**
- Is running on the artifact or the component? Do we want to change the TFX API?
  - Three options:
    - As-is [preferred, see below]
    - Make components (for example CsvExampleGen) each be a raw Python function
    - Make the output artifact(s) of a component (for example CsvExampleGen) get materialized via a future/deferred execution
  - We prefer the current API since we want to maintain parity between the TFX notebook API and the TFX pipeline API. Users should be able to use the same core API in both environments to allow easier hops between them, and the only difference is the execution model. Changing either of the below options is a breaking change for existing users or introduces a new paradigm in the notebook that's inconsistent with the pipeline declaration. We’ve evaluated this execution model with internal users who have responded that it’s clear and matches their expectations. 

- Should we invest in interfaces other than a notebook?
  - We think the TFX OSS notebook can provide immediate value for some TFX users as both (1) an onboarding tool and (2) an iterative environment for experimentation with an easy path to production -- so we should continue investing in it. 
    - TFX's product surface has several layers (and entry points), and the TFX notebook is a new entry point into the Components layer of TFX. From the notebook, it’s possible to go up a layer into Pipelines (by exporting) and down a layer into the Libraries (by calling the TFX libraries directly). See diagram from the TFX talk at KDD 2019:
o9t4NDgbDwf
    - The notebook is a flexible entry point, but not the only entry point to TFX. Users may choose to start using TFX at any layer, depending on their personal familiarity (or just preference) of environment, as well constraints of existing infrastructure. 
  - Parts of the notebook can also be reused and can inform design decisions in the future TFX frontend.
  - Another longer term project for the TFX OSS team is to create a frontend that allows monitoring, editing, and controlling TFX pipelines. This will be an IDE-like experience for interacting with TFX pipelines.

PR #138, opened 2019-08-22
Add way to allocate numbers to RFCs

ewilderj:
Start allocating each RFC a number, for ease of future reference, based on the pull request. This enables tracing the RFC back to its discussion in the GitHub PR. 
Further work: we can also go back and add these to pre-existing RFCs.


krazyhaas:
LGTM!

PR #139, opened 2019-08-22
RFC: Native SavedModel execution in Node.js

kangyizhang:
This RFC will be open for comment until Thursday, September 5th, 2019.

# Native SavedModel execution in Node.js

| **Author(s)** | kangyizhang@google.com |
| **Sponsor**   | smilkov@google.com, nsthorat@google.com, piyu@google.com, kreeger@google.com |
| **Updated**   | 2019-08-21                                           |

## Objective

This project is aiming to enable native TF SavedModel execution in Node.js environment without conversion.

### Goals

*   Implement an API to load and execute TF SavedModel in Node.js for inference, supporting both TF 1.x and 2.0.
*   Wrap the loaded SavedModel as a new subtype implementing tf.InferenceModel, so users could call default signatureDef or other signatureDefs with tag.
*   Enable the ability to inspect the SavedModel signatureDefs in Node.js with protocol buffers in JavaScript (stage 2 feature).

### DIFF
--- a/rfcs/20190821-nodejs-saved-model.md
+++ b/rfcs/20190821-nodejs-saved-model.md
@@ -1,6 +1,6 @@
 # Native SavedModel execution in Node.js
 
-| Status        | Proposed       |
+| Status        | Accepted       |
 :-------------- |:---------------------------------------------------- |
 | **Author(s)** | kangyizhang@google.com |
 | **Sponsor**   | smilkov@google.com, nsthorat@google.com, piyu@google.com |
--- a/rfcs/20190821-nodejs-saved-model.md
+++ b/rfcs/20190821-nodejs-saved-model.md
@@ -308,8 +308,8 @@ To manage and execute loaded sesion from SavedModel, a new TFSavedModel javascri
 class TFSavedModel implement InferenceModel {
  private readonly id: number;
  private deleted: boolean;
- private readonly inputOpName: string;
- private readonly outputOpName: string;
+ private readonly inputOpName: string[];
+ private readonly outputOpName: string[];
 
  constructor(id: number, backend: NodeJSKernelBackend) {}
 
--- a/rfcs/20190821-nodejs-saved-model.md
+++ b/rfcs/20190821-nodejs-saved-model.md
@@ -195,23 +195,19 @@ SavedModel
 *   MetaGraphsList
     *   MetaInfoDef
         *   …
-
-          -    graphDef
-
+    *   GraphDef
         *   …
-
-          -    ...
-
-*   Signature Definition Map:
-    *   Default SignatureDef
-        *   Input op metadata
-        *   Output op metadata
-    *   SignatureDef 1
-        *   Input op metadata
-        *   Output op metadata
-    *   SignatureDef 2
-        *   Input op metadata
-        *   Output op metadata
+    *   …
+    *   Signature Definition Map:
+        *   Default SignatureDef
+            *   Input op metadata
+            *   Output op metadata
+        *   SignatureDef 1
+            *   Input op metadata
+            *   Output op metadata
+        *   SignatureDef 2
+            *   Input op metadata
+            *   Output op metadata
 
 In TF Java TF go, TF_LoadSessionFromSavedModel and TF_SessionRun are also used to load and execute session from SavedModel. The SavedModel is parsed as protocol buffers in Java/go and input/output name is extracted from the default signature def.
 
--- a/rfcs/20190821-nodejs-saved-model.md
+++ b/rfcs/20190821-nodejs-saved-model.md
@@ -12,9 +12,9 @@ This project is aiming to enable native TF SavedModel execution in Node.js envir
 
 ### Goals
 
-*   Implement an API to load and execute TF SavedModel in Node.js for inference that is created in either TF 1.x or 2.0.
-*   Wrap the loaded SavedModel as a new subtype implementing tf.InferenceModel, so users could call default signatureDef or other signatureDefs with tag
-*   Enable the ability to inspect the SavedModel signatureDefs in Node.js with protocol buffers in JavaScript. This is a stage 2 implementation detail and will not be discussed deeply in this doc.
+*   Implement an API to load and execute TF SavedModel in Node.js for inference, supporting both TF 1.x and 2.0.
+*   Wrap the loaded SavedModel as a new subtype implementing tf.InferenceModel, so users could call default signatureDef or other signatureDefs with tag.
+*   Enable the ability to inspect the SavedModel signatureDefs in Node.js with protocol buffers in JavaScript (stage 2 feature).
 
 ### Non-goals
 
--- a/rfcs/20190821-nodejs-saved-model.md
+++ b/rfcs/20190821-nodejs-saved-model.md
@@ -1,0 +1,405 @@
+# Native SavedModel execution in Node.js
+
+| Status        | Proposed       |
+:-------------- |:---------------------------------------------------- |
+| **Author(s)** | kangyizhang@google.com |
+| **Sponsor**   | smilkov@google.com, nsthorat@google.com, piyu@google.com, kreeger@google.com |
+| **Updated**   | 2019-08-21                                           |
+
+## Objective
+
+This project is aiming to enable native TF SavedModel execution in Node.js environment without conversion.
+
+### Goals
+
+*   Implement an API to load and execute TF SavedModel in Node.js for inference that is created in either TF 1.x or 2.0.
+*   Wrap the loaded SavedModel as a new subtype implementing tf.InferenceModel, so users could call default signatureDef or other signatureDefs with tag
+*   Enable the ability to inspect the SavedModel signatureDefs in Node.js with protocol buffers in JavaScript. This is a stage 2 implementation detail and will not be discussed deeply in this doc.
+
+### Non-goals
+
+*   Enable support for training a SavedModel
+*   Enable support for exporting a SavedModel
+
+## **Motivation**
+
+[TensorFlow.js](https://www.tensorflow.org/js/) brings TensorFlow into the JavaScript world. Currently users could use [tfjs-converter](https://github.com/tensorflow/tfjs-converter) to convert TensorFlow SavedModel and TensorFlow Hub module to JS friendly format and to run inference through TensorFlow.js. The tfjs-converter tool requires developers to install Python TensorFlow and some parameters/configurations, which we have noticed users are struggling with.
+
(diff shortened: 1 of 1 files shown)


rfcs/20190821-nodejs-saved-model.md
```diff
@@ -0,0 +1,401 @@
...
+*   Extra effort to load signatures and checkpoint
+*   Need to manage ops in graph
+*   Need to update proto files occasionally
+
+Using TF C API to load and execute SavedModel is preferred, because it requires smaller scope of work to reconstruct and execute the SavedModel and no need to actively update the code for future changes in TF SavedModel.
+
+But bringing Protocol Buffers in JS might be useful to inspect SavedModel. It brings the ability to inspect SavedModel in Node.js. Further discussion is added in a later section.
+
+
+#### Do inference through running the loaded Session
```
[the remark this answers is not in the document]
pyu10055: I think based on current available APIs, we shall evaluate two options:
1. use the session for model execution, but make sure tensors are mapped properly when they are crossing boundary between session and eager context. And a migration plan when saved model execution API is available in C.
2. implement saved model loading API in C, and construct functions for the eager context. And we need to make sure we support both V1 and V2 saved models.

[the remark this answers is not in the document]
martinwicke: What do you mean by v1 and v2 saved models? What are the differences?

rfcs/20190821-nodejs-saved-model.md
```diff
@@ -0,0 +1,401 @@
...
+
+*   SavedModel metadata and graph details are available in JS
+
+Cons:
+
+*   Extra effort to load signatures and checkpoint
+*   Need to manage ops in graph
+*   Need to update proto files occasionally
+
+Using TF C API to load and execute SavedModel is preferred, because it requires smaller scope of work to reconstruct and execute the SavedModel and no need to actively update the code for future changes in TF SavedModel.
```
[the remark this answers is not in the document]
kangyizhang: As we discussed, I will update this doc to 

1. add more details about using session for model execution, such as tensor conversion between TFE context and session, inability to execute tf.function in SavedModel
2. list the scope and things to do to load and execution SavedModel through C API in V2 style.

The original design review on this Thursday will be cancelled. Once above things are done, I will reschedule the design review.

rfcs/20190821-nodejs-saved-model.md
```diff
@@ -0,0 +1,401 @@
...
+The TFSaveModel object also has an `execute()` function to execute the inference for the input tensors and return activation values for specified output node names.
+
+
+```
+const input = tensor1d([123], 'int32');
+// Execute default signatureDef of the SavedModel
+const output = savedMode.predict([input_tensors]);
+
+// Execute multiple signatureDefs of the SavedModel
+const output = savedMode.execute({'input_op_names':input_tensors}, ['output_op_names']);
```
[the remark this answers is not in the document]
kangyizhang: This API matches current Model execution API in TensorFlow.js. This can also be implemented as signature name which matches pythong API.

kangyizhang:
Notice: this doc will be revised and the review will be rescheduled. Please keep tuned.

rfcs/20190821-nodejs-saved-model.md
```diff
@@ -0,0 +1,401 @@
...
+*   Heavy maintenance duty for tfjs-node
+*   Extra tfjs_bindings API to maintain
+*   Need to copy all proto files from TensorFlow to tfjs-node
+
+
+##### Decision
+
+Option 1, ask user to provide the input/output names, could be the temp solution for stage 1 of this project.
+
+Option 2, add function in TF C API to get input/output op names from SavedModel, is a good long term solution. It separates tfjs-node from TF C library details and keeps tfjs_bindings API simple, but it requires extra release cycle for the next TensorFlow C library.
```
tonanhngo: I would agree that adding the TF C API is a good long term solution that would benefit other language bindings, old and new.  For instance, the current Go and Java implementation for these same functions can be factored out and replaced with the new API calls, simplifying code maintenance.  Should this approach be proposed in parallel while option 1 is used as the temporary solution?

rfcs/20190821-nodejs-saved-model.md
```diff
@@ -0,0 +1,401 @@
...
+
+### User-facing code
+
+This project will provide a new API `tf.node.loadSavedModel` to load a SavedModel as a new class `TFSavedModel` in Node.js which can be used by the user to execute the SavedModel for inference.
+
+The loadSavedModel API takes a `path` param, which is the absolute path to the SavedModel directory. It returns a `TFSavedModel` object, implementing [InferenceModel](https://github.com/tensorflow/tfjs/blob/81225adc2fcf6fcf633b4119e4b89a3bf55be824/tfjs-core/src/model_types.ts#L36) class.
+
+
+```
+const savedMode = tf.node.loadSavedModel(__dirname + 'saved_model_dir');
```
tonanhngo: nitpicking:  guess you mean "savedModel" ?

rfcs/20190821-nodejs-saved-model.md
```diff
@@ -0,0 +1,401 @@
...
+const input = tensor1d([123], 'int32');
+// Execute default signatureDef of the SavedModel
+const output = savedMode.predict([input_tensors]);
+
+// Execute multiple signatureDefs of the SavedModel
+const output = savedMode.execute({'input_op_names':input_tensors}, ['output_op_names']);
+```
+
+
+The TFSavedModel object also has a `delete()` function to free the SavedModel related memory.
```
tonanhngo: Is this call considered a good practice or a requirement?

rfcs/20190821-nodejs-saved-model.md
```diff
@@ -0,0 +1,401 @@
...
+
+A [SavedModel](https://www.tensorflow.org/beta/guide/saved_model) is a directory containing serialized signatures and the states needed to run them.
+
+
+```
+assets  saved_model.pb  variables
+```
+
+
+The directory has a saved_model.pb (or saved_model.pbtxt) file containing a set of named signatures, each identifying a function. One of these signatures is keyed as “serving_default”  and used as the default function when user does not specify a signatureDef when executing the SavedModel. This saved_model.pb file contains encoded structure data that needs to be read as binary and parsed as [SavedModel](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/saved_model.proto) proto format.
```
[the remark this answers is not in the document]
kangyizhang: Yes, the current design is to throw an error when the SavedModel does not have serving_default signature.

rfcs/20190821-nodejs-saved-model.md
```diff
@@ -0,0 +1,401 @@
...
+
+
+```
+const savedMode = tf.node.loadSavedModel(__dirname + 'saved_model_dir');
+```
+
+
+The returned TFSavedModel object has a `predict()` function to execute the SavedModel serving_default signature for inference. The param of this predict() function would be a single tensor if there is single input for the model, an array of tensors or named tensor map if the model has multiple inputs.
+
+The TFSaveModel object also has an `execute()` function to execute the inference for the input tensors and return activation values for specified output node names.
```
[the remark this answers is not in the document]
kangyizhang: That makes sense. I'll update the design.

rfcs/20190821-nodejs-saved-model.md
kangyizhang: Yes that's the plan. I'm updating the doc to list all the changes need to be added in the TF C API, which should benefit all bindings.

rfcs/20190821-nodejs-saved-model.md
kangyizhang: Yes, thanks for catching this.

rfcs/20190821-nodejs-saved-model.md
kangyizhang: It's not a hard requirement, but it will help with memory management, especially for servers that keeps running.

rfcs/20190821-nodejs-saved-model.md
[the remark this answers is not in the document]
alextp: I don't understand how you know which signature to execute though

rfcs/20190821-nodejs-saved-model.md
```diff
@@ -0,0 +1,401 @@
...
+
+Thus almost all the proto files in tensorflow are required to be translated to js version. This can be achieved by clustering all the proto files in [tensorflow repo](https://github.com/tensorflow/tensorflow) together into one folder,  running the command above and moving the generated translated_pb.js files into tfjs-node repo.
```
alextp: Why is this different from just using bazel to generate js proto rules?

kangyizhang:
Hi, this doc has been update and a review has been rescheduled on October 8th.

The major changes from previous version include:

1. Clarified that this project is aiming to expand the current model API in TensorFlow.js, so that when users want to use a TF SavedModel for inference in Node.js,  they do not need to use python tool [tfjs-converter](https://github.com/tensorflow/tfjs/tree/master/tfjs-converter) to convert the model.
2. The proposed model loading API `tf.node.loadSavedModel()` will take 3 parameters: `path`, `tag`(metagraph), and `signature`. So it's clear to user that the loaded object is only used for inference with one Signature. This also aligns with current SaveModel usage workflow with tfjs-converter.
3. A new API will be added in tfjs-node to inspect a SavedModel, so that users can know the metagraph and signature information in the model. Protobuf will be added in tfjs-node to support this functionality.
4. The input/output op name will be parsed based on the provided signature, so users do not need to provide them.

Please take a look at the updated doc. Thank you.

rfcs/20190821-nodejs-saved-model.md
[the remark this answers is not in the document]
kangyizhang: Thanks for the comment. I haven't updated the doc until several minutes ago. 

Then loading a SavedModel, the `tf.node.loadSavedModel` function will take 3 parameters: `path to model`, `tag`(metagraph), and `signature`. The input/out op names will be parsed based on the signature. So when calling `model.predict()`, users only need to provide input tensors. 

Users are able to load multiple signatures by calling tf.node.loadSavedModel() multiple times. The node.js backend will keep track of loaded session and use the same session for multiple signatures if user want to use multiple signatures from the same model..

rfcs/20190821-nodejs-saved-model.md
kangyizhang: Using the [Protocol Compiler](https://github.com/bazelbuild/bazel/tree/master/third_party/protobuf/3.6.1/js#protocol-buffers---googles-data-interchange-format) should be the same as using bazel to generate js proto rules. It will only be used once to generate the saved_model_proto.js file and will not be used by users.

rfcs/20190821-nodejs-saved-model.md
kangyizhang: The latest design is to add protocol buffers in JavaScript, so 1) js developers could inspect the signatures of a SavedModel and 2) no need to add extra N-APIs to make the bindings more complicate and hard to maintain.

PR #144, opened 2019-08-28
RFC: TFX Resolver

ruoyu90:
**Review will be open through 2019-09-12**

# TFX Resolver

Status        | Proposed
:------------ | :--------------------------------------------
**Author(s)** | Ruoyu Liu (ruoyu@google.com)
**Updated**   | 2019-08-28

## Objective

This RFC proposes the design of Resolver, which serves as an optional plugin in TFX DSL to handle artifact resolution before a component execution. The following can be achieved with this design:

*   Enable artifact resolution across pipeline run boundaries
*   Make artifact resolution easy to customize directly through pipeline definition

### DIFF
--- a/rfcs/20190828-tfx-resolver.md
+++ b/rfcs/20190828-tfx-resolver.md
@@ -1,0 +1,156 @@
+# TFX Resolver
+
+Status        | Proposed
+:------------ | :--------------------------------------------
+**Author(s)** | Ruoyu Liu (ruoyu@google.com)
+**Sponsor**   | Konstantinos Katsiapis (katsiapis@google.com)
+**Updated**   | 2019-08-28
+
+## Objective
+
+This RFC proposes the design of Resolver, which serves as an optional plugin in
+TFX DSL to handle artifact resolution before a component execution. The
+following can be achieved with this design:
+
+*   Enable artifact resolution across pipeline run boundaries
+*   Make artifact resolution easy to customize directly through pipeline
+    definition
+
+## Motivation
+
+In the original design of TFX orchestration, Driver is used to prepare all
+artifacts needed for a component execution and feed the result artifacts into
+Executor for execution. The default behavior of input artifact resolution is to
+take the outputs of upstream components in the same pipeline run. Any behavior
+other than that requires a customized driver. While Driver is sufficient in
+terms of functionality, it is essentially a blackbox for TFX end users that is
+hard to reason about. Customization and maintenance are also hard since a Driver
+also contains other logic such as execution decisions making.
+
+To address the aforementioned problem, we propose to extract the artifact
+resolution part into a separate unit, named Resolver. It has the following
+attributes:
+
+*   It is an optional plugin. Users that do not need this feature do not need to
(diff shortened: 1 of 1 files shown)


rfcs/20190828-tfx-resolver.md
```diff
@@ -0,0 +1,156 @@
...
+      input_dict: Dict[Text, Channel]) -> Dict[Text, Iterable[Artifact]]:
+    raise NotImplementedError
+```
+
+The parameter `metadata_handler` passed into `resolve()` is read-only since no
+write should be allowed during artifact resolution stage. The other parameter
+input is a mapping from tags to Channels. Each Channel provides the type
+information that will be used when querying ML metadata.
+
+### DSL intergration
```
rcrowe-google: Integrate

rfcs/20190828-tfx-resolver.md
```diff
@@ -0,0 +1,156 @@
...
+
+There are two options to intergrate *Resolver* into TFX DSL:
```
rcrowe-google: There are two options to integrate *Resolver* into TFX DSL:

jinnovation:
@ruoyu90: It looks like this design could be used for any sort of Artifact transformation, e.g. taking an output URI from an upstream component and resolving it to a different, derivative URI—say, a sub-directory—before passing it to a downstream. Is this intended functionality?

ruoyu90:
@jinnovation Thanks for bringing up this point! The short answer is no. This design can be used for theoretically any MLMD query that will return a list of (already existing) artifacts. Artifact transformation will not be valid since it will produce new artifacts. A safety check will be added to check against output artifacts.

jinnovation:
@ruoyu90: Makes sense. Thanks for clarifying.

In tensorflow/tfx#495, you mention the possibility of TFX enabling primitive outputs from Components. Is consideration of possible primitive outputs outside of the scope of this RFC? If not, would users be able to use a similar Resolver-esque pattern to enable such transformation?

ruoyu90:
@jinnovation Primitive outputs is outside of the scope of this RFC. But it's closely related to the ongoing proposal for supporting conditional in DSL we're working on. The initial plan on the primitive side is that we expect to have new artifact types map to primitive types. And all these are still tracked in MLMD.

rcrowe-google:
Touching to update for CLA

ewilderj:
Please update the status of this document to "Accepted", and then I will merge it. Can you add any notes from the design review meeting suitable for public consumption to this PR as well, for the record?
Thank you!

ruoyu90:
> Please update the status of this document to "Accepted", and then I will merge it. Can you add any notes from the design review meeting suitable for public consumption to this PR as well, for the record?
> Thank you!

Done. Btw there is no major concern from design review.

theadactyl:
@ruoyu90  I believe there's still a CLA issue to resolve here. Can you ensure that all commit authors  and co-authors  have signed?

ewilderj:
Manually accepting CLA as authors are googlers, but please diagnose this for future!

PR #146, opened 2019-09-04
RFC: TFX Generic Container-based Component

ucdmkt:
**Review will be open through 2019-10-9**

# TFX Generic Container-based Component

Status        | Proposed
:------------ | :-------
**Author(s)** | Ajay Gopinathan (<a href="mailto:ajaygopinathan@google.com">ajaygopinathan@google.com</a>), Hongye Sun (<a href="mailto:hongyes@google.com">hongyes@google.com</a>), Makoto Uchida (<a href="mailto:muchida@google.com">muchida@google.com</a>)
**Updated**   | 2019-09-04

## Objective

This document proposes a design to enable users to attach an arbitrary
containerized program as a component to a pipeline authored using the TFX DSL,
in a way that inter-operates with other components.

This RFC assumes some clarification on the TFX’s use of artifacts and metadata
as explained in [this section of TFX user guide](https://www.tensorflow.org/tfx/guide#artifacts).

### DIFF
--- a/rfcs/20190904-tfx-generic-container-based-component.md
+++ b/rfcs/20190904-tfx-generic-container-based-component.md
@@ -214,9 +214,9 @@ from and to storage on behalf of the wrapped containerized application program.
 
 We propose the following syntax for wrapping user’s containers in the DSL by
 introducing `ExecutorSpec`. This syntax complements and extends the
-[ComponentSpec](https://github.com/tensorflow/tfx/blob/4cafb7a3ddd73463d70072dade71f380e86072e6/tfx/types/component_spec.py#L73)
-design previously implemented in TFX, and generalizes
-[EXECUTOR_CLASS](https://github.com/tensorflow/tfx/blob/767968262233bb8d37837e8de9e11d1065b07c7b/tfx/components/base/base_component.py#L66)
+[ComponentSpec](https://github.com/tensorflow/tfx/blob/0.14.0/tfx/types/component_spec.py#L73)
+design previously implemented in TFX, and generalize `EXECUTOR_CLASS` to
+[EXECUTOR_SPEC](https://github.com/tensorflow/tfx/blob/0.14.0/tfx/components/base/base_component.py#L63)
 attribute of components. We propose to use the same `ComponentSpec` base class
 to describe the custom container-based component’s input and output artifacts,
 and its parameters.
@@ -347,7 +347,7 @@ my_trainer_exec_spec = ExecutorContainerSpec(
 # the command line flags that my_trainer.py takes are different from what
 # TFX stock trainer takes. Nevertheless, it does produce an instance of
 # Model artifacts, which can then be consumed by downstream components.
-class MyContainerBasedTrainerSpec(ContainerBasedComponentSpec):
+class MyContainerBasedTrainerSpec(ComponentSpec):
   """ComponentSpec to drive my_trainer.py as a component."""
   # Input artifacts.
   INPUTS = {
@@ -422,40 +422,39 @@ def create_pipeline():
 # in which case, an Exception is raised at the time when the logical pipeline
 # is compiled for execution by the TfxRunner.
 # See the next section of this doc about ComponentLauncher.
-_ = KubeflowDagRunner().run(_create_pipeline())
+_ = KubeflowDagRunner().run(create_pipeline())
 
 ```
 
 ### ComponentLauncher to launch the container-based application
 
 With the introduction of `ExecutorContainerSpec` which does not specify
-`executor_class`, the default `ComponentLauncher` cannot execute the
-container-based component. Furthermore, different orchestrator (i.e. an instance
-of `TfxRunner`) may have different ways to launch the containerized application
-program.
+`executor_class`, the default implementation of `BaseComponentLauncher` cannot
+execute the container-based component. Furthermore, different orchestrator
+(i.e. an instance of `TfxRunner`) may have different ways to launch
+the containerized application program.
 
-We propose to extend the base `ComponentLauncher` to define
-orchestrator-specific ways to execute the containerized program, including the
-ways to translate input artifacts to the complete command line, by filling the
-[Jinja template](https://jinja.palletsprojects.com/en/2.10.x/) for
-`ExecutorContainerSpec.command` and `ExecutorContainerSpec.args`, and to
+We propose to extend the `BaseComponentLauncher` to define orchestrator-specific
+ways to execute the containerized program, including the ways to translate input
+artifacts to the complete command line, by filling the [Jinja template](https://jinja.palletsprojects.com/en/2.10.x/)
+for `ExecutorContainerSpec.command` and `ExecutorContainerSpec.args`, and to
 translate output from the containerized application to keep track of metadata of
 it and write back to Metadata storage.
 
-Below is one possible implementation of `ComponentLauncher` that implements a
-way to launch container-based components with `KubeflowDagRunner`, with ability
-to configure low level k8s configurations. This `KubeflowComponentLauncher`
-would use the k8s Pod API to launch the container [ref]. On top of this,
-`KubeflowDagRunner` allows to apply
-[additional k8s APIs](https://github.com/tensorflow/tfx/commit/74f9b6ab26c51ebbfb5d17826c5d5288a67dcf85#diff-6e0294d7fcca3bea10cb2c38bc7860c3),
+Below is one possible implementation of `BaseComponentLauncher` that implements
+a way to launch container-based components with `KubeflowDagRunner`, with
+ability to configure low level k8s configurations. This
+`KubeflowComponentLauncher` would use the k8s Pod API to launch the container
+through underlying Kubeflow Pipelines SDK implementation [ref](https://github.com/tensorflow/tfx/blob/0.14.0/tfx/orchestration/kubeflow/base_component.py#L118).
+On top of this, `KubeflowDagRunner` allows to apply [additional k8s APIs](https://github.com/tensorflow/tfx/blob/0.14.0/tfx/orchestration/kubeflow/kubeflow_dag_runner.py#L188),
 such as volume mount and secret management to pods.
 
 ```python
 # Demonstration of a ComponentLauncher that has the ability to launch
 # container-based component, in addition to executor-class based component,
 # with KubeflowDagRunner.
-class KubeflowComponentLauncher(ComponentLauncher):
-  """Demonstration of ComponentLauncher specific to KubeflwoRunner."""
+class KubeflowComponentLauncher(BaseComponentLauncher):
+  """Demonstration of a ComponentLauncher specific to KubeflwoRunner."""
 
   def __init__(self, ..., platform_config=...):
     # platform_config delines any Runner-specific, for example k8s-specific
@@ -535,7 +534,7 @@ component to capture it as properties of output artifacts. File I/O is
 consistent with the way how `KubeflowDagRunner` passively logs output artifact
 metadata as of `tfx` 0.13, hence natural extension to it.
 
-### Binding between CompomentSpec and ExecutorContainerSpec
+### Binding between ComponentSpec and ExecutorContainerSpec
 
 (A subclass of) `ComponentSpec` defines input and output artifact specification,
 and execution property of the component, but does not define ‘what’ to execute
@@ -563,7 +562,7 @@ This is as illustrated as the code snippet in the previous sections.
             needs to be defined in different places and kept in tightly
             consistent.
             1.  Command line flags to `my_example_gen.py`.
-            1.  Jinja template defined in `my_example_gen_exec_spec.commnad` and
+            1.  Jinja template defined in `my_example_gen_exec_spec.command` and
                 `my_example_gen_exec_spec.args`.
             1.  `INPUT`, `OUTPUT` and `PARAMETERS` in
                 `MyExampleGenComponentSpec`.
@@ -763,7 +762,7 @@ class MyContainerBasedExampleGen(BaseComponent):
 
 ## Appendix
 
-### Ppeline Compilation and Release
+### Pipeline Compilation and Release
 
 The proposed `ExecutorContainerSpec` and related modification will reside in the
 TFX repository.
@@ -794,16 +793,16 @@ to create a helper to build an image, configure a `python` command entrypoint
 from a naked Python function, and construct command line arguments under the
 hood, as a specialized subclass of it. Such helper shall eventually converge to
 the other way of implementing a
-[custom component](https://github.com/tensorflow/tfx/tree/80a80ffb0f833cf02344ac2898017cf1b73e166d/tfx/examples/custom_components)
+[custom component](https://github.com/tensorflow/tfx/tree/0.14.0/tfx/examples/custom_components)
 for TFX via a custom `Executor` class written in Python, and package it in a
 container image to release.
 
 Detailed RFC particularly on this point will follow.
 
-### Component Arhcetypes
+### Component Archetypes
 
 As of `tfx` 0.14, there are
-[10 known artifact types](https://github.com/tensorflow/tfx/blob/767968262233bb8d37837e8de9e11d1065b07c7b/tfx/types/standard_artifacts.py)
+[10 known artifact types](https://github.com/tensorflow/tfx/blob/0.14.0/tfx/types/standard_artifacts.py)
 defined and used.
 
 *   ExternalArtifact
@@ -818,7 +817,7 @@ defined and used.
 *   PushedModel
 
 Based on the above known artifact types, TFX defines the following
-[9 component archetypes](https://github.com/tensorflow/tfx/tree/767968262233bb8d37837e8de9e11d1065b07c7b/tfx/components).
+[9 component archetypes](https://github.com/tensorflow/tfx/tree/0.14.0/tfx/components).
 
 | **Component**    | **Inputs**                | **Outputs**              |
 | :--------------- | :------------------------ | :----------------------- |
@@ -844,7 +843,7 @@ know the actual business logic inside the container application.
 
 As of tfx 0.14, the schema (list of properties) of metadata for each artifact
 type is defined implicitly when it is created and used
-([example](https://github.com/tensorflow/tfx/blob/767968262233bb8d37837e8de9e11d1065b07c7b/tfx/components/transform/component.py#L99)).
+([example](https://github.com/tensorflow/tfx/blob/0.14.0/tfx/components/transform/component.py#L118)).
 
 In order for the proposed generic component-based component to utilize artifacts
 and its metadata in a standardized way, such metadata schema definition needs to
--- a/rfcs/20190904-tfx-generic-container-based-component.md
+++ b/rfcs/20190904-tfx-generic-container-based-component.md
@@ -25,9 +25,9 @@ is letting users orchestrate arbitrary containers as part of a Machine Learning
 languages other than Python (e.g. in R, Java, C++ etc), and the ability to chain
 existing containerized application programs with other pipeline steps through a
 Python DSL is valuable. As of 2019/09
-([tfx 0.14](https://github.com/tensorflow/tfx/blob/0.14.0rc1//RELEASE.md#version-0140),
+([tfx 0.14](https://github.com/tensorflow/tfx/blob/0.14.0/RELEASE.md#version-0140),
 even though TFX supports `KubeflowDagRunner` as an orchestrator, the
-[TFX DSL](https://github.com/tensorflow/tfx/tree/0.14.0rc1/tfx) does not provide
+[TFX DSL](https://github.com/tensorflow/tfx/tree/0.14.0/tfx) does not provide
 a mechanism to accomplish this.
 
 ## User Benefit
@@ -857,6 +857,6 @@ of artifacts must be compatible with the way metadata (via properties of
 artifacts) defines, which allows components to properly consume artifacts.
 
 We anticipate that
-[standard_artifacts.py](https://github.com/tensorflow/tfx/blob/0.14.0rc1/tfx/types/standard_artifacts.py)
+[standard_artifacts.py](https://github.com/tensorflow/tfx/blob/0.14.0/tfx/types/standard_artifacts.py)
 will serve as the catalog of known artifact types. We also anticipate that this
 catalog might evolve with more properties of a type, or more types themselves.
--- a/rfcs/20190904-tfx-generic-container-based-component.md
+++ b/rfcs/20190904-tfx-generic-container-based-component.md
@@ -812,7 +812,7 @@ defined and used.
 *   ExampleAnomalies
 *   TransformGraph
 *   Model
-*   ModelEvalulation
+*   ModelEvaluation
 *   ModelBlessing
 *   PushedModel
 
--- a/rfcs/20190904-tfx-generic-container-based-component.md
+++ b/rfcs/20190904-tfx-generic-container-based-component.md
@@ -764,8 +764,10 @@ class MyContainerBasedExampleGen(BaseComponent):
 
 ### Pipeline Compilation and Release
 
-The proposed `ExecutorContainerSpec` and related modification will reside in the
-TFX repository.
+The proposed `ExecutorContainerSpec` and any related extension of DSL APIs
+will reside in the TFX repository. Pending code completion, we may choose
+to place some or all of the new APIs under `experimental` namespace until
+we admit it to core APIs.  
 
 If run with `KubeflowDagRunner`, it will be executed by `run()` method to
 compile into Argo pipeline spec. As a result, there is no need to have any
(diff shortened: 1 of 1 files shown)


ucdmkt:
\cc @krazyhaas @neuromage @hongye-sun

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+class MyContainerBasedExampleGen(BaseComponent):
+  """Wraps my_example_gen.py."""
+  SPEC_CLASS = MyContainerBasedExampleGenSpec
+
+  EXECUTOR_SPEC = my_example_gen_exec_sepc
+
+  # Optionally, if custom driver behavior is desired, such as checking
+  # mtime for file updates, one can define a custom Driver class to control
+  # the behavior of my_exmaple_gen.py inside the container.
+  DRIVER_CLASS = ...
```
jessiezcc: from naming perspective, would it be better to use the following for consistency?
COMP_SPEC (specification for the component) 
EXECUTOR_SPEC (specification for the executor)
DRIVER_SPEC (specification of the driver)

Current SPEC_CLASS and EXECUTOR_SPEC gives different meanings to "SPEC", might cause confusion.

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+from a naked Python function, and construct command line arguments under the
+hood, as a specialized subclass of it. Such helper shall eventually converge to
+the other way of implementing a
+[custom component](https://github.com/tensorflow/tfx/tree/80a80ffb0f833cf02344ac2898017cf1b73e166d/tfx/examples/custom_components)
+for TFX via a custom `Executor` class written in Python, and package it in a
+container image to release.
+
+Detailed RFC particularly on this point will follow.
+
+### Component Arhcetypes
```
hughmiao: nit, Archetype?

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+defined and used.
+
+*   ExternalArtifact
+*   Examples
+*   Schema
+*   ExampleStatistics
+*   ExampleAnomalies
+*   TransformGraph
+*   Model
+*   ModelEvalulation
```
hughmiao: nit, ModelEvaluation?

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+*   ModelBlessing
+*   PushedModel
+
+Based on the above known artifact types, TFX defines the following
```
hughmiao: artifact type used in the following table needs some update, see inline comments

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+[9 component archetypes](https://github.com/tensorflow/tfx/tree/767968262233bb8d37837e8de9e11d1065b07c7b/tfx/components).
+
+| **Component**    | **Inputs**                | **Outputs**              |
+| :--------------- | :------------------------ | :----------------------- |
+| ExampleGen       | ExternalPath (optional)   | Examples                 |
```
hughmiao: ExternalArtifact?

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+long as the wrapped component adheres to one of the above the input- and output-
+archetypes, it enables drop-in replacement, while retaining interactions with
+the rest of the components in the pipeline, and also the pipeline not having to
+know the actual business logic inside the container application.
+
+### Specification of Artifacts
+
+As of tfx 0.14, the schema (list of properties) of metadata for each artifact
+type is defined implicitly when it is created and used
+([example](https://github.com/tensorflow/tfx/blob/767968262233bb8d37837e8de9e11d1065b07c7b/tfx/components/transform/component.py#L99)).
```
hughmiao: is this the updated link for tfx 0.14?

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+custom built container-based component would not be able to implement the
+interaction with other components via such artifact types, in turn the custom
+container-based component would not be able to make use of the interoperability
+and shareability with other components in a pipeline. Furthermore, the payload
+of artifacts must be compatible with the way metadata (via properties of
+artifacts) defines, which allows components to properly consume artifacts.
+
+We anticipate that
+[standard_artifacts.py](https://github.com/tensorflow/tfx/blob/0.14.0rc1/tfx/types/standard_artifacts.py)
+will serve as the catalog of known artifact types. We also anticipate that this
```
hughmiao: nit, use 0.14 link?

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+
+### Specification of Artifacts
+
+As of tfx 0.14, the schema (list of properties) of metadata for each artifact
+type is defined implicitly when it is created and used
+([example](https://github.com/tensorflow/tfx/blob/767968262233bb8d37837e8de9e11d1065b07c7b/tfx/components/transform/component.py#L99)).
+
+In order for the proposed generic component-based component to utilize artifacts
+and its metadata in a standardized way, such metadata schema definition needs to
+be made explicit, possibly as a Python class. In other words, unless the known
```
hughmiao: `possibly as a Python class`, would you please provide some example or pointers here?

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+    *   Potentially, `BaseExecutor` can extend this model to support a class
+        method `validate_component_spec()` to support user executor static
+        validation of any logic.
+*   **Cons**
+    *   `dynamic_spec_class()` style may be cumbersome.
+    *   The dynamic class cannot be shared like static class.
+
+## Appendix
+
+### Ppeline Compilation and Release
```
hughmiao: nit, Pipeline

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+## Motivation
+
+A key value proposition provided by
+[Kubeflow Pipelines (KFP)](https://www.kubeflow.org/docs/pipelines/overview/pipelines-overview/)
+is letting users orchestrate arbitrary containers as part of a Machine Learning
+(ML) pipeline. Many users already have custom ML applications written in
+languages other than Python (e.g. in R, Java, C++ etc), and the ability to chain
+existing containerized application programs with other pipeline steps through a
+Python DSL is valuable. As of 2019/09
+([tfx 0.14](https://github.com/tensorflow/tfx/blob/0.14.0rc1//RELEASE.md#version-0140),
```
hughmiao: nit, update to use the 0.14.0 link?

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+
+*   **Artifacts**: These include the input CSV file, the output CSV file with
+    training and eval examples, and the final Model file.
+*   **Executions**: These include execution steps `MyExampleGen`, with runtime
+    execution property `num_csv_columns`, and `MyTrainer`, with runtime
+    execution property `train_steps`.
+
+Notably, since `my_example_gen.py` produces *Examples*, we make the pipeline
+understand that it is *Examples*, and record it as such in ML Metadata. Doing so
+will enable downstream components, such as Trainer, to understand that it is
```
jlewi: Is schema being tracked out of band here? Why isn't schema tightly coupled to the data itself? For example, can schema be encoded in the file itself. I believe Avro files for example and lots of other file formats allow attach metadata that includes schema.

If your data is stored in a suitable database/feature store, wouldn't that store track schema? 

If schema is attached to the data itself then given a URI downstream files should know what kind of data they are consuming just by reading appropriate metadata.

Do TFRecord files not support encoding metadata and schema in the file.

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+
+  # An in-process Python Executor class, derived from TFX
+  # base_executor.Executor class.
+  executor_class = ...
+
+  def __init__(self, executor_class):
+    ...
+
+
+class ExecutorContainerSpec(ExecutorSpec):
```
jlewi: How does this class help a Kubernetes/Kubeflow/KFP user define Kubernetes resources using idiomatic python?

Here's a concrete example. Here's a K8s job to run one of our notebooks to train an xgboost model
https://github.com/kubeflow/examples/blob/master/xgboost_synthetic/testing/job.yaml

So imagine if we wanted to drop that job into a pipeline.

Its a 51 line YAML file. The image and command just cover 2 of those 51 lines.

So to specify the rest of the spec; init containers, environment variables, volume mounts etc...
I assume I would end up just having to put a raw yaml/json text string into the platform_config.

So the DSL/Client library isn't really helping me construct K8s objects by writing idiomatic python.

Furthermore, I'm guessing all those Kubernetes features (init containers, volumes, etc...) would break the ability to run anywhere but Kubernetes (and probably even the K8s cluster where those secrets are installed).

So TFX's support for multiple runners isn't really helping me. And in the meantime, since TFX doesn't want to treat K8s resources as first class entitites because it wants to be runner agnostic. I'm not sure how its ultimately going to be able to help me write idiomatic python code to orchestrate K8s resources.

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+  """Execution specification of the container-based component."""
```
jlewi: As a Kubernetes user how am I supposed to reason about and understand ExecutorContainerSpec?

What is the underlying Kubernetes resource that it maps to? Is it a Pod? Is it a Job? Is it a Tekton Task?

As a Kubernetes user my default for batch jobs is a K8s job?
https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/

That gives me a bunch of base line features for batch jobs such as backoff and retry limits. There are also TTL features that allow me to control how long jobs persist in the API Server.

Are these features present?

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+ways to translate input artifacts to the complete command line, by filling the
+[Jinja template](https://jinja.palletsprojects.com/en/2.10.x/) for
+`ExecutorContainerSpec.command` and `ExecutorContainerSpec.args`, and to
+translate output from the containerized application to keep track of metadata of
+it and write back to Metadata storage.
+
+Below is one possible implementation of `ComponentLauncher` that implements a
+way to launch container-based components with `KubeflowDagRunner`, with ability
+to configure low level k8s configurations. This `KubeflowComponentLauncher`
+would use the k8s Pod API to launch the container [ref]. On top of this,
```
jlewi: Why Pod API? As a Kubernetes user how am I supposed to reason about what's happening? Pods are not resilient. So there needs to be another layer like a K8s Job that restarts failed pods.

Where are the semantics defined? Are  you defining new semantics or using well defined semantics; e.g. an appropriate Kubernetes resource.

rfcs/20190904-tfx-generic-container-based-component.md
neuromage: The goal of TFX is not to enable one to write Python code to orchestrate k8s resources. It's to author a logical machine-learning pipeline that is portable. In order to achieve this goal, we need to separate the specification for a logical pipeline from the runtime configuration. When running TFX on Kubeflow Pipelines, one way to achieve this is to supply that configuration to `KubeflowDagRunner`, which is what currently compiles the TFX logical pipeline to the Argo spec that can be deployed on KFP. That separation may look something like this (made up syntax):

```
pipeline = .... # construct logical pipeline
platform_config = {
  'MyTrainer': { 'use_gpu': True, 'memory': '1GB' },   
  ....
}
KubeflowDagRunner.run(pipeline, platform_config=platform_config)  # produce YAML spec
```

rfcs/20190904-tfx-generic-container-based-component.md
neuromage: This specification isn't meant to be Kubernetes specific. It is intended to capture the notion that a step in the pipeline may consist of running a container. This container may need to be executed on Kubernetes, or locally, or in some managed orchestration service in the Cloud. The pipeline user shouldn't need to worry about backoff and retry limits when authoring the component. Instead, we intend for these to be specified as part of the platform configuration. This latter part is still being designed, and isn't really being addressed in this RFC.

jlewi:
Thanks @neuromage for putting this together.

Who is this proposal targeting? Is the goal to target users who are already using TFX and need an escape hatch to run arbitrary containers?

Or is the goal to make it easier for  Kubernetes/Kubeflow/KFP users to start using TFX to do orchestration?

Coming from Kubeflow/Kubernetes its not clear to me how this provides a good on ramp to orchestrating my jobs.

From a Kubernetes perspective the nature flow is

1. Write some python code
1. Build a docker container
1. Define a K8s job 
1. Combine K8s jobs into a workflow.

Argo provides a natural way to do this
https://argoproj.github.io/docs/argo/examples/readme.html#kubernetes-resources

I can just inline my K8s Jobs in an Argo workflow and organize them into a DAG.

The problem this encounters is that writing YAML becomes cumbersome for sufficiently complex workflows. 

I thought one of the key goals of the KFP DSL was to make it easier to define workflows using idiomatic python.

Looking at the proposal I don't see how that problem is being addressed

* As my inline comment shows, I think most of my config (secrets, environment variables, volumes, init containers, etc...) ends up being shoved as raw YAML/JSON into platform config.

   * So its still left to me as an exercise to figure out how to construct those objects using python
   * At which point if I'm writing python to define K8s objects, why wouldn't I just go one step further and create the Argo workflow?

* As a Kubernetes user, the proposal is introducing new concepts/resources e.g. ExecutorSpec that don't map onto Kubernetes users. So before I can use this I  need to reverse engineer how these concepts mapt to K8s resources

     * For example, in K8s using a Job is the obvious way to achieve run until completion semantics and control basic behavior like number of retries

     * Since the proposal isn't using K8s jobs I need to reverse engineer how I would set this functionality using the TFX DSL

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: This is an illustration of how one might design (a customized equivalent of) ExampleGen. 

It is totally an option so that the customized (containerized) ExampleGen to also output `Schema` artifact in addition to `Examples` artifact, specialized for particular physical format that has schema information embedded, and the downstream component to consume the produced `Schema` artifact  in the exactly the same way as it would be produced by the stock SchemaGen, so that SchemaGen is no longer necessary.

The proposed DSL extension opens up a door to make such development possible.

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: This will be no different than how `KubeflowDagRunner` launches the default container in k8s ([code](https://github.com/tensorflow/tfx/blob/0.14.0/tfx/orchestration/kubeflow/base_component.py#L118)).  The actual invocation is further abstracted inside KFP sdk (`kfp.dsl.ContainerOp`), which is an implementation of `KubeflowDagRunner` .

The `KubeflowDagRunner` also implements an API to apply KFP-specific modification on the ContainerOp ([code](https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/kubeflow/kubeflow_dag_runner.py#L135)), but is abstracted in the `KubeflowDagRunner` so that logical pipeline construction with the DSL (the `create_pipeline()`) does not see it.

ucdmkt:
Thank you @jlewi for comments.

The goal of this RFC is to primarily target those who would like to build a TFX pipeline using the DSL that TFX defines, but may want to use different implementations of components that TFX provides off the shelf ([code](https://github.com/tensorflow/tfx/tree/master/tfx/components)), which may be or may not be written in Python, without losing utility of input- and output- metadata management, and orchestration based on metadata.

This happens to address a complementary problem; provide the utility of built-in metadata management inside the DSL, for any arbitrary containerized applications orchestrated using `TfxRunner`, one implementation of which could be `KubeflowDagRunner`. If the application was a plain Python function, there is already an API to define a custom component ([doc](https://www.tensorflow.org/tfx/guide/custom_component)). This RFC further expands this capability to arbitrary containerized application. How exactly to build the said containerized application, is not in the scope of RFC.

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+
+Given the two (containerized) applications above, our goal is to chain them into
+a 2-step pipeline which will automatically track metadata related to:
+
+*   **Artifacts**: These include the input CSV file, the output CSV file with
+    training and eval examples, and the final Model file.
+*   **Executions**: These include execution steps `MyExampleGen`, with runtime
+    execution property `num_csv_columns`, and `MyTrainer`, with runtime
+    execution property `train_steps`.
+
```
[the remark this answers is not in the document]
ucdmkt: We don't have mlmd:Contexts in the scope of this RFC. Runtime parameterization of pipeline execution will be covered in a subsequent development.

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: @ruoyu90 and @charlesccychen to comment on this. 

I believe there was a legitimate reason why ComponentSpec is a class, but EXECUTOR_SPEC and DRIVER_SPEC are instances.

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: Done

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: Done

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+    behavior.
+
+    *   *Example*: Pusher can choose to proceed or halt pushing depending on
+        output artifact from ModelValidator.
+
+*   **Ability to share and reuse a wrapped component as drop-in replacement with
+    another**: As a result of the artifact centric, strongly typed input- and
+    output- signatures, it enables robust sharing and drop-in replacement of
+    components, so long as signatures (the list of input- and output- artifact
+    *types*) are the same.
```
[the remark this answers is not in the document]
ucdmkt: It is expected that executor (the containerized application) would adjust behavior according according to payload, which is expected to be part of the property of the artifact, and raise runtime error if the executor cannot understand the particular payload.

https://www.tensorflow.org/tfx/guide#artifacts has clarification on this point. 

Also added clarification to this section and appendix.

It is indeed an option to extend the proposed `ExecutorContainerSpec` to also express expected properties, and perform validation check at pipeline's compilation time as opposed to runtime. In fact, this applies not only container-based component's executor, but any executor (for example, any component could/should adjust its behavior based on the value of `split` property).

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: Done

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: Done

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: Added pointer to current Artifact base class implementation, which prescribes the default set of properties.

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: Done.

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: Updated

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: Done

rfcs/20190904-tfx-generic-container-based-component.md
ucdmkt: Done

rfcs/20190904-tfx-generic-container-based-component.md
```diff
@@ -0,0 +1,863 @@
...
+
+## Design Proposal
+
+As stated previously, custom containers may be written in arbitrary languages.
+The input interface to containers is restricted to the command-line used to
+invoke the application, while the output interface is through file I/O (either
+STDOUT, STDOUT, or container-local files).
+
+Since container applications may not have access to TFX libraries, they are not
+able to (or don’t even wish to) serialize/de-serialize
```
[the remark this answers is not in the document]
ucdmkt: Do you mean a case where some upstream component would select a particular instance of `HParams`, and the downstream component to consume it as execution property?

If so, because we model any output from component as an artifact, the current way of thinking is to define an `HParams` artifact as the output from the upstream component, and materialize it as an artifact, and have the downstream component consume it as one of the input artifacts. Doing so has side benefit to materialize the selected `HParams` in Metadata store.

Having said that, we are open to suggestion if there would be an alternative.

\cc @neuromage

jlewi:
Thanks @ucdmkt 

@ucdmkt @jessiezcc @neuromage 

Is this proposal related to the discussions in the KFP community meetings about a new version of the KFP DSL?

rfcs/20190904-tfx-generic-container-based-component.md
charlesccychen: The ComponentSpec defines the input / output properties of a component, and an instance of that class can also be created to hold concrete values of those properties.  So the ComponentSpec class defines the execution schema while instances hold concrete instantiations.  Given this, it makes sense to have concrete ComponentSpec classes for each type of component.

rfcs/20190904-tfx-generic-container-based-component.md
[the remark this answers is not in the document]
hughmiao: For hparams, that makes sense. An alternative here could be add mlmd::Execution as an output from component, which includes the status (e.g., success/fail/aborted/cancel ..) of the execution and extra/arbitrary run time properties of the execution, the orchestrator runtime reads this Execution, and update its record in mlmd too.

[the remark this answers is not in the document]
neuromage: @hughmiao this is a nice idea. Using `mlmd::Execution` seems like a nice structured approach to parsing returned data. As for the mechanism, I'm still leaning towards file-based approach, which scales to different executors.

neuromage:
> Thanks @ucdmkt
> 
> @ucdmkt @jessiezcc @neuromage
> 
> Is this proposal related to the discussions in the KFP community meetings about a new version of the KFP DSL?

@jlewi yes, it is. We expect this DSL to be the primary way for authoring pipelines at some point down the road, once we achieve some level of feature parity.

theadactyl:
Notes from the design review meeting on Oct. 1, 2019: (thanks @jsimsa)

Q (Erik, Lyft): Mounting a tmp volume and writing the files to temp volume. Does it involve a size limit?
A: I don't think it is a problem. This is to output metadata, which should not be large (URI that points to data).

Q (Jiri, Google): How do you expect to interoperate?
A: Adhere to the ontology.

Q (Jiri, Google): How do you encode file format in the type?
A: Things are connectable with each other. Properties two things: container where the payloads are store (TFRecord), payload

Q (Jiri, Google): How can you validate artifact?
A: Note that this custom components are about doing something fully custom. If the checks are in the properties, we can do checks ahead of time.
 
Q (Aleks, Google): We have a type called Model which can be produced by PyTorch, TensorFlow, … If we produce TF Model
A: Whether we can compose across components. TF universe use same components where same payloads and containers.
A (Gus, Google): Two major solutions: 1) subtyping -- some contravariant types, 2) through properties.
A (Gus, Google): Even in leaves we might have problems. What we need to expose is an API - give me as Dataset, give me as PCollection.

Q (Aleks, Google): We already can create custom components even without this proposal. This proposal is to make it simpler. Do we have enough examples to guide what the abstractions should be?
A (design owners): In Kubeflow, you can chain arbitrary components, while it TFX it is not possible to do that.

Q: (Aleks, Google): Removing the boilerplate does not necessarily simplify the user experience. I don't want to go into more APIs before we understand the use case.
A (design owners): Some of the boilerplate is translating the metadata into the execution. It is hard to ask the user to do that. 
A (Gus, Google): It comes down to how much battle tested design / implementation experience we have about this. If we have enough, it makes sense to go forth with this. From TFX perspective we do not have a lot of experience with containers. The battle-tested bit would need to come from KFP.
Q (Adam): I have experience containerizing TensorFlow and I can say it is non-trivial to do everything in a single container with and I do not see anything here that would allow me to specify initialization, etc.
Q (Zhitao, Google): Are enit properties are supported in KFP?
A (design owners): The platform config alluded to in the design proposal is how one would do that.
A (Zhitao, Google): Given that it might make sense to labeling the design as experimental.

Q (Gus, Google): Adam would labeling the proposal as experimental address your concern?
A (Adam): Once you enable people to create containers, there is a lot of gotchas with using Docker / Kubernetes

ucdmkt:
@theadactyl 

Thank you everyone for public review.  I addressed two major feedback from the design review.

theadactyl:
Manually checking CLA was signed by @ucdmkt, who used an incorrect email on one commit message.

PR #155, opened 2019-09-20
Addons: List all files which are related to tf version

facaiy:
Those are the files which must freeze tf version before release. cc @seanpmorgan


facaiy:
cc @seanpmorgan

PR #162, opened 2019-10-03
RFC: TFX Standardized Inputs

brills:
**Review period open through 2019-10-17**

Status        | Proposed
:------------ | :------------------------------------------------------------
**Author(s)** | Zhuo Peng (zhuo@google.com), Kester Tong (kestert@google.com)
**Updated**   | 2019-10-03

## Objective

*   To define a common in-memory data representation that:
    *   is powerful enough to encode the following logical training data format:
        flat ([`tf.Example`](https://github.com/tensorflow/tensorflow/blob/abfba15cd9734cec7ecd3d0661b146fc251c842d/tensorflow/core/example/example.proto#L88)), sequence
        ([`tf.SequenceExample`](https://github.com/tensorflow/tensorflow/blob/abfba15cd9734cec7ecd3d0661b146fc251c842d/tensorflow/core/example/example.proto#L298)) or structured data (e.g. [Protocol Buffers](https://developers.google.com/protocol-buffers) or [Apache Avro](https://avro.apache.org/)).
    *   all TFX components can understand and can support their own unique use
        cases with.
*   To define an I/O abstraction layer that produces the above in-memory
    representation from supported physical storage formats, while hiding TFX’s
    choice of such storage formats from TFX users.
*   To define a bridge from the above in-memory representation to TF feedables
    (i.e. Tensors and certain [CompositeTensors](https://github.com/tensorflow/tensorflow/blob/abfba15cd9734cec7ecd3d0661b146fc251c842d/tensorflow/python/framework/composite_tensor.py#L1)).


rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+## Standardized Inputs APIs
+
+We propose a set of APIs that TFX components will call, and need to be
+implemented for each of the supported combination of {physical, logical} format.
+
+```py
+class TFXIO(object):
+  """Abstract basic class of all Standardized TFX inputs API implementations."""
+  def __init__(
+      self, pipeline_env,
```
ChenkaiKuang: There is no mention of pipeline_env elsewhere, can you briefly clarify?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+    """Returns the schema of the Arrow RecordBatch generated by BeamSource().
+
+    May raise an error if the TFMD schema was not provided at construction time.
+    """
+
+  @abc.abstractmethod
+  def TFDataset(self, ...) -> tf.data.Dataset:
+    """Returns a Dataset of TF inputs.
+
+    May raise an error if the TFMD schema was not provided at construction time.
```
ChenkaiKuang: Looks like schema is strictly required for those methods, so they must raise (as opposed to may) an error if schema is absent?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+    *   [TFMA](https://github.com/tensorflow/model-analysis) may slice the
+        dataset by certain columns in the data.
+*   it may feed the data to TensorFlow. Note that feeding TF alone may not
+    require understanding the data. For example, TFMA may feed a TF model with
+    serialized tf.Example, which may be the raw form of the data.
+
+Currently, each TFX component has its own in-memory data representation to cover
+the two use cases by a different approach:
+
+|    | TFDV | TFT  | TFMA | BulkInference |
```
ucdmkt: TensorFlow model (Trainer) can consume  any form of in-memory data format as input, which is convertd to tensors, in input_fn(), typically by tf.data.Dataset for training, in-memory for serving.

In theory, there is nothing precluding me from implementing a Custom ExampleGen that produces a CSV or JSON, and a Trainer that consumes CSV or JSON as an input, so long as my input_fn does the right thing to consume CSV or JSON.

I would like some discussion on anticipated future direction after this proposal. With introduction of TFXIO, would it move the system to the direction to tighten such looseness (of in-memory and on-disk format for TensorFlow model itself), or, would it be orthogonal so that Trainer (model) continues to define own ways to define input format as it needs.

If it is in the direction to tighten it, what is anticipated impact on custom components such as [1, 2], which can choose to define own format conversion of Examples (such as CSV, JSON), as depicted in [3].

[1]  https://www.tensorflow.org/tfx/guide/custom_component
[2]  https://github.com/tensorflow/community/pull/146
[3]  https://www.tensorflow.org/tfx/guide#artifacts

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+    TypeSpecs can be used to construct placeholders or tf.function signatures.
+    """
+
+  def ToBatchTensors(
+      self, record_batch: pyarrow.RecordBatch,
+      projections: Optional[List[TensorName]]=None
+  ) -> Dict[str, TFFeedable]:  # TFFeedable: np.ndarrays or tf.EagerTensor
+                               # (or compositions of them, i.e.
+                               # CompositeTensors).
+    """Converts a record batch to batched tensors.
```
karmel: How are these batched? Do you mean the output is a single tensor with shape [batch_size, D0 ... Di]? Or is some other data structure used for batching here?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+
+    Each will conform to the corresponding TypeSpec.
+
+    Args:
+      projections: if not None, only specified subset of tensors will be
+      converted.
```
karmel: Can you explain this? An example would help. How are subsets specified?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+
+Where `TensorAdapter` is:
+
+```py
+class TensorAdapter(object):
+
+  def __init__(
+      self,
+      tensor_representations: Dict[str, TensorRepresentation]):
+    """Initializer.
```
karmel: Can you add an example of this being created/used?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+```protobuf
+message TensorRepresentation {
+  oneof {
+    DenseTensor { … }  // column_name, dtype, shape, default_value
+    VarLenSparseTensor { … }  // column_name, dtype
+    SparseTensor { }  // dtype, value_column_name, indice_column_names
+    VarLenRaggedTensor { … } // dtype
+    RaggedTensor { } // dtype, value_column_name, row_partition_column_names, ...
+    StructuredTensor { } // column_names
+  }
```
karmel: Will new tensor-like types need to be updated manually here? It seems like it will be easy to miss types here over time.

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+(composite) Tensors that are of the same type as the corresponding
+`TensorAdapter()` would return for a `RecordBatch`. See
+[this section](#recommended-way-of-implementing-a-tfxio) about how to minimize
+the code needs to be written for a new `TFXIO` implementation.
+
+The `TFDataset()` interface will accept common knobs that a user may need to
+tweak:
+
+*   Batch size
+*   Random shuffle
```
karmel: There are many, many knobs. Does it make sense to just pass all kwargs along?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+
+# Questions and Discussion Topics
+
+## OSS build / release issues
+
+### We don’t have a strong representation in the Arrow community
+
+This has led to some issues. For example, the PyPI/Wheel packaging for pyarrow
+currently is unfunded and lacks volunteers and there is a risk of support being
+dropped, but we do rely on the pyarrow wheel as TFX is released on PyPI.
```
karmel: This seems to argue that maybe the StructuredTensor approach is preferable? Can you comment on why this isn't an issue concerning enough to warrant not using Arrow?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+    *   Some TFX components functionally do not require TF to work. For example,
+        TFDV can analyze any data set. TFMA can analyze a model while treating
+        the model as a blackbox. In both cases, the ML library that trained the
+        model is irrelevant and TF should not be assumed. And coupling with TF
+        does not only introduce a heavy dependency, but also forces the user to
+        learn about TF if they need to implement a TFXIO for their data format.
+
+    *   Operations against StructuredTensors boils down to Python + TF ops. The
+        overhead of either is much higher than just calling an Arrow C++ API
+        that does the same operation.
```
karmel: There has been interest in moving composite tensors down to C++, which would reduce this overhead. @jaingaurav  FYI

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+
+    *   The main extension point of TF is Ops which don't understand the nested
+        structures. Compared to using Arrow’s C++ APIs, implementing an Op that
+        deals with StructuredTensors will be much complicated.
```
karmel: I don't understand this point; can you elaborate? An example might help.

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+DT_VARIANT tensor that points to the parsed Arrow RecordBatch. Then we provide
+one C++ library that translates the Arrow RecordBatch to Tensors, which can also
+be reused in a TF op (as the downstream of the Dataset, or in a Python wrapper).
+
+# Alternatives Considered
+
+We’ve considered an alternative where
+[**StructuredTensor**](https://github.com/tensorflow/community/blob/master/rfcs/20190910-struct-tensor.md)
+is the unified in-memory representation, and **tf.Data** is the unified I/O
+abstraction.
```
karmel: After reading this section, I'm actually less convinced that StructuredTensors aren't the answer here than when I started. Performance and convenience overheads could be minimized if necessary, and the question of requiring TF could be resolved by providing an inverse adapter-- rather than Arrow -> TF, convert TF -> Arrow on an as-needed basis. It seems like this route would simplify some of the build problems noted and reduce the total number of conversions required. @edloper and @alextp for StructuredTensors-- what would you argue for here?

rfcs/20191017-tfx-standardized-inputs.md
wesm: The odds of wheel support being dropped from the project seems extraordinarily unlikely. 

The pyarrow wheel has been installed almost 30 million times

```
$ pypistats overall pyarrow
|    category     | percent | downloads  |
|-----------------|--------:|-----------:|
| with_mirrors    |  50.16% | 14,307,951 |
| without_mirrors |  49.84% | 14,218,655 |
| Total           |         | 28,526,606 |
```

This is on the same order of magnitude as tensorflow and pandas

```
$ pypistats overall tensorflow
|    category     | percent | downloads  |
|-----------------|--------:|-----------:|
| with_mirrors    |  50.25% | 21,595,341 |
| without_mirrors |  49.75% | 21,377,666 |
| Total           |         | 42,973,007 |

$ pypistats overall pandas
|    category     | percent |  downloads  |
|-----------------|--------:|------------:|
| with_mirrors    |  50.10% |  86,117,641 |
| without_mirrors |  49.90% |  85,783,561 |
| Total           |         | 171,901,202 |
```

pyarrow is being depended on by some other Google products, like the BigQuery Python API (cc @tswast). 

My tiny not-for-profit org (https://ursalabs.org/) has been doing most of the wheel maintenance in 2019, perhaps all that is needed is for Google to allocate a little bit of engineering time to help with the wheels now and then?

wesm: `pyarrow` seems to also be pulled in by `pip install apache_beam`

https://github.com/apache/beam/blob/master/sdks/python/setup.py#L122

rfcs/20191017-tfx-standardized-inputs.md
brills: The direction question is worth discussing in the design review.

Technically though, I could imagine a custom component adopting TFXIO to gain understanding of the data. I could also imagine that a separate set of API to allow custom ExampleGens to write data, into something opaque, so the rest of the pipeline still uses TFXIO to gain the access. That being said, it's still possible to not adopt TFXIO, and not  (or not let native TFX components to) understand the data, or establish some agreement among certain custom components.

rfcs/20191017-tfx-standardized-inputs.md
brills: The output is a dict of (Composite) Tensors with their batch dimensions being aligned.

What (Composite) Tensors are in that dict is determined by the dict of TensorRepresentations passed in through the initializer. A TensorRepresentation specifies a mapping from one or more columns in the Arrow RecordBatch to a (Composite) Tensor.

I'll update the doc to clarify.

rfcs/20191017-tfx-standardized-inputs.md
brills: New tensor-like types that TFXIO will support (i.e. the types the user can get from the dataset returned by TFXIO.TFDataset()) need to be updated manually here. The support needs to be built, so it's unlikely that we "miss" something here, but we lag behind in supporting them.

Note that this (TFX support lagging behind for various tensor-like types) has always been the case. Currently TFX (mainly TFT) only supports dense tensors (with fixed shape), var len (ragged) sparse tensors and (poorly) sparse tensors (through COO encoding). We are already missing RaggedTensors whose parsing support has been added recently.

What's worse is, what will be encoded in TensorRepresentation currently is encoded in various fields in the TFMD schema, for example:

Schema::Feature.value_count.min == Schema::Feature.value_count.max ==> Dense Tensor

Schema::Feature.value_count.min != Schema::Feature.value_count.max || Feature.value_count.min == 0 ==> var len (ragged) SparseTensor

Schema::SparseFeature specified ==> (COO) SparseTensor

VoVAllen:
Do you have any plan to support zero-copy tensor on the cuda devices with apache arrow?

rfcs/20191017-tfx-standardized-inputs.md
brills: There will likely to be many switches to control the visible behavior of the Dataset, but ideally no knobs (that tunes performance).

We can look at tf.data.experimental.make_batched_features_dataset() [1] as it is quite similar to the proposed API here (i.e. it is a one-stop shop for a Dataset of tensors, in contrast of chaining tf.data transforms)

Many kwargs passed into it are knobs that we shouldn't be exposing to our users:
- prefetch_buffer_size
- reader_num_threads
- parser_num_threads

TFX is a more controlled environment (the pipeline is aware of what type of machine it is running on), and I would argue that TFX knows better how to set those knobs than the user does.

The rest of kwargs are controlling the behavior: batch_size, num_epochs, shuffle, shuffle_buffer_size, shuffle_seed, sloppy_ordering and yes.. they will have to be passed along, because most of them are expressing record-level constraints / properties of the dataset. The most efficient way of implementing them depends on the storage format. I'm not sure we can afford converting a columnar representation of a batch of records into individual records (that's unbatch() if implemented naively).

[1] https://github.com/tensorflow/tensorflow/blob/abb92231a1a0bffcd68ce09f051851ce761cc7ca/tensorflow/python/data/experimental/ops/readers.py#L750

rfcs/20191017-tfx-standardized-inputs.md
brills: I assume you agree that OPs are the main extension point of TF (i.e. if I want to do custom computation over TF (composite) tensors, I would want to implement a custom OP).

A TF OP that takes a StructuredTensor as input is more complex to implement due to:

1) there is no corresponding C++ data structure that can describe what StructuredTensor describes.
2) there is no C++ API to re-construct a CompositeTensor from flattened list of tensorflow::Tensors (which is the raw input interface to an OP).

And in general, implementing a TF OP involves much more boilerplate than implementing a custom function that takes Arrow entities.

Moreover, TF has more serious C++ ABI compatibility issues until Modular TF.

Maybe the above can be partially addressed by C++ composite tensor. But I don't know how exactly that's going to be implemented.

rfcs/20191017-tfx-standardized-inputs.md
brills: I was evaluating the options with the current state. Technically I believe we could bridge all the gaps mentioned above if we invest enough.

I'd like to argue that at the high level, even if the gap is bridged, there's still value to adopt Arrow rather than a TF construct:

1) TFX components do more data analysis than training a model. And that part does not have to depend on TF. Choosing a TF construct may subsequently mean choosing TF as the computation platform but I'm not sure TF is (or intends to be) the best data analysis platform. We might find us in a situation where our need (for data analysis) is not well aligned with where TF is going.

2) TFX components like TFDV and TFMA can still work without assuming TF as the ML library. And I believe TFX is moving towards the direction where supporting a different ML library is possible. Requiring users that don't use TF to learning about StructuredTensor seems to be weird. I'm not sure how much would an Arrow->StructuredTensor adapter would help, because they would still need to learn about both ends in order to do it correctly.

rfcs/20191017-tfx-standardized-inputs.md
karmel: @brills - might be worth noting the above in the doc; as it is, the description sounds more dire than the apparent reality.

rfcs/20191017-tfx-standardized-inputs.md
alextp: I am strongly against having two slightly different ways of representing this type of structured data in the TF ecosystem. I also really wish these concerns were brought up during the review for structured tensor.

I also don't really understand the arguments here. A reasonable alternative to this document is "every tfx component uses tf ops to parse the data into a structuredtensor of eager tensors, and goes from there however it wants". That reuses only existing pieces, imposes no additional restrictions on the on-disk format of tfx components (since by definition it's anything that TF can read into a structured tensor, which will be anything TF can train a model over), and is fully configurable by the users. It also does not require a new arrow-like type in TF or even bidirectional conversions between structured tensors and arrow, though such conversions are relatively easy to write.

rfcs/20191017-tfx-standardized-inputs.md
alextp: No you don't need to implement an op to do computation over TF composite tensors; if the underlying tensors in the composite are stored on the CPU you can no-copy convert them into numpy arrays and no-copy convert resulting arrays back into tensors, so you can write your C++ code however you want to interact with them (in a post-eager-execution world).

You only need to implement a TF op if you want it to commute with tf.function, gradients, etc, which will likely not be the case for most TFX components.

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+which replaced all the C++ APIs with C-APIs, we will be able to eliminate the
+risk by using the same toolchain that builds Arrow.
+
+Furthermore, the Apache Arrow community is discussing about an
+[ABI-stable C-struct](https://github.com/apache/arrow/pull/5442) that describes
+Arrow Arrays. This will allow to build Apache Arrow from source and link
+statically with our code, and only talk with pyarrow through that ABI-stable
+interface.
+
+Since in Google we build everything from HEAD, using the same toolchain, there
```
alextp: But we need to support opensource too, so ABI issues are nontrivial.

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+
+### ABI compatibility with libarrow
+
+The OSS library, tfx_bsl will depend on Arrow and TensorFlow’s DSOs (dynamic
+shared objects). Because both libraries currently expose C++ APIs, there are
+always risks of incompatible ABIs as TensorFlow and Arrow are likely to be built
+using different toolchains, we cannot completely eliminate the risks.
+
+With [Modular TensorFlow](https://github.com/tensorflow/community/pull/77),
+which replaced all the C++ APIs with C-APIs, we will be able to eliminate the
```
martinwicke: Modular TensorFlow is not yet implemented. We should probably first check whether arrow and TF can coexist. The main problems are usually protobuf and absl.

wesm: We're working on removing Boost from our build toolchain (though it may be required to build some unit tests). See https://issues.apache.org/jira/browse/ARROW-6637 for information on the project

Thrift is indeed required for reading Parquet files. We'd be glad to work with you to resolve any concerns around the build toolchain.

Ultimately migrating to Bazel from CMake might be a good thing long term, but that is a very large project.

yongtang: @wesm Thanks for the information!

Thrift is needed to generate parquet_types.h and parquet_types.cpp, but that could be a one-time effort, instead of installing bison and flex (dependencies) and build parquet_types.h/parquet_types.cpp in every compile. 

In tensorflow/io I added a patch with the following as a workaround:
https://github.com/tensorflow/io/blob/master/WORKSPACE#L89-L107

I will looking into Arrow repo, and will try to create PR if I find some place where I could help.

wesm: Having the compiled files is not sufficient -- they depend on parts or the Thrift library (either libthrift.a or libthrift.so)

yongtang: @wesm Ah yes 👍 , I wasn't clear in the last comment. The thrift library will always be needed to build parquet support. Though having the complied files of parquet_types.[h|cpp] will drop the need to install `bison` and `flex` binary, which could be easier for other programs (e.g., tensorflow/io) to integrate Arrow C++.

wesm: I see. I don't think we'd be opposed to checking those in.

wesm: I opened https://issues.apache.org/jira/browse/ARROW-6821. Seems like a straightforward build toolchain improvement

brills: Currently the way tfx_bsl [1] (the proposed repo/package to host this project) and TFDV [2] build with arrow is that we build with headers shipped with the pyarrow wheel and link with libarrow.so and libarrow_python.so shipped with the same wheel. 

all we need is a way to communicate RecordBatches / Arrays with pyarrow. And there is indeed risk if we want to build with TF and pyarrow at the same time, because TF and pyarrow are not built with the same toolchain.

I believe @yongtang avoided the risk in tensorflow_io by using the IPC format between TF and pyarrow, and statically build and link with arrow's source in the arrow dataset op. However producing the IPC format AFAIK is not copy-free. I was hoping that the proposal of the ABI stable C-struct [3] can provide a safe way of passing things in-process.

[1] https://github.com/tensorflow/tfx-bsl
[2] https://github.com/tensorflow/data-validation
[3] https://github.com/apache/arrow/pull/5608

rfcs/20191017-tfx-standardized-inputs.md
brills: Removed as it's distracting

rfcs/20191017-tfx-standardized-inputs.md
brills: I did not add an example, but added several sentences below to explain where to get these parameters.

rfcs/20191017-tfx-standardized-inputs.md
brills: Revised the doc. Did it help?

rfcs/20191017-tfx-standardized-inputs.md
brills: thanks @wesm !

I removed that description about the risk of wheels being dropped.

TheNeuralBit: Yes we also use `pyarrow` in `apache_beam` (and thus Cloud Dataflow) for parquet support.

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+| Understand the data and conduct analysis | input data is encoded losslessly as RecordBatches. | the in-mem representation may be lossy. | Relies on the model’s input layer, and the format is Dict[str, np.ndarray]. | N/A |
+| Feed TF | N/A | the in-mem representation is TF feedable. | Feed “raw data” to the model. | Feed “raw data” to the model |
+
+This has created many issues:
+
+*   Users of individual components need to adapt their data (if input format not
+    already supported) to each component they want to use.
+*   Individual components rely on unenforceable assumptions on how to interpret
+    the input data consistently.
+*   The complexity of adding new logical data representations (for example,
```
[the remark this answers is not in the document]
brills: It seems "parse the data" here implies a record based I/O abstraction?

Unfortunately in TFX, parsing has to be considered together with I/O because there might not be a record serialization of the data. And the TF Ops that do I/O are tf.Dataset, so unless we bridge the gap between beam's PSource and tf.Dataset, we can't easily use TF OPs to do parsing.

The co-existence of beam's PSource  and tf.Dataset in TFX and their incompatibility is one  (but not the only one) of the factors led to this design -- if I can't use TF for I/O + parsing, why using a TF data structure?

rfcs/20191017-tfx-standardized-inputs.md
brills: But imagine that I want to write a C++ function that takes the entire structure as input; due to lack of a C++ class for StructuredTensors, my C++ code can only see individual numpy arrays and have to recover the structure from those numpy arrays.

rfcs/20191017-tfx-standardized-inputs.md
brills: I view it from a slightly different angle. This proposal is not introducing yet another way of representing structured data to the TF ecosystem (I assume the TF ecosystem is centered at TF, but TFX may not be (i.e. TF, the ML library in TFX, can potentially be replaced with other libraries). 

This proposal does not propose any new TF APIs, but only things to be built around TF, to work with TF, using existing APIs, and StructuredTensor is one of the APIs to be used.

/cc @katsiapis

brills: And again, unless we can use tf.data as a PSource in beam, tf.Dataset + StructuredTensor cannot fully replace this proposal.

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+        ([`tf.SequenceExample`](https://github.com/tensorflow/tensorflow/blob/abfba15cd9734cec7ecd3d0661b146fc251c842d/tensorflow/core/example/example.proto#L298))
+        or structured data (e.g.
+        [Protocol Buffers](https://developers.google.com/protocol-buffers) or
+        [Apache Avro](https://avro.apache.org/)).
+    *   all TFX components can understand and can support their own unique use
+        cases with.
+*   To define an I/O abstraction layer that produces the above in-memory
+    representation from supported physical storage formats, while hiding TFX’s
+    choice of such storage formats from TFX users.
+*   To define a bridge from the above in-memory representation to TF feedables
```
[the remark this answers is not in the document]
brills: Yes. I was referring to Union[CompositeTensors, EagerTensor]. (since EagerTensor is not CompositeTensor)..

rfcs/20191017-tfx-standardized-inputs.md
brills: Also, if we were to build the proposed tf.Dataset that does I/O and produces a DT_VARIANT tensor of Arrow RecordBatch, and the OP that translates a RecordBatch to Tensors, they will be statically linking against arrow source code instead of libarrow.so, because there we don't need to talk with pyarrow. (tensorflow_io is doing the same).

brills:
@VoVAllen
> Do you have any plan to support zero-copy tensor on the cuda devices with apache arrow?

This proposal is more about using Apache Arrow as the in-memory representation in TFX components that do data analysis as well as feeding TF. So in most cases, the data has to be present in the main RAM for data analysis tasks and some copying has to happen if the data is fed to a TF graph exectured with GPU kernels.

The only place zero-copy might make sense is in the TF Trainer (which is only one of handful of TFX components), and there is a dedicated TFXIO API, TFDataset() for it. I could imagine in that Dataset one could prepare RecordBatches backed by CudaBuffers, but the following conversion to Tensors has to be copy-free and compute-free in order to profit. I'm not familiar with the performance characteristics of GPUs (i.e. how expensive is copying a usually large batch)  so I'm not sure if it's worth it.

brills:
/cc
@jsimsa 
@zoyahav
@rohan100jain
@mdreves
@zhitaoli
@rose-rongliu
@emkornfield

rfcs/20191017-tfx-standardized-inputs.md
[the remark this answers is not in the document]
mrry: > And the TF Ops that do I/O are tf.Dataset, so unless we bridge the gap between beam's PSource and tf.Dataset, we can't easily use TF OPs to do parsing.

How difficult would that be? There are a lot of custom `tf.data.Dataset` adapters in tensorflow/io, so there's some precedent here. Not all of them require a record-based serialization.

I'm not familiar with Beam's extension points, but other similar systems provide a way for users to provide a "mapper" that controls iteration across an individual partition (e.g. Spark's [`RDD.mapPartitions()`](https://spark.apache.org/docs/latest/api/scala/index.html#org.apache.spark.rdd.RDD@mapPartitions[U](f:Iterator[T]=>Iterator[U],preservesPartitioning:Boolean)(implicitevidence$6:scala.reflect.ClassTag[U]):org.apache.spark.rdd.RDD[U])).

For the TFX components that *are* passing their input directly into TF, there's probably some value in optimizing the path to tensors, instead of indirecting via RecordBatch.

/cc @katsiapis @yongtang @jsimsa

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,700 @@
...
+This design proposes **a common in-memory data representation**, **a way to
+translate that into TF feedables** (np.ndarray or EagerTensors) and **a set of
+APIs** each component can use to get both.
+
+![alt_text](20191017-tfx-standardized-inputs/overview.png)
+
+## Common in-memory data representation
+
+[Apache Arrow](https://arrow.apache.org/) will be used as the common in-memory
+data representation. Beam-based TFX components will accept
```
mrry: Does this depend on any new functionality in Apache Beam? (I tried searching for Arrow/Beam integration information, and in the codebase, and couldn't find much, but perhaps I was looking in the wrong place....)

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,700 @@
...
+*   It’s OSS friendly.
+    *   Community support for more storage format I/O (e.g. Apache Parquet)
+    *   Friendly to other OSS data formats, both in-memory and on disk (e.g.
+        Pandas)
+    *   Friendly to numpy / TF: many Arrow array types share the same memory
+        layout with numpy ndarrays and certain type of TF (composite) Tensors.
+*   TF neutral.
+    *   Leaves the possibility of supporting other ML libraries open.
+
+## Translation from Arrow to TF feedables
```
mrry: Will it be possible to translate any Arrow `RecordBatch` to a TF representation? e.g. Will we need to do anything special for timestamps, unions, etc.?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,700 @@
...
+
+The key to efficient conversions is to avoid copying of data. The prerequisites
+to do so are:
+
+*   Same memory alignment
+*   Same memory layout
+
+Currently 64-byte alignment is the standard in both Tensorflow's `TensorBuffer`
+and Apache Arrow's `Buffer`. Forthermore, it can be guaranteed by implementing
+our own version of `arrow::MemoryPool` that is backed by a
```
mrry: Would this require changes in Beam to ensure that it uses our custom `MemoryPool`?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,700 @@
...
+*   the user might not know how to open those files.
+*   there might not be an efficient representation of a “record” (this is true
+    for columnar storage formats like Apache Parquet) but only an efficient
+    representation of a batch of records.
+
+Thus we propose that to most users, the TF Trainer only exposes a handle to a
+`tf.data.Dataset` of parsed (composite) Tensors.
+
+Each `TFXIO` implementation will implement a `TFDataset()` interface to return
+such a `tf.data.Dataset`. This dataset contains logically a set of batched
```
mrry: Can we instead return a `Dataset` of records, and let the user control their own batching using `Dataset.batch()`? For efficiency, tf.data could easily provide rewrites that push the batching (as well as projections!) down into the reader and map vectorization could push the batching closer to the source.

/cc @jsimsa @rachellim

rfcs/20191017-tfx-standardized-inputs.md
brills: This does not depend on Apache Arrow integration with Apache Beam (which I think is at a very early stage..)

We simply pass around PCollection[RecordBatch].

rfcs/20191017-tfx-standardized-inputs.md
brills: It's only possible for a subset of Arrow types (listed in this doc, and does not include timestampls or unions).

StructuredTensor was designed to mirror arrow's StructArray which allows nested structured data.

rfcs/20191017-tfx-standardized-inputs.md
brills: No, at least for now, because we do decoding on our own and have full control of production of Arrow entities.

rfcs/20191017-tfx-standardized-inputs.md
brills: In that case, we need something to represent a record. It could be a one-row RecordBatch. But if rewriting does not happen, that is going to be inefficient..

rfcs/20191017-tfx-standardized-inputs.md
TheNeuralBit: Deeper Arrow integration with Beam is still theoretical. I gave a talk about with some thoughts at the Beam Summit in September: https://s.apache.org/beam-arrow-vegas-2019

The tl;dr is you can do a lot with `PCollection[RecordBatch]` right now especially if you aren't doing anything with event times, windowing, etc.. but we have some more work to do (namely making schemas available across languages) before it's a real possibility.

theadactyl:
@brills what's the status of resolving comments to this proposal?

brills:
@theadactyl 
I left them unresolved because they are important questions. I'm preparing a revision to the RFC to include more background to help address those questions, but I'd still like to keep them open until after the review.

Or, is the protocol that I have to resolve all the comments before the review?

martinwicke:
All major questions should be resolved with the stakeholders involved
beforehand. We can resolve details in the review (often, naming and such),
but there shouldn't be open architecture or directional questions.

rfcs/20191017-tfx-standardized-inputs.md
brills: On the other hand, aligning with how the Generic Trainer approaches the I/O abstraction problem, which is to provide a dataset that contains "records" to allow record-based operations like shuffling, we could provide a similar interface that provide "record" based access, the only difference is that the records here are 1-row Arrow RecordBatches. Record-based operations are allowed, but at the same time, "records" are opaque to the user. In the end, they will need to use an op provided by us to rebatch those records into one big RecordBatch.

rfcs/20191017-tfx-standardized-inputs.md
brills: I updated this section to compare StructuredTensor with RecordBatch per the requirements (the "requirements" section was also newly added). PTAL.

I also added a colab to elaborate what I meant here:
https://colab.sandbox.google.com/drive/1SCQs88J4Tc6HKk2AfpYvELaJDxa4h4IQ

rfcs/20191017-tfx-standardized-inputs.md
brills: Ack. We could revisit this point once there's a design for that.

rfcs/20191017-tfx-standardized-inputs.md
brills: I added more background about TFX to the RFC (see section "Motivation" and "Requirements") to better describe the problem this proposal is trying to solve and what the workload that a TFX component / library typically bears. I also updated the "Alternatives considered" section with a head-to-head comparison. Hope this would clarify things.

rfcs/20191017-tfx-standardized-inputs.md
brills: So far the ABI issue has been manageable: tfx_bsl is already built with libarrow.so dynamically linked, and the problem can be greatly mitigated with https://github.com/apache/arrow/pull/5442, which would allow tfx_bsl build statically with arrow source and talk with pyarrow through this ABI-stable interface.

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -0,0 +1,689 @@
...
+# Standardized TFX Inputs
+
+Status        | Proposed
+:------------ | :------------------------------------------------------------
+**RFC #**     | [162](https://github.com/tensorflow/community/pull/162)
+**Author(s)** | Zhuo Peng (zhuo@google.com), Kester Tong (kestert@google.com)
+**Sponsor**   | Konstantinos Katsiapis (katsiapis@google.com)
+**Updated**   | 2019-10-03
+
+# Objective
```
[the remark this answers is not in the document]
brills: I revised this RFC to include more background about the problem it tries to solve and the typical workload for a TFX component. I hope that can help distinguish Arrow's RecordBatch from StructuredTensor.

brills:
@alextp, @karmel :

I revised the following sections of the RFC to include what we thought was missing / unclear:
- Motivation
- Requirements
- Alternatives Considered

And the internal design doc (at go/standardized-tfx-inputs) contains more numbers / evidence to support the claims in those sections. 

I hope they will help clarify things and answer better your question "why not StructuredTensor?" Please take another look.

karmel:
@brills -- it's really hard to tell what changed based on the most recent commit; is there a diff that doesn't rewrite the whole file?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -116,6 +134,47 @@ to compute approximate heavy hitters over this in-memory representation. We can
...
+# Requirements
+
+The in-memory representation should:
+
+*   Be columnar.
+
+    A columnar in-memory representation works better than a row based on under
+    typical workload of TFX libraries:
+    +   Compute statistics over a column.
+    +   Feed a batch of rows to TensorFlow.
```
karmel: Why is this easier when columnar versus row-based? This seems to be an argument for row-based.

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -116,6 +134,47 @@ to compute approximate heavy hitters over this in-memory representation. We can
...
+
+*   Be able to losslessly encode the logical data format. Specifically, it
+    should be able to distinguish a null (unpopulated) value from an empty list.
```
karmel: What is an empty list in this context? Can you give an example of when the two distinct values appear?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -600,68 +659,52 @@ be reused in a TF op (as the downstream of the Dataset, or in a Python wrapper).
...
+
+Remarks:
+
+1.  We could revise the design of StructuredTensor to include the nullibility
+    support.
+2.  Only when the backing buffers are aligned correctly. Currently both TF
+    and Apache Arrow has 64-byte alignment. And this can be enforced by
+    implementing our own Arrow MemoryPool wrapping a TF allocator.
+3.  See the comparison in [this colab notebook](https://colab.sandbox.google.com/drive/1SCQs88J4Tc6HKk2AfpYvELaJDxa4h4IQ).
+    *   It’s worth calling out that Arrow is meant to be a data analysis library
```
karmel: I am not the expert here, but @jsimsa @edloper perhaps can comment on the particular comparisons made in the notebook.

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -600,68 +659,52 @@ be reused in a TF op (as the downstream of the Dataset, or in a Python wrapper).
...
+| Efficient slicing, filtering and concatenation | Yes                | No (see remark 3) |
+| Interoperability with the rest of the world | Good through Apache Arrow | Needs adaptation (see remark 4) |
+
+Remarks:
+
+1.  We could revise the design of StructuredTensor to include the nullibility
+    support.
+2.  Only when the backing buffers are aligned correctly. Currently both TF
+    and Apache Arrow has 64-byte alignment. And this can be enforced by
+    implementing our own Arrow MemoryPool wrapping a TF allocator.
```
karmel: It would be useful to see a comparison here similar to the one in (3)

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -600,68 +659,52 @@ be reused in a TF op (as the downstream of the Dataset, or in a Python wrapper).
...
+3.  See the comparison in [this colab notebook](https://colab.sandbox.google.com/drive/1SCQs88J4Tc6HKk2AfpYvELaJDxa4h4IQ).
+    *   It’s worth calling out that Arrow is meant to be a data analysis library
+        and better data analysis support (for example, support for a “group-by”
+        clause) will be added over time.
+4.  We may gain similar interoperability by creating an Arrow to
+    StructuredTensor adapter.
+    *    Beyond the technical aspects, we believe by having all the TFX libraries
+         directly adopting a popular OSS in-memory format will send a positive
+         message that TFX is meant to work well with the rest of the world.
```
karmel: I agree that this is the most compelling argument. Another thing to consider is hardware acceleration-- eg, via NVIDIA Rapids.

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -31,57 +31,75 @@ Status        | Proposed
...
-*   TFX wants to be able to choose the optimal storage format based on user’s
-    workload, in a user-transparent manner. A unified I/O abstraction would make
-    it easier to support a new physical format in TFX, since one would not have
-    to understand every single TFX component in order to implement such support.
+Currently TFX (mostly) assumes tf.Example on TFRecord. However because TFX is a
+managed environment, it is desired that its choice of physical format of data is
+an implementation detail and is opaque to the components and the users. TFX
+would like to explore switching to a columnar physical format like Apache
+Parquet and it can be imagined that there will be a migration at some point.
+Such a migration must happen in either of the following two ways:
```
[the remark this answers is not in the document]
brills: At the end of the day, it's a migration of user pipelines, but by migration I mean both of

1. implementing support in libraries / components that need to read data.
2. changing user pipelines (trainer).

There is already such a migration happening internally (please see the internal design doc), which has followed the second approach because the first approach would require too many changes to all components, in addition to many user pipelines, if not all. And the second approach does not allow to realize all the efficiency benefits of a columnar storage format.

A similar migration will likely to happen in OSS, even if this proposal is rejected, but then we would have to make the same difficult choice.

The argument I was trying to make, is that if we have a abstract I/O layer, we wouldn't need to change any component / library, and we wouldn't need to change most user pipelines (as long as they use TFXIO.dataset()).

rfcs/20191017-tfx-standardized-inputs.md
brills: - Batching rows into batch-aligned tensors is in-efficient.
- computing (aggregating) statistics row-by-row is in-efficient.

The inefficiency came from python loops / function calls, and will be from having to read a columnar storage in a row-based manner if the storage is columnar.

rfcs/20191017-tfx-standardized-inputs.md
brills: I built another notebook to show that feeding an Arrow array (through the arrow->numpy bridge provided by Arrow) is not worse than feeding directly an aligned numpy array.
However it's a bit slower than just feeding an EagerTensor (which I would expect). But feeding an unaligned numpy array is >5x slower.

https://colab.sandbox.google.com/drive/1VKAQUPGB-bmaMJZUW5oi4JDwFt7Qirmh#scrollTo=bsHNuG_h6MD7

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -31,57 +31,75 @@ Status        | Proposed
...
+*   The complexity of adding new logical data representations scales with the
+    number of components. For example, to add SequenceExample support in TFX,
+    one may need to come up with an in-memory representation for each of the
+    components, to keep the consistency within the component.
+*   TFX library users whose data format is not supported natively by TFX would
+    have to implement the decoding logic for each of the libraries
+    they want to use.  For example, had TFX not supported the CSV format, a user
+    would have to implement
+    [one CSV decoder for TFT](https://github.com/tensorflow/transform/blob/master/tensorflow_transform/coders/csv_coder.py),
+    and [another CSV decoder for TFDV](https://github.com/tensorflow/data-validation/blob/master/tensorflow_data_validation/coders/csv_decoder.py)
```
alextp: Why is this the case? In TF you can implement a parsing op once and use it many times; why do TFX components care so much about the format the data is in?

As in, why do we care about the format of the data at rest instead of just assuming that there is a tf function or whatever that will take an at-rest format into the arrays you expect?

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -31,57 +31,75 @@ Status        | Proposed
...
+would like to explore switching to a columnar physical format like Apache
+Parquet and it can be imagined that there will be a migration at some point.
+Such a migration must happen in either of the following two ways:
 
-## TFX interoperability with the rest of the world
+*   Change every single TFX library and component that needs to read data to
+    add support for reading the new physical format (into each library’s own
+    in-memory representation)
+*   Rely on an indirection through tf.Example and give up some performance
+    because of the translation.
```
alextp: There is a third option here: make it not TFX's problem and TF's problem to parse data-at-rest into data-in-memory.

rfcs/20191017-tfx-standardized-inputs.md
alextp: Karmel, currently when you feed multi-column data into TF the most efficient way to do so is to have one tensor per column with one row in the tensor per row in the data. I think this is what this is alluding to?

That said I still don't know why the in-memory representation here cannot just be an eager StructuredTensor

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -600,68 +659,52 @@ be reused in a TF op (as the downstream of the Dataset, or in a Python wrapper).
...
+| Columnar                          | Yes                | Yes               |
+| Lossless encoding (nullity)       | Yes                | No (see remark 1) |
+| Efficient translatable to Tensors | Yes (see remark 2) | Yes               |
+| Efficient slicing, filtering and concatenation | Yes                | No (see remark 3) |
+| Interoperability with the rest of the world | Good through Apache Arrow | Needs adaptation (see remark 4) |
+
+Remarks:
+
+1.  We could revise the design of StructuredTensor to include the nullibility
+    support.
```
alextp: And this seems like a minor extension.

rfcs/20191017-tfx-standardized-inputs.md
alextp: Also @mzinkevi

alextp: I believe adding slicing, filtering, etc into StructuredTensor is very easy.

rfcs/20191017-tfx-standardized-inputs.md
```diff
@@ -600,68 +659,52 @@ be reused in a TF op (as the downstream of the Dataset, or in a Python wrapper).
...
+    *   It’s worth calling out that Arrow is meant to be a data analysis library
+        and better data analysis support (for example, support for a “group-by”
+        clause) will be added over time.
+4.  We may gain similar interoperability by creating an Arrow to
+    StructuredTensor adapter.
+    *    Beyond the technical aspects, we believe by having all the TFX libraries
+         directly adopting a popular OSS in-memory format will send a positive
+         message that TFX is meant to work well with the rest of the world.
+
+## tf.Data
```
alextp: I don't understand this section.

tf.data isn't a file a format, or an in-memory layout, it's just a way of using a TF graph to produce some tensors.

If you have something that gives you a set of numpy arrays it's very easy to wrap it as a tf.data source (using from_generator) and go from there, so you could even use your gcp thing with dynamic work rebalancing.

rfcs/20191017-tfx-standardized-inputs.md
brills: You are absolutely right, but you are making the assumption that everybody agrees on the same in-memory format and this is not true in TFX (libraries) currently, as described in this section. Please note we are covering a portfolio of TFX libraries that emerged independently, and have evolved relatively independently so far.

rfcs/20191017-tfx-standardized-inputs.md
brills: Does the "alternative considered" section help at all?

I'm not sure if Eager StructuredTensors can meet our performance requirements in certain operations (filtering / slicing / concatnating, etc)*, and is suitable for our development as we don't want to use TF to compute our custom statistics.

*Even if they can be easily implemented, given the fact that StructuredTensor is a collection of a bunch of EagerTensors (py objects) and nested metadata (still py objects), I'm not sure they can easily be implemented efficiently.

rfcs/20191017-tfx-standardized-inputs.md
brills: let's take RaggedTensor as an example -- I would assume StructuredTensor's implementation of those operations would not be faster than RaggedTensors (as it would rely on RaggedTensor's). Indeed, there is already built-in slicing and concatenating support, but tf.concat on RaggedTensors is 100x (or 200x in Eager mode) slower than pa.concat_arrays() on Arrow ListArrays [1].

[1] https://colab.sandbox.google.com/drive/1wGfmNnkVDFtq48d1RTSaEkxePPBJ0go4#scrollTo=U7eareQUdKWN

alextp: That sounds like a performance but on raggedtensor which should be fixable.

theadactyl:
Notes from the design review meeting held Nov 7, 2:00PM PST:

**Legend**
KK: Gus
KA: Karmel
Z: Zhuo
PD: Pavel Dournov

**Apache Arrow vs tf.StructuredTensor**
**Still, why not use tf.StructuredTensor as the in-memory representation?** 
**Are custom TFX components forced to use TFXIO?**
KK: Custom components can work with well-known abstractions in Arrow and benefit from the supporting infrastructure.
**ABI compatibility**
**Maintenance of pyarrow wheel**
KA: what’s the worst-case scenario?
Z: Arrow will likely hand the release over to someone else. The current process is handled through a rotation. The burden is unclear but seems low in the latest release. Most issues in the past have been with TF which was not manylinux compliant. But this has been solved now (TF is manylinux compliant).
**How can users get access to a record (row) based Dataset?**
KA: have we reached out to modellers for feedback?
Z: similar issues exist internally and the corresponding efforts mirror the proposal.
KK: there is break glass functionality but it is highly discouraged.
KA: any idea of break-glass usage internal?
Z: one known case is where the user bypasses tf.Dataset.
KK: one way to look at is that TFX APIs align with TF, which already recommends tf.Dataset. We should start with more constrained APIs to enable transparent optimizations and gradually open them up. When there is a need for flexibility then open up the API further (with the associated “cans of worms”). Perhaps follow-up RFCs can indicate what extensions are needed.
PD: At what level is the API opinionated and locked?
KK: There is no requirement to use tfx.io -- the component can go directly to the raw files but miss out on optimizations.
PD: Can CSV data be used with tfx.io?
Z: It depends on where data lives. Internal components are not likely to work with CSV as an internal storage format. CSV data can be imported with a conversion to an internal storage format.
PD: after data is ingested, can the user inject a custom component?
Z: Yes, the component can just the tfx.io API. The user can also freeze the format of the data through an option and then the component can handle the specific format.
Z: TFX will support a small set of formats. CSV will probably not be an internal representation due to inefficiency.

theadactyl:
@brills Do you think we are good to merge?

brills:
@theadactyl I think so. Thanks!

PR #165, opened 2019-10-21
Create rfc-admin-guide.md

ewilderj:
Describe what community admins do as part of the RFC process

### DIFF
--- a/governance/rfc-admin-guide.md
+++ b/governance/rfc-admin-guide.md
@@ -1,0 +1,138 @@
+# Admin guide for RFCs
+
+## Overview
+
+This document describes the process for community managers administering
+TensorFlow RFCs.
+
+|Author                        |Edd Wilder-James [@ewilderj](https://github.com/ewilderj)    |
+:------------------------------|:-----------------------------|
+|Last updated                  |2019-10-21                    |
+
+## RFC Submission Process
+
+### 1. PR is submitted to `tensorflow/community`
+
+When a PR is submitted containing an RFC proposal, check for basic
+formatting concerns.
+
+* The filename should be `rfcs/YYYYMMDD-my-rfc.md` - where YYYYMMDD is the
+  date, and hyphens connect any naming parts. No underscores. No uppercase
+  letters.
+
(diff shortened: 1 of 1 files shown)


bhack:
A meta-RFC :)
What I see Is that RCF management started quite well but currently all the deadline are a  little bit out of control.
Many RFC seems in a not so clear status.

What about a weekly or bi-weekly admin pass over the RFC PR?
Generally there are not so many parallel RFC so could be easy to have an internal bi-weekly day/clock for a status review pass.

bhack:
Just some example:
https://github.com/tensorflow/community/pull/130
https://github.com/tensorflow/community/pull/133
https://github.com/tensorflow/community/pull/77

theadactyl:
@bhack I think there's a challenge when the people doing the RFC management aren't able to meaningfully influence the resolution of RFCs that are in a tricky/unresolved state. Newer RFCs than those you noted that were more straightforward have progressed fine. We could have an internal reminder to review, but we may also need time frame guidelines around when to move RFCs to "Abandoned".

bhack:
@theadactyl I understand your point of view  but honestly without internal informations we really don't know where the Team "attention" is (a limited resource in every context).
Was It abondoned? Does It need a new public deadline? Is it stalled?

Without off the band/internal informations we really cannot classify the RFC status.

If we have a regular clock from TF management team over RFC you can codify GitHub labels and it will be a keep alive signal for community.

Expecially when RFC are open months after the reported deadline an there isn't any public signal to understand what It is happening.

This Is in topic here cause this PR Is about RFC admin and community and IMHO recrusive ping It Is not a nice solution.

I think that we need to handle a clock to pass over RFC PR and use labels or deadline extension on every pass if needed.

theadactyl:
This is really helpful feedback, will mull it over to figure out a better
solution.

bhack:
If you see also recently merged MLIR OR was accepted after implementation and months after the deadline:
https://github.com/tensorflow/community/pull/115#issuecomment-546752732

IMHO we are losing a little bit the RFC role.

PR #168, opened 2019-10-25
RFC: High level Keras API

dhruvrajan:
Please see sigs/jvm/20191023-tensorflow-keras.md for a description of the proposed high-level Keras API. Would love to get feedback and comments; hopefully we can work towards an API that is convenient for users.


dhruvrajan:
My bad, submitted incorrectly.

PR #179, opened 2019-11-23
RFC: TensorFlow Build Improvements

bmzhao:
This RFC will be open for comment until Friday, December 13th, 2019.

cc @gunan @martinwicke 

# Tensorflow Build TLC

| Status        | Proposed                         |
| :------------ | :------------------------------- |
| **Author(s)** | Brian Zhao (bmzhao@google.com)   |
| **Sponsor**   | Gunhan Gulsoy (gunan@google.com) |
| **Updated**   | 2019-11-22                       |

## Objective

Simplify Tensorflow’s build so that
1. Builds are more granular
2. Builds are easier to reason about
3. TF’s build is well positioned to immediately benefit from Bazel shared library support


rfcs/20191122-tf-build-tlc.md
```diff
@@ -0,0 +1,213 @@
+# Tensorflow Build TLC
```
ewilderj:
```suggestion
# Tensorflow Build Improvements
```

rfcs/20191122-tf-build-tlc.md
```diff
@@ -0,0 +1,213 @@
...
+| **Author(s)** | Brian Zhao (bmzhao@google.com)   |
+| **Sponsor**   | Gunhan Gulsoy (gunan@google.com) |
+| **Updated**   | 2019-11-22                       |
+
+## Objective
+
+Simplify Tensorflow’s build so that
+1. Builds are more granular
+2. Builds are easier to reason about
+3. TF’s build is well positioned to immediately benefit from [Bazel shared library support](https://docs.google.com/document/d/1qYHGcxP9BEi7t9b53r8STr9G3XVtvwy9V-EXqr54Xxg/edit#)
```
ewilderj: Please link to a public document on this feature, or remove it.

ewilderj: looks like this is the one https://docs.google.com/document/d/13nolQXvXDsJ4yjWP1FRd-WscP2-rHZcR3Zb__js6mLA/edit#heading=h.7795zogmlqtv

rfcs/20191122-tf-build-tlc.md
bmzhao: Done.

rfcs/20191122-tf-build-tlc.md
bmzhao: Done (although the original doc link was a public doc).

reuben:
It's great to see this proposal. For DeepSpeech we have similar constraints as Android, we want to keep code size down, so for a while now we've been using [very fine grained deps](https://github.com/mozilla/DeepSpeech/blob/670d24ae0323ba168b3e21009950aa20b9e8d7dc/native_client/BUILD#L126-L170), and for one case we had to [break up the TensorFlow target by hand](https://github.com/mozilla/tensorflow/commit/c350d0c68661fdb2a3e036e9cb52932ade359b20) to avoid a large binary size increase from a seemingly small dependency. That change in particular leads to several megabytes difference in our final binary size compared to simply depending on `//tensorflow/core/kernels:cwise_op`. This proposal seems to be going in the right direction and is much appreciated.

bmzhao:
> It's great to see this proposal. For DeepSpeech we have similar constraints as Android, we want to keep code size down, so for a while now we've been using [very fine grained deps](https://github.com/mozilla/DeepSpeech/blob/670d24ae0323ba168b3e21009950aa20b9e8d7dc/native_client/BUILD#L126-L170), and for one case we had to [break up the TensorFlow target by hand](https://github.com/mozilla/tensorflow/commit/c350d0c68661fdb2a3e036e9cb52932ade359b20) to avoid a large binary size increase from a seemingly small dependency. That change in particular leads to several megabytes difference in our final binary size compared to simply depending on `//tensorflow/core/kernels:cwise_op`. This proposal seems to be going in the right direction and is much appreciated.

Thank you! We hope this will help make Tensorflow's build more approachable and usable in other projects!

rfcs/20191122-tf-build-tlc.md
```diff
@@ -0,0 +1,213 @@
...
+
+This parallel build graph exists because Android would like to produce a minimal Tensorflow shared object by pulling in only a subset of Tensorflow's codebase. The reason why Android couldn't simply use the existing Tensorflow Build targets is because they are [too large](https://docs.google.com/document/d/1kg_tVB1g3c5isPaP1OXW_KZwF_hcTiKz5afLkKjiSz8/edit#heading=h.8tfaww3ddb83). Therefore, if we address [problem 1](https://docs.google.com/document/d/1kg_tVB1g3c5isPaP1OXW_KZwF_hcTiKz5afLkKjiSz8/edit#heading=h.ytms1uc3twy), we should be able to refactor the Android build to re-use these new smaller targets.
```
seanpmorgan: Were these meant to link to a google doc instead of headers in this doc? If so they are private.

bmzhao: Ah good catch! Yes these were supposed to links back to headers in the same markdown file. Just fixed, thanks!

rfcs/20191122-tf-build-tlc.md
```diff
@@ -0,0 +1,213 @@
...
+
+First, we will start adding BUILD files under subdirectories of tensorflow where they are currently missing. However, these BUILD files will create new [package boundaries](https://docs.bazel.build/versions/master/build-ref.html#packages). This will break the existing build because
+
+1.  packages cannot reference cross-package files unless they are [explicitly exported](https://docs.bazel.build/versions/master/be/functions.html#exports_files)
+
+2.  [globs](https://docs.bazel.build/versions/master/be/functions.html#glob) cannot cross package boundaries
+
+For tensorflow/core/BUILD, we will introduce BUILD files in:
+
+-   [tensorflow/core/common_runtime](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/core/common_runtime)
```
claynerobison: There are MKL components in several of these folders but default builds don't include them and won't detect errors during the refactoring process. How can we help make this smoother?

bmzhao: Hi Clayne,

That's a great point; so far the changes we've made don't touch anything MKL specific, but as we start refactoring core/common_runtime we will need to be careful not to break MKL integration.

My understanding is that Tensorflow builds use MKL by default via Eigen: https://github.com/tensorflow/tensorflow/blob/d26ba01420ed9e4e1a55f22ca0e3034fdd3d3671/tensorflow/core/kernels/BUILD#L756-L758, meaning hopefully our CI will start yelling at us if a breakage occurs; but I'll double check with @gunan

gunan: Unfortunately, most of MKL is not covered under the presubmits we are running. By default, only tensor contraction ops use MKL-DNN.

I think the majority of MKL code is under tensorflow/core/kernels, as MKL kernel implementations. The requirement there is, nothing outside tensorflow/core/kernels should define kernels as dependencies (except for binaries, or pip package).
I think MKL also has some code outside kernels, and for these, it depends on where exactly the code is. As Brian suggested, we will  start from the bottom layer of the code, and slowly build up our work.
But if you could refactor your code to minimize number of dependencies, break any include cycles and build a clear hierarchy within your code/files, that would help us greatly.

claynerobison: We can look at that. Most of the MKL code is in mkl_* files and/or surrounded by `#ifdef INTEL_MKL_DNN` code blocks so should be fairly easy to identify when restructuring BUILDs

bmzhao:
As mentioned in [SIG Build](https://docs.google.com/document/d/10_3IQ5aF-88ADJNLF0WOpb09bZ15x-sBnRSnDHNCNr8/edit#bookmark=id.7ffnhzr7fqw5), here's a link to the master spreadsheet of the current status of the refactoring so far: 
https://docs.google.com/spreadsheets/d/11QLBDRa2XHNAtA_Wby81Q9xWnDHTp6ChJrVkstbBRBE/edit?usp=sharing

bhack:
> This RFC will be open for comment until Friday, December 13th, 2019.

Can we extend the deadline? Is this closed for comment? 
Can we use a specific label after the public deadline in the case it is closed for comment and just waiting the internal approval queue?
Generally we have no visibility about this processes after the deadline.

mihaimaruseac:
Hi, @bhack 
Since this PR is still open, if there are comments please comment on the RFC using the GitHub review interface.
If there are comments against a closed PR for RFCs, you should open an issue and there will be discussion. Depending on the implementation status (which can always be checked on the other ecosystem repositories), the implementation might change based on the discussion.

bhack:
@mihaimaruseac Without any other label feedback I don't understand the status of an open PR after the public RFC deadline. Can you clarify this a little bit related to the point 5 of our RFC governance at https://github.com/tensorflow/community/blob/master/governance/TF-RFCs.md?

theadactyl:
@ematejska can you check on status of this RFC and make sure it's in the right bucket? 
@bhack let's make sure to save process meta-commentary in corresponding issues to make them easier to track and refer back to - feel free to open one on this topic or ping a stale issue if one is already open

bmzhao:
Oops as I was updating my fork, I accidentally rewrote this commit, which automatically closed this PR; unfortunately, it looks like re-pushing will create a new PR.

PR #182, opened 2019-11-27
RFC: Improved pip package structure 

annarev:
*Comment period is open until 2019-11-13*

# Improved pip package structure

| **RFC #**     | [182](https://github.com/tensorflow/community/pull/182) |
| **Author(s)** | Anna Revinskaya (annarev@google.com)                 |
| **Sponsor**   | Alex Passos (apassos@tensorflow.org)                 |
| **Updated**   | 2019-11-27                                           |

## Objective

We propose to simplify TensorFlow pip package structure to enable IDE features such as autocomplete, jump-to-definition and quick-documentation.


rfcs/20191127-pip-structure.md
```diff
@@ -0,0 +1,212 @@
...
+TensorFlow Metapackage would be a thin package defining composition of TensorFlow which includes base, estimator, keras and tensorboard.
+Note that this 2-package approach is not implemented yet. However, its proposal demonstrates how keeping a virtual pip package could be beneficial in the future.
+
+![alt_text](https://github.com/annarev/community/blob/pip_structure_rfc/rfcs/20191127-pip-structure/modular_structure.png "Proposed modular TensorFlow structure.")
+
+Current structure looks more like this (except *tensorflow/* and *tensorflow\_core/* are directories as opposed to separate pip packages) and meant to be the first step towards structure above:
+
+![alt_text](https://github.com/annarev/community/blob/pip_structure_rfc/rfcs/20191127-pip-structure/current_structure.png "Current TensorFlow structure.")
+
+### Current state of IDE code features
```
caisq: For completeness, does it make sense to add a section can we add a section about CoLab and Jupyter? Many TF users use those.

rfcs/20191127-pip-structure.md
```diff
@@ -0,0 +1,212 @@
...
+Current code that looks like:
+```python
+import tensorflow as tf
+
+tf.estimator
+tf.keras
+tf.summary
+```
+
+Would be changed to:
```
caisq: These are API-breaking changes, right? If so, it may be good to spell out the version horizon of this. This will not happen until 3.0, presumably.

rfcs/20191127-pip-structure.md
```diff
@@ -0,0 +1,212 @@
...
+import tensorflow_keras as keras
+from tenosorboard import summaries
+```
+
+Rationale for this change:
+
+* One way dependencies (estimator depends on tensorflow and not vise-versa).
+* Minimal overhead for users. Adding an extra import is easy.
+
+Note that this change cannot be done in TensorFlow 2.0 due to API guarantees. However, we can keep `tf.estimator`, `tf.keras` (once it is moved out of TensorFlow), `tf.summary` available while encouraging users to import from tensorflow\_estimator, tensorflow\_keras (once available) and tensorboard instead. This would require some work to make sure these packages contain the right API (for e.g. tensorflow\_estimator.estimator currently always contains V1 API).
```
caisq: s/2.0/2.x/ ?

rfcs/20191127-pip-structure.md
```diff
@@ -0,0 +1,212 @@
...
+| Status        | (Proposed / Accepted / Implemented / Obsolete)       |
+:-------------- |:---------------------------------------------------- |
+| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)|
+| **Author(s)** | Anna Revinskaya (annarev@google.com)                 |
+| **Sponsor**   | Alex Passos (apassos@tensorflow.org)                 |
+| **Updated**   | 2019-11-27                                           |
+
+## Objective
+
+We propose to simplify TensorFlow pip package structure to enable IDE features such as autocomplete, jump-to-definition and quick-documentation.
```
seanpmorgan: Another IDE feature that is missing due to the current packaging is the ability to verify compliance with abstract base classes. I believe this should be taken care of by the proposal, but it's worth verifying as a useful feature. 

E.g. when implementing a custom keras metric, your IDE should tell you that you're missing `update_state`:
https://github.com/tensorflow/tensorflow/blob/ddca4b92b4eb4e187716b03f443362c904bdad7e/tensorflow/python/keras/metrics.py#L215

rfcs/20191127-pip-structure.md
```diff
@@ -0,0 +1,212 @@
...
+* Quick documentation works.
+
+#### VS Code 1.40 (October 2019 release)
+* Autocomplete:
+  * Works in most cases.
+  * Doesn’t work for `tf.estimator` or `tf.keras`.
+  * Doesn’t work for `tf.compat.v1.keras` and `tf.compat.v2.keras`.
+  * Doesn’t work for keras if importing it using from import (i.e. `from tensorflow import keras`).
+* Jump-to-definition doesn’t work.
+* Quick documentation doesn’t work.
```
lgeiger: The same holds true for IDEs with [Jedi](https://jedi.readthedocs.io/en/latest/) based completion and introspection (e.g. [`ide-python`](https://github.com/lgeiger/ide-python) for [Atom](https://atom.io/), [`jedi-vim`](https://github.com/davidhalter/jedi-vim), [Spyder](https://www.spyder-ide.org/)).

For more info see https://github.com/davidhalter/jedi/issues/1391

rfcs/20191127-pip-structure.md
alextp: I don't think those are adversely affected by our package structure though

seanpmorgan: Just checked `tf.keras.layers.*` tab completion in jupyter and you're correct that it's not affected.

rfcs/20191127-pip-structure.md
foxik: I am using Jedi in vim and it is working, by including `tensorflow` into https://jedi.readthedocs.io/en/latest/docs/settings.html#jedi.settings.auto_import_modules -- that way tensorflow is not statically analyzed, bud imported on the Python level, and autocompletion works the same way as in ipython.

rfcs/20191127-pip-structure.md
```diff
@@ -0,0 +1,212 @@
...
+import tensorflow as tf
+import tensorflow_estimator as estimator
+import tensorflow_keras as keras
+from tenosorboard import summaries
+```
+
+Rationale for this change:
+
+* One way dependencies (estimator depends on tensorflow and not vise-versa).
+* Minimal overhead for users. Adding an extra import is easy.
```
foxik: Myself I vote against (if I can get a vote :-) Adding an extra import is easy, but adding 3 additional imports is less so, and add an import each time you use more and more functionality is bothersome. Especially `keras` and `tensorboard` seem basic part of TF 2 (at least in my view), and importing them from another package seems wrong.

On the next line you talk about keeping aliases like `tf.summary` -- if you plan to support that, than I would be grateful. But maybe the plan is to deprecate and move users to individual imports?

annarev: Primary reason for this is to avoid circular structure. keras, estimator and tensorboard currently all depend on tensorflow pip package (they would use import tensorflow.*). If we make tf.estimator, tf.keras and tf.summary available (which is the case now), then we have a reverse dependency as well. This is a cycle.

That being said, I am definitely interested in hearing opinions about having these separate imports! If tf.keras and tf.estimator are very useful, we could consider ways to deal with circular imports.

There are two ways I can think of to deal with circular imports now:
1. Lazy loading.
2. There is another option in Alternatives Considered right below. Basically, instead of having these separate imports, we can rename tensorflow package to tensorflow_core. estimator, keras and tensorboard would then use "from tensorflow_core.". Then we have another package on top called tensorflow that will combine tensorflow_core, estimator, keras and tensorboard into one package.

Both of these options seem less clean to me than just having the extra imports, but I am open to other options based on feedback.

rfcs/20191127-pip-structure.md
```diff
@@ -0,0 +1,212 @@
+# Title of RFC
```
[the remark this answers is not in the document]
annarev: Thank you for catching this, updated.

martinwicke:
I think lazy loading is a good option. It is complicated by our current
practice of testing which mixes cross-package integration tests with the
rest. Otherwise it would not be unclean enough to bother me.

brijk7:
@annarev : is this ready for design review?

annarev:
Yes, it is ready for design review. There are still some things I wanted to test out following comments (unfortunately haven't gotten around to it). But it should be ok to have design review at this point.

rfcs/20191127-pip-structure.md
annarev: Sorry for not addressing this comment sooner. I just tried it out today.
New version of PyCharms would give abstract class warning out of the box. The old version that I tested while writing this doc (2019.1.1) would not give the warning and the change proposed in this RFC would fix it.

I tried in VS Code where some autocomplete does not work (tf.compat.v1.keras.) but could not find abstractclass warning even if I define a sample repro example.

rfcs/20191127-pip-structure.md
annarev: The current plan is to make it possible to import from stand alone packages (for e.g. tensorflow_estimator and tensorflow_keras) but at the same time keep `tf.estimator` and `tf.keras` available through lazy loading.

annarev:
> I think lazy loading is a good option. It is complicated by our current practice of testing which mixes cross-package integration tests with the rest. Otherwise it would not be unclean enough to bother me.

@martinwicke Do you have an example in mind when an integration test would fail because we use lazy-loading as opposed to adding tensorflow_core package?

Would these cases be resolved once estimator just depends on public APIs? I am guessing we can split up the tests if they still cause an issue then.

martinwicke:
For tests which naturally live in Estimator, yes, they should be fine. If
the dependency tree is a real tree without cycles, then putting integration
tests in the highest spot needed to see all dependencies for the test will
make this clean, independent of lazy loading.

annarev:
Ideally it should be a tree except the lazy-loaded edges. I see some `from tensorflow_estimator` references in python/tpu and python/estimator directories. My guess is these should be going away and were probably left from before estimator migration.

annarev:
Note that `from tensorflow import keras` issues have been fixed by this PR: https://github.com/tensorflow/tensorflow/pull/34629 from @lgeiger.

rfcs/20191127-pip-structure.md
```diff
@@ -0,0 +1,204 @@
...
+tf.estimator
+tf.keras
+tf.summary
+```
+
+Would be changed to:
+```python
+import tensorflow as tf
+import tensorflow_estimator as estimator
+import tensorflow_keras as keras
```
gabrieldemarmiesse: As discussed in #202
```suggestion
import keras
```

gabrieldemarmiesse: Thanks @annarev!

mihaimaruseac:
I think this can be merged as it has been already implemented

bhack:
> I think this can be merged as it has been already implemented
@mihaimaruseac Do you meant that the implementation was already merged in the public repo?

mihaimaruseac:
Yes, the commit with the implementation is literally the latest one on this PR.

1EmKunuzL31

martinwicke:
@ematejska can you go through the accepted by unmerged PRs? I think there are a few.

@bhack thanks for pinging these.

PR #188, opened 2019-12-13
RFC: Keras categorical inputs

tanzhenyu:
*Comment period is open till Dec 31, 2019.*

# Keras categorical inputs

| **RFC #** | [188](https://github.com/tensorflow/community/pull/188) |
| **Author(s)** | Zhenyu Tan (tanzheny@google.com), Francois Chollet (fchollet@google.com)|
| **Sponsor**   | Karmel Allison (karmel@google.com), Martin Wicke (wicke@google.com) |
| **Updated**   | 2019-12-12                                           |

## Objective

This document proposes 4 new preprocessing Keras layers (`CategoryLookup`, `CategoryCrossing`, `CategoryEncoding`, `CategoryHashing`), and 1 additional op (`to_sparse`) to allow users to:
* Perform feature engineering for categorical inputs
* Replace feature columns and `tf.keras.layers.DenseFeatures` with proposed layers
* Introduce sparse inputs that work with Keras linear models and other layers that support sparsity

### DIFF
--- a/rfcs/20191212-keras-categorical-inputs.md
+++ b/rfcs/20191212-keras-categorical-inputs.md
@@ -314,8 +314,28 @@ def to_sparse(input, ignore_value):
 ### Alternatives Considered
 An alternative is to provide solutions on top of feature columns. This will make user code to be slightly cleaner but far less flexible.
 
+### Performance Implications
+End to End benchmark should be same as other preprocessing layers.
+
+### Dependencies
+This proposal does not add any new dependencies.
+
+### Engineering Impact
+These changes will include more layers and thus binary size and build time. It will not impact startup time.
+This code can be tested in its own and maintained in its own buildable unit.
+
+### Platforms and Environments
+This proposal should work in all platforms and environments.
+
+### Best Practices, Tutorials and Examples
+This proposal does not change the best engineering practices.
+
 ### Compatibility
 No backward compatibility issues.
+
+### User Impact
+User facing changes to migrate feature column based Keras modeling to preprocessing layer based Keras modeling, as the example workflow suggests.
+
 
 ## Code Snippets
 
--- a/rfcs/20191212-keras-categorical-inputs.md
+++ b/rfcs/20191212-keras-categorical-inputs.md
@@ -15,6 +15,17 @@ This document proposes 4 new preprocessing Keras layers (`CategoryLookup`, `Cate
 
 Other proposed layers for replacement of feature columns such as `tf.feature_column.bucketized_column` and `tf.feature_column.numeric_column` has been discussed [here](https://github.com/keras-team/governance/blob/master/rfcs/20190502-preprocessing-layers.md) and are not the focus of this document.
 
+## Motivation
+
+Specifically, by introducing the 4 layers, we aim to address these pain points:
+* Users have to define both feature columns and Keras Inputs for the model, resulting in code duplication and deviation from DRY (Do not repeat yourself) principle. See this [Github issue](https://github.com/tensorflow/tensorflow/issues/27416).
+* Users with large dimension categorical inputs will incur large memory footprint and computation cost, if wrapped with indicator column through `tf.keras.layers.DenseFeatures`.
+* Currently there is no way to correctly feed Keras linear model or dense layer with multivalent categorical inputs or weighted categorical inputs.
+
+## User Benefit
+
+We expect to get rid of the user painpoints once migrating off feature columns.
+
 ## Example Workflows
 
 Two example workflows are presented below. These workflows can be found at this [colab](https://colab.sandbox.google.com/drive/1cEJhSYLcc2MKH7itwcDvue4PfvrLN-OR#scrollTo=22sa0D19kxXY).
@@ -24,6 +35,9 @@ Two example workflows are presented below. These workflows can be found at this 
 The first example gives an equivalent code snippet to canned `LinearEstimator` [tutorial](https://www.tensorflow.org/tutorials/estimator/linear) on the Titanic dataset:
 
 ```python
+dftrain = pd.read_csv('https://storage.googleapis.com/tf-datasets/titanic/train.csv')
+y_train = dftrain.pop('survived')
+
 CATEGORICAL_COLUMNS = ['sex', 'n_siblings_spouses', 'parch', 'class', 'deck', 'embark_town', 'alone']
 NUMERICAL_COLUMNS = ['age', 'fare']
 # input list to create functional model.
@@ -49,9 +63,6 @@ linear_logits = linear_model(linear_inputs)
 model = tf.keras.Model(model_inputs, linear_logits)
 
 model.compile('sgd', loss=tf.keras.losses.BinaryCrossEntropy(from_logits=True), metrics=['accuracy'])
-
-dftrain = pd.read_csv('https://storage.googleapis.com/tf-datasets/titanic/train.csv')
-y_train = dftrain.pop('survived')
 
 dataset = tf.data.Dataset.from_tensor_slices((
 	(tf.to_sparse(dftrain.sex, "Unknown"), tf.to_sparse(dftrain.n_siblings_spouses, -1),
@@ -173,13 +184,6 @@ x = tf.keras.layers.CategoryEncoding(num_categories=len(vocabulary_list)+num_oov
 linear_model = tf.keras.premade.LinearModel(units)
 linear_logits = linear_model(x)
 ```
-
-## Pain Points
-
-Specifically, by introducing the 4 layers, we aim to address these pain points:
-* Users have to define both feature columns and Keras Inputs for the model, resulting in code duplication and deviation from DRY (Do not repeat yourself) principle. See this [Github issue](https://github.com/tensorflow/tensorflow/issues/27416).
-* Users with large dimension categorical inputs will incur large memory footprint and computation cost, if wrapped with indicator column through `tf.keras.layers.DenseFeatures`.
-* Currently there is no way to correctly feed Keras linear model or dense layer with multivalent categorical inputs or weighted categorical inputs.
 
 ## Design Proposal
 We propose a CategoryLookup layer to replace `tf.feature_column.categorical_column_with_vocabulary_list` and `tf.feature_column.categorical_column_with_vocabulary_file`, a `CategoryHashing` layer to replace `tf.feature_column.categorical_column_with_hash_bucket`, a `CategoryCrossing` layer to replace `tf.feature_column.crossed_column`, and another `CategoryEncoding` layer to convert the sparse input to the format required by linear models.
@@ -307,6 +311,12 @@ def to_sparse(input, ignore_value):
   pass
 ```
 
+### Alternatives Considered
+An alternative is to provide solutions on top of feature columns. This will make user code to be slightly cleaner but far less flexible.
+
+### Compatibility
+No backward compatibility issues.
+
 ## Code Snippets
 
 Below is a more detailed illustration of how each layer works. If there is a vocabulary list of countries:
@@ -355,3 +365,6 @@ cross_out.indices = <tf.Tensor: id=186, shape=(2, 2), dtype=int64, numpy=
                         array([[0, 0], [0, 1]])>
 cross_out.values = <tf.Tensor: id=187, shape=(2,), dtype=int64, numpy=array([3, 3])>
 ```
+
+## Questions and Discussion Topics
+We'd like to gather feedbacks on `CategoryLookup`, specifically we propose migrating off from mutually exclusive `num_oov_buckets` and `default_value` and replace with `num_oov_tokens`.
--- a/rfcs/20191212-keras-categorical-inputs.md
+++ b/rfcs/20191212-keras-categorical-inputs.md
@@ -8,12 +8,14 @@
 
 ## Objective
 
-This document proposes 4 new preprocessing Keras layers (`CategoryLookup`, `CategoryCrossing`, `CategoryEncoding`, `CategoryHashing`), and an extension to existing op (`tf.sparse.from_dense`) to allow users to:
+This document proposes 4 new preprocessing Keras layers (`Lookup`, `CategoryCrossing`, `Vectorize`, `FingerPrint`), and an extension to existing op (`tf.sparse.from_dense`) to allow users to:
 * Perform feature engineering for categorical inputs
 * Replace feature columns and `tf.keras.layers.DenseFeatures` with proposed layers
 * Introduce sparse inputs that work with Keras linear models and other layers that support sparsity
 
 Other proposed layers for replacement of feature columns such as `tf.feature_column.bucketized_column` and `tf.feature_column.numeric_column` has been discussed [here](https://github.com/keras-team/governance/blob/master/rfcs/20190502-preprocessing-layers.md) and are not the focus of this document.
+
+The proposed layers should support ragged tensors.
 
 ## Motivation
 
@@ -48,8 +50,8 @@ for feature_name in CATEGORICAL_COLUMNS:
 	feature_input = tf.keras.Input(shape=(1,), dtype=tf.string, name=feature_name, sparse=True)
 	vocab_list = sorted(dftrain[feature_name].unique())
 	# Map string values to indices
-	x = tf.keras.layers.CategoryLookup(vocabulary=vocab_list, name=feature_name)(feature_input)
-  x = tf.keras.layers.CategoryEncoding(num_categories=len(vocab_list))(x)
+	x = tf.keras.layers.Lookup(vocabulary=vocab_list, name=feature_name)(feature_input)
+  x = tf.keras.layers.Vectorize(num_categories=len(vocab_list))(x)
 	linear_inputs.append(x)
 	model_inputs.append(feature_input)
 
@@ -87,7 +89,7 @@ fc = tf.feature_column.categorical_feature_column_with_vocabulary_list(
 Proposed:
 ```python
 x = tf.keras.Input(shape=(1,), name=key, dtype=dtype)
-layer = tf.keras.layers.CategoryLookup(
+layer = tf.keras.layers.Lookup(
             vocabulary=vocabulary_list, num_oov_tokens=num_oov_buckets)
 out = layer(x)
 ```
@@ -103,7 +105,7 @@ fc = tf.feature_column.categorical_column_with_vocab_file(
 Proposed:
 ```python
 x = tf.keras.Input(shape=(1,), name=key, dtype=dtype)
-layer = tf.keras.layers.CategoryLookup(
+layer = tf.keras.layers.Lookup(
             vocabulary=vocabulary_file, num_oov_tokens=num_oov_buckets)
 out = layer(x)
 ```
@@ -119,7 +121,7 @@ fc = tf.feature_column.categorical_column_with_hash_bucket(
 Proposed:
 ```python
 x = tf.keras.Input(shape=(1,), name=key, dtype=dtype)
-layer = tf.keras.layers.CategoryHashing(num_bins=hash_bucket_size)
+layer = tf.keras.layers.FingerPrint(num_bins=hash_bucket_size)
 out = layer(x)
 ```
 
@@ -151,11 +153,11 @@ Proposed:
 ```python
 x1 = tf.keras.Input(shape=(1,), name=key_1, dtype=dtype)
 x2 = tf.keras.Input(shape=(1,), name=key_2, dtype=dtype)
-layer1 = tf.keras.layers.CategoryLookup(
+layer1 = tf.keras.layers.Lookup(
            vocabulary=vocabulary_list,  
            num_oov_tokens=num_oov_buckets)
 x1 = layer1(x1)
-layer2 = tf.keras.layers.CategoryHashing(
+layer2 = tf.keras.layers.FingerPrint(
            num_bins=hash_bucket_size)
 x2 = layer2(x2)
 layer = tf.keras.layers.CategoryCrossing(num_bins=hash_bucket_size)
@@ -176,27 +178,27 @@ Proposed:
 ```python
 x1 = tf.keras.Input(shape=(1,), name=key, dtype=dtype)
 x2 = tf.keras.Input(shape=(1,), name=weight_feature_key, dtype=weight_dtype)
-layer = tf.keras.layers.CategoryLookup(
+layer = tf.keras.layers.Lookup(
            vocabulary=vocabulary_list,   
            num_oov_tokens=num_oov_buckets)
 x1 = layer(x1)
-x = tf.keras.layers.CategoryEncoding(num_categories=len(vocabulary_list)+num_oov_buckets)([x1, x2])
+x = tf.keras.layers.Vectorize(num_categories=len(vocabulary_list)+num_oov_buckets)([x1, x2])
 linear_model = tf.keras.premade.LinearModel(units)
 linear_logits = linear_model(x)
 ```
 
 ## Design Proposal
-We propose a `CategoryLookup` layer to replace `tf.feature_column.categorical_column_with_vocabulary_list` and `tf.feature_column.categorical_column_with_vocabulary_file`, a `CategoryHashing` layer to replace `tf.feature_column.categorical_column_with_hash_bucket`, a `CategoryCrossing` layer to replace `tf.feature_column.crossed_column`, and another `CategoryEncoding` layer to convert the sparse input to the format required by linear models.
+We propose a `Lookup` layer to replace `tf.feature_column.categorical_column_with_vocabulary_list` and `tf.feature_column.categorical_column_with_vocabulary_file`, a `FingerPrint` layer to replace `tf.feature_column.categorical_column_with_hash_bucket`, a `CategoryCrossing` layer to replace `tf.feature_column.crossed_column`, and another `Vectorize` layer to convert the sparse input to the format required by linear models.
 
 ```python
-`tf.keras.layers.CategoryLookup`
-CategoryLookup(PreprocessingLayer):
+`tf.keras.layers.Lookup`
+Lookup(PreprocessingLayer):
 """This layer transforms categorical inputs to index space.
    If input is dense/sparse, then output is dense/sparse."""
 
   def __init__(self, max_tokens=None, num_oov_tokens=1, vocabulary=None,
                name=None, **kwargs):
-    """Constructs a CategoryLookup layer.
+    """Constructs a Lookup layer.
 
     Args:
       max_tokens: The maximum size of the vocabulary for this layer. If None,
@@ -251,21 +253,22 @@ CategoryCrossing(PreprocessingLayer):
     """
     pass
 
-`tf.keras.layers.CategoryEncoding`
-CategoryEncoding(PreprocessingLayer):
+`tf.keras.layers.Vectorize`
+Vectorize(PreprocessingLayer):
 """This layer transforms categorical inputs from index space to category space.
    If input is dense/sparse, then output is dense/sparse."""
 
-  def __init__(self, num_categories, mode="sum", axis=-1, name=None, **kwargs):
-    """Constructs a CategoryEncoding layer.
+  def __init__(self, num_categories, mode="count", axis=-1, sparse_out=True, name=None, **kwargs):
+    """Constructs a Vectorize layer.
     Args:
       num_categories: Number of elements in the vocabulary.
-      mode: how to reduce a categorical input if multivalent, can be one of "sum",  
-          "mean", "binary", "tfidf". It can also be None if this is not a multivalent input,
+      mode: how to reduce a categorical input if multivalent, can be one of "count",  
+          "avg_count", "binary", "tfidf". It can also be None if this is not a multivalent input,
           and simply needs to convert input from index space to category space. "tfidf" is only
           valid when adapt is called on this layer.
       axis: the axis to reduce, by default will be the last axis, specially true 
           for sequential feature columns.
+      sparse_out: boolean to indicate whether the output should be dense or sparse tensor.
       name: Name to give to the layer.
      **kwargs: Keyword arguments to construct a layer.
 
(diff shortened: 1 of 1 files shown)


rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,357 @@
...
+	linear_inputs.append(feature_input)
+	model_inputs.append(feature_input)
+
+linear_model = tf.keras.experimental.LinearModel(units=1)
+linear_logits = linear_model(linear_inputs)
+model = tf.keras.Model(model_inputs, linear_logits)
+
+model.compile('sgd', loss=tf.keras.losses.BinaryCrossEntropy(from_logits=True), metrics=['accuracy'])
+
+dftrain = pd.read_csv('https://storage.googleapis.com/tf-datasets/titanic/train.csv')
```
martinwicke: It seems that this is referenced before assignment? Does this code run?

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,357 @@
...
+
+Original:
+```python
+fc = tf.feature_column.categorical_column_with_identity(
+       key, num_buckets, default_value)
+```
+Proposed:
+```python
+x = tf.keras.Input(shape=(1,), name=key, dtype=dtype)
+layer = tf.keras.layers.Lambda(lambda x: tf.where(tf.logical_or(x < 0, x > num_buckets), tf.fill(dims=tf.shape(x), value=default_value), x))
```
martinwicke: If we allowed to specify the hash function, this could also be folded into the CategoryHashing with an IdentityHash.

tanzhenyu: CategoryHashing does not check oov values, so if we have that then it's complicating the signature.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Fix applied.

tanzhenyu:
> Can you add (or not delete) the other questions in the template? The API and workflow section is very good.

Done.

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+linear_model = tf.keras.premade.LinearModel(units)
+linear_logits = linear_model(x)
+```
+
+## Design Proposal
+We propose a CategoryLookup layer to replace `tf.feature_column.categorical_column_with_vocabulary_list` and `tf.feature_column.categorical_column_with_vocabulary_file`, a `CategoryHashing` layer to replace `tf.feature_column.categorical_column_with_hash_bucket`, a `CategoryCrossing` layer to replace `tf.feature_column.crossed_column`, and another `CategoryEncoding` layer to convert the sparse input to the format required by linear models.
+
+```python
+`tf.keras.layers.CategoryLookup`
+CategoryLookup(PreprocessingLayer):
```
ebrevdo: Please don't forget to implement the correct `compute_output_signature` for these classes, since they will accept SparseTensorSpecs, and must emit SparseTensorSpecs in this case.

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+
+## User Benefit
+
+We expect to get rid of the user painpoints once migrating off feature columns.
+
+## Example Workflows
+
+Two example workflows are presented below. These workflows can be found at this [colab](https://colab.sandbox.google.com/drive/1cEJhSYLcc2MKH7itwcDvue4PfvrLN-OR#scrollTo=22sa0D19kxXY).
+
+### Workflow 1
```
ebrevdo: Would it make sense to provide an example workflow where you have to get the vocabulary from e.g. a csv file using tf.data and in particular, tf.data.experimental.unique and tf.data.experimental.get_single_element to read out the tensor?  @jsimsa wdyt?

This is gonna be very common, i think, in real use cases.

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+Original:
+```python
+fc = tf.feature_column.categorical_column_with_vocab_file(
+       key, vocabulary_file, vocabulary_size, dtype,
+       default_value, num_oov_buckets)
+```
+Proposed:
+```python
+x = tf.keras.Input(shape=(1,), name=key, dtype=dtype)
+layer = tf.keras.layers.CategoryLookup(
```
ebrevdo: Ah I see, so here you allow the user to provide a vocabulary file directly, so the tf.data example may not be necessary.  May still be useful if users have vocab that needs to be munged a bit before reading directly.  But less important.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Thanks for the reminder!

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+      max_tokens: The maximum size of the vocabulary for this layer. If None,
+              there is no cap on the size of the vocabulary. This is used when `adapt`
+              is called.
+      num_oov_tokens: Non-negative integer. The number of out-of-vocab tokens. 
+              All out-of-vocab inputs will be assigned IDs in the range of 
+              [0, num_oov_tokens) based on a hash. When
+              `vocabulary` is None, it will convert inputs in [0, num_oov_tokens)
+      vocabulary: the vocabulary to lookup the input. If it is a file, it represents the 
+              source vocab file; If it is a list/tuple, it represents the source vocab 
+              list; If it is None, the vocabulary can later be set.
```
jsimsa: what is the format of the file? how do you set the vocabulary later? what is the expected use of the `adapt` method?

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+    If one input sample is `["a", "c", "d", "a", "x"]` and the vocabulary is ["a", "b", "c", "d"],
+    and a single OOV token is used (`num_oov_tokens=1`), then the corresponding output sample is
+    `[1, 3, 4, 1, 0]`. 0 stands for an OOV token.
+    """
+    pass
+
+`tf.keras.layers.CategoryCrossing`
+CategoryCrossing(PreprocessingLayer):
+"""This layer transforms multiple categorical inputs to categorical outputs
+   by Cartesian product. and hash the output if necessary.
```
jsimsa: nit: remove extra `.`

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+   If any input is sparse, then output is sparse, otherwise dense."""
```
jsimsa: OOC, why is the wording here different than in the other API endpoints (it seems that the intended behavior is the same?)

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+
+  def __init__(self, depth=None, num_bins=None, name=None, **kwargs):
+    """Constructs a CategoryCrossing layer.
+    Args:
+      depth: depth of input crossing. By default None, all inputs are crossed
+             into one output. It can be an int or tuple/list of ints, where inputs are
+             combined into all combinations of output with degree of `depth`. For example,
+             with inputs `a`, `b` and `c`, `depth=2` means the output will be [ab;ac;bc]
```
jsimsa: the example should be moved to the "Example" section below

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+
+    Example:
+
+    If the layer receives two inputs, `a=[[1, 2]]` and `b=[[1, 3]]`,
+    and if depth is 2, then
+    the output will be a single integer tensor `[[i, j, k, l]]`, where:
+    i is the index of the category "a1=1 and b1=1"
+    j is the index of the category "a1=1 and b2=3"
+    k is the index of the category "a2=2 and b1=1"
+    l is the index of the category "a2=2 and b2=3"
```
jsimsa: I don't understand this example. What is `a1` vs `a2`? What will the "single integer tensor `[[i, j, k, l]]`" actually look like for the given inputs?

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+    """
+    pass
+
+`tf.keras.layers.CategoryEncoding`
+CategoryEncoding(PreprocessingLayer):
```
jsimsa: Please add example for this layer.

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+      name: Name to give to the layer.
+     **kwargs: Keyword arguments to construct a layer.
+
+    Input: a int tensor of shape `[batch_size, d1, ..., dm-1, dm]`
+    Output: a float tensor of shape `[batch_size, d1, ..., dm-1, num_categories]`
+    """
+    pass
+
+`tf.keras.layers.CategoryHashing`
+CategoryHashing(PreprocessingLayer):
```
jsimsa: Please add example for this layer.

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+      **kwargs: Keyword arguments to construct a layer.
+
+    Input: a int tensor of shape `[batch_size, d1, ..., dm]`
+    Output: a int tensor of shape `[batch_size, d1, ..., dm]`
+    """
+    pass
+
+```
+
+We also propose a `to_sparse` op to convert dense tensors to sparse tensors given user specified ignore values. This op can be used in both `tf.data` or [TF Transform](https://www.tensorflow.org/tfx/transform/get_started). In previous feature column world, "" is ignored for dense string input and -1 is ignored for dense int input.
```
jsimsa: I would prefer if the `SparseTensor` class had a `from_dense` method.

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+
+```python
+`tf.to_sparse`
+def to_sparse(input, ignore_value):
+  """Convert dense/sparse tensor to sparse while dropping user specified values.
```
jsimsa: What is the benefit of calling this API with a SparseTensor input?

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: 1) format of the file is same as 
a) any other TFX vocab file, or 
b) this test file: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/feature_column/testdata/warriors_vocabulary.txt

2) users from feature columns world will set it during __init__, but this layer also allow users to call 'adapt' to derive/set the vocabulary from dataset.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Good question. This is the only layer that can accept multiple inputs. Other API only accept a single Tensor/SparseTensor.
So by multiple inputs, if any one of them is sparse, the output will be sparse.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: to allow users to filter specified values, e.g., if the original input is already sparse:
indices = [[0,0], [1, 0], [1,1]]
values = ['A', '', 'C']
the user can still filter '' from it,

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: If we don't need the functionality of `sparse_output = to_sparse(sparse_input)`, then `from_dense` is probably better.
This "imagined" functionality is not used anywhere though. In TFT I think any tf.io.VarLenFeature should automatically be sparse input, we just need to call SparseTensor.from_dense for any tf.io.FixedLenFeature.

WDYT?

rfcs/20191212-keras-categorical-inputs.md
jsimsa: My point is that this should be documented. Stating the the vocabulary can be set later without showing how is not useful.

rfcs/20191212-keras-categorical-inputs.md
jsimsa: Got it. Maybe you should say, "If any of the inputs is sparse, then all outputs will be sparse. Otherwise, all outputs will be dense."

rfcs/20191212-keras-categorical-inputs.md
jsimsa: I realized, we already have [from_dense](https://www.tensorflow.org/api_docs/python/tf/sparse/from_dense?version=stable), so perhaps you should just extend it with the option to set the element to be ignore?

rfcs/20191212-keras-categorical-inputs.md
jsimsa: This filtering can be built out of existing operations. You can call, [tf.where](https://www.tensorflow.org/api_docs/python/tf/where?version=stable) on `values` and pass the result to [tf.sparse.retain](https://www.tensorflow.org/api_docs/python/tf/sparse/retain?version=stable), which is simple enough that I do not see the point of introducing syntactic sugar for that.

workingloong:
With `tf.feature_column`, we can use `embedding_column` to wrap `category_column` and convert `category_column` output to dense tensor. Whether can Keras Category layers support the function like `embedding_column`?
Meanwhile, `tf.keras.layers.Embedding` cannot support sparseTensor inputs which may be the output of `CategoryLookup`. I have created an issue https://github.com/tensorflow/tensorflow/issues/33880 to embedding lookup with sparseTensor.

tanzhenyu:
> With `tf.feature_column`, we can use `embedding_column` to wrap `category_column` and convert `category_column` output to dense tensor. Whether can Keras Category layers support the function like `embedding_column`?
> Meanwhile, `tf.keras.layers.Embedding` cannot support sparseTensor inputs which may be the output of `CategoryLookup`. I have created an issue [tensorflow/tensorflow#33880](https://github.com/tensorflow/tensorflow/issues/33880) to embedding lookup with sparseTensor.

You can use dense tensor input to CategoryLookup, which gives you dense tensor output, and feed that into `tf.keras.layers.Embedding`.

Maybe we should support sparse input in embedding layer.

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+CategoryLookup(PreprocessingLayer):
+"""This layer transforms categorical inputs to index space.
+   If input is dense/sparse, then output is dense/sparse."""
+
+  def __init__(self, max_tokens=None, num_oov_tokens=1, vocabulary=None,
+               name=None, **kwargs):
+    """Constructs a CategoryLookup layer.
+
+    Args:
+      max_tokens: The maximum size of the vocabulary for this layer. If None,
```
brightcoder01: Before executing the training loop, we will call preprocessLayers.adapt to calculate the statistical value. if the dataset is very huge, how does `adapt` handle this case? Does it support distributed process?

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,390 @@
...
+              there is no cap on the size of the vocabulary. This is used when `adapt`
```
brightcoder01: IMO, the statistical process will be executed in a pre-process stage before model training just as the stage where TF-Transform stands in the End-to-End pipeline. How do we pass the statistical result from preprocess stage to the model training stage using keras preprocess input layer?

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Done.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Done.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Yeah that's better. Done.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Both Example for each layer description, and a code snippet below.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Yeah it is confusing. Updated.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Done.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Done.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Yeah good point. I wasn't aware of this op. We should just extend it. Done.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Makes sense. Let's just extend the `tf.sparse.from_dense` op.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: I think this layer is more of a complimentary to that, i.e., tf.data can parse records and generate vocab file, of read vocab file and do other processing and still return string tensors. This layer is taken from that and convert things to indices before it gets to embedding.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: This really depends on what the execution engine is, and not part of the layer's responsibility, but instead should be responsibility of [ProcessingStage](https://github.com/keras-team/governance/blob/master/rfcs/20190502-preprocessing-layers.md#prepocessingstage).
So in Keras, the only choice for now (same as TextVectorization layer), is single process because Keras is not designed for distributed running. In TFX that relies on Apache Beam to execute TensorFlow operations, this will be on distributed process.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: You're right, the `adapt` call is more useful in TF-Transform, not in Keras, if you have a large dataset. However we still provide `adapt` in Keras for smaller datasets.
The way to pass statistical values is through `tft.compute_and_apply_vocabulary`, which was using raw TF operations before and it is targeted to be replaced by Keras preprocessing layers.

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,398 @@
...
+
+Original:
+```python
+fc = tf.feature_column.categorical_column_with_identity(
+       key, num_buckets, default_value)
+```
+Proposed:
+```python
+x = tf.keras.Input(shape=(1,), name=key, dtype=dtype)
+layer = tf.keras.layers.Lambda(lambda x: tf.where(tf.logical_or(x < 0, x > num_buckets), tf.fill(dims=tf.shape(x), value=default_value), x))
```
brightcoder01: The output of `categorical_column_with_identity` is a SpareTensor.
The output of `layer = tf.keras.layers.Lambda(lambda x: tf.where(tf.logical_or(x < 0, x > num_buckets), tf.fill(dims=tf.shape(x), value=default_value), x))` seems to be a Tensor, not a SparseTensor.

workingloong:
> > With `tf.feature_column`, we can use `embedding_column` to wrap `category_column` and convert `category_column` output to dense tensor. Whether can Keras Category layers support the function like `embedding_column`?
> > Meanwhile, `tf.keras.layers.Embedding` cannot support sparseTensor inputs which may be the output of `CategoryLookup`. I have created an issue [tensorflow/tensorflow#33880](https://github.com/tensorflow/tensorflow/issues/33880) to embedding lookup with sparseTensor.
> 
> You can use dense tensor input to CategoryLookup, which gives you dense tensor output, and feed that into `tf.keras.layers.Embedding`.
> 
> Maybe we should support sparse input in embedding layer.

Look forward to an embedding layer with sparse input. And do you have a time plan when to implement the proposed preprocessing Keras layers?

tanzhenyu:
> > > With `tf.feature_column`, we can use `embedding_column` to wrap `category_column` and convert `category_column` output to dense tensor. Whether can Keras Category layers support the function like `embedding_column`?
> > > Meanwhile, `tf.keras.layers.Embedding` cannot support sparseTensor inputs which may be the output of `CategoryLookup`. I have created an issue [tensorflow/tensorflow#33880](https://github.com/tensorflow/tensorflow/issues/33880) to embedding lookup with sparseTensor.
> > 
> > 
> > You can use dense tensor input to CategoryLookup, which gives you dense tensor output, and feed that into `tf.keras.layers.Embedding`.
> > Maybe we should support sparse input in embedding layer.
> 
> Look forward to an embedding layer with sparse input. And do you have a time plan when to implement the proposed preprocessing Keras layers?

Should be ready in TF 2.2

workingloong:
> Should be ready in TF 2.2

Do you know when to release TF 2.2? We are designing a data transformation process with keras preprocess layers or `feature_column` for [SQLFLow](https://github.com/sql-machine-learning/sqlflow) and [ElasticDL](https://github.com/sql-machine-learning/elasticdl) based on TF 2.x.

tanzhenyu:
> > Should be ready in TF 2.2
> 
> Do you know when to release TF 2.2? We are designing a data transformation process with keras preprocess layers or `feature_column` for [SQLFLow](https://github.com/sql-machine-learning/sqlflow) and [ElasticDL](https://github.com/sql-machine-learning/elasticdl) based on TF 2.x.

@goldiegadde will have more accurate info on this. If you have special requirements on this, feel free to comment on it, or maybe ask @goldiegadde to see if you want to participate in the design review.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: that's why we proposed a tf.sparse.from_dense(value, ignore_value), and our philosophy here is tensor type in / tensor type out, i.e., you can specifically convert to sparse tensors before it gets to here, which is pretty useful in TF Transform you all preprocessing things are done in memory efficient way.

theadactyl:
@tanzhenyu were there notes from the design review meeting that could be posted here?

tanzhenyu:
> @tanzhenyu were there notes from the design review meeting that could be posted here?

Yeah I will update it soon.

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,401 @@
...
+# input list to create functional model.
+model_inputs = []
+# input list to feed linear model.
+linear_inputs = []
+for feature_name in CATEGORICAL_COLUMNS:
+	feature_input = tf.keras.Input(shape=(1,), dtype=tf.string, name=feature_name, sparse=True)
+	vocab_list = sorted(dftrain[feature_name].unique())
+	# Map string values to indices
+	x = tf.keras.layers.Lookup(vocabulary=vocab_list, name=feature_name)(feature_input)
+  x = tf.keras.layers.Vectorize(num_categories=len(vocab_list))(x)
```
haifeng-jin: incorrect indent.

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,401 @@
...
+	linear_inputs.append(feature_input)
+	model_inputs.append(feature_input)
+
+linear_model = tf.keras.experimental.LinearModel(units=1)
+linear_logits = linear_model(linear_inputs)
+model = tf.keras.Model(model_inputs, linear_logits)
+
+model.compile('sgd', loss=tf.keras.losses.BinaryCrossEntropy(from_logits=True), metrics=['accuracy'])
+
+dataset = tf.data.Dataset.from_tensor_slices((
```
haifeng-jin: Is it possible to have a single input node and use some layer to decompose the input to these single column nodes?

rfcs/20191212-keras-categorical-inputs.md
workingloong: Why do you not use a layer to the lamda layer?

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: use tf.split for that?

ematejska:
The CLA looks good. tanzhenyu signed the CLA but some of the commits were from his laptop acocunt.

tanzhenyu:
> The CLA looks good. tanzhenyu signed the CLA but some of the commits were from his laptop acocunt.

Yeah I created https://github.com/tensorflow/community/pull/209 as a workaround -- which one should we merge here?

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -0,0 +1,401 @@
+# Keras categorical inputs
+
+| Status        | Proposed                                             |
```
ematejska: Please change to Accepted

ematejska:
Let's continue in this one since it has the history and get this one merged.  Also, do you have notes from the review meeting you could post here?

brightcoder01:
> > > > With `tf.feature_column`, we can use `embedding_column` to wrap `category_column` and convert `category_column` output to dense tensor. Whether can Keras Category layers support the function like `embedding_column`?
> > > > Meanwhile, `tf.keras.layers.Embedding` cannot support sparseTensor inputs which may be the output of `CategoryLookup`. I have created an issue [tensorflow/tensorflow#33880](https://github.com/tensorflow/tensorflow/issues/33880) to embedding lookup with sparseTensor.
> > > 
> > > 
> > > You can use dense tensor input to CategoryLookup, which gives you dense tensor output, and feed that into `tf.keras.layers.Embedding`.
> > > Maybe we should support sparse input in embedding layer.
> > 
> > 
> > Look forward to an embedding layer with sparse input. And do you have a time plan when to implement the proposed preprocessing Keras layers?
> 
> Should be ready in TF 2.2

Hi, @tanzhenyu , from the [API spec for TF2.2rc](https://tensorflow.google.cn/versions/r2.2/api_docs/python/tf/keras/preprocessing), the preprocessing layers are not included. Will this layers be released in TF2.2 as planned?

tanzhenyu:
> Let's continue in this one since it has the history and get this one merged. Also, do you have notes from the review meeting you could post here?

It doesn't seem I could update this once the original branch is gone. What can we do to fix here?

workingloong:
> > With `tf.feature_column`, we can use `embedding_column` to wrap `category_column` and convert `category_column` output to dense tensor. Whether can Keras Category layers support the function like `embedding_column`?
> > Meanwhile, `tf.keras.layers.Embedding` cannot support sparseTensor inputs which may be the output of `CategoryLookup`. I have created an issue [tensorflow/tensorflow#33880](https://github.com/tensorflow/tensorflow/issues/33880) to embedding lookup with sparseTensor.
> 
> You can use dense tensor input to CategoryLookup, which gives you dense tensor output, and feed that into `tf.keras.layers.Embedding`.
> 
> Maybe we should support sparse input in embedding layer.

In the RFC, the preprocessing layers like `Lookup` and `FingerPrint` support `SparseTenor` in and out. But now, we cannot feed the output `SparseTensor` into `tf.keras.layers.Embedding`. Will the embedding with sparse input be released with those preprocessing layers?

tanzhenyu:
> > > With `tf.feature_column`, we can use `embedding_column` to wrap `category_column` and convert `category_column` output to dense tensor. Whether can Keras Category layers support the function like `embedding_column`?
> > > Meanwhile, `tf.keras.layers.Embedding` cannot support sparseTensor inputs which may be the output of `CategoryLookup`. I have created an issue [tensorflow/tensorflow#33880](https://github.com/tensorflow/tensorflow/issues/33880) to embedding lookup with sparseTensor.
> > 
> > 
> > You can use dense tensor input to CategoryLookup, which gives you dense tensor output, and feed that into `tf.keras.layers.Embedding`.
> > Maybe we should support sparse input in embedding layer.
> 
> In the RFC, the preprocessing layers like `Lookup` and `FingerPrint` support `SparseTenor` in and out. But now, we cannot feed the output `SparseTensor` into `tf.keras.layers.Embedding`. Will the embedding with sparse input be released with those preprocessing layers?

That's a very good question. We're currently gathering use cases for supporting sparse with Embedding layer, `Lookup` is definitely the most important one, but if you have other use cases, would you mind sharing with us?

rfcs/20191212-keras-categorical-inputs.md
brightcoder01: How about add a layer to do the Identity Work? 
1. The lambda expression is a little complicated
2. We may need the layer to handle both dense tensor and SparseTensor.

workingloong:
> That's a very good question. We're currently gathering use cases for supporting sparse with Embedding layer, `Lookup` is definitely the most important one, but if you have other use cases, would you mind sharing with us?

Yes. Except `Lookup`, `Hashing` and `Bucketize` which support sparse in and sparse out should be use cases. In [ElasticDL](https://github.com/sql-machine-learning/elasticdl), we are developing some preprocessing layers like [RoundIdentity](https://github.com/sql-machine-learning/elasticdl/pull/1842). What's more, we sometimes need to convert the inputs to sparse tensor ignoring missing value and transform the sparse tensor using Keras preprocessing layers. The detail is in the ElasticDL [issue](https://github.com/sql-machine-learning/elasticdl/issues/1844). 

In TF 2.1.0, I find that the `tf.keras.layers.Embedding` has supported `RaggedTensor` and output `RaggedTensor` with embedding vectors. In my opinion, the function of `RaggedTensor` is the same as `SparseTensor`. And we can use `tf.keras.layers.Embedding` with ragged tensor and `tf.reduce` operators like `tf.reduce_sum` to implement the feature of `tf.nn.safe_embedding_lookup_sparse` with combiner. How do you think about the difference and relationship between `SparseTensor` and `RaggedTensor` along with embedding?

tanzhenyu:
> > That's a very good question. We're currently gathering use cases for supporting sparse with Embedding layer, `Lookup` is definitely the most important one, but if you have other use cases, would you mind sharing with us?
> 
> Yes. Except `Lookup`, `Hashing` and `Bucketize` which support sparse in and sparse out should be use cases. In [ElasticDL](https://github.com/sql-machine-learning/elasticdl), we are developing some preprocessing layers like [RoundIdentity](https://github.com/sql-machine-learning/elasticdl/pull/1842). What's more, we sometimes need to convert the inputs to sparse tensor ignoring missing value and transform the sparse tensor using Keras preprocessing layers. The detail is in the ElasticDL [issue](https://github.com/sql-machine-learning/elasticdl/issues/1844).
> 
> In TF 2.1.0, I find that the `tf.keras.layers.Embedding` has supported `RaggedTensor` and output `RaggedTensor` with embedding vectors. In my opinion, the function of `RaggedTensor` is the same as `SparseTensor`. And we can use `tf.keras.layers.Embedding` with ragged tensor and `tf.reduce` operators like `tf.reduce_sum` to implement the feature of `tf.nn.safe_embedding_lookup_sparse` with combiner. Except `tf.keras.layers.Embedding`, the `tf.keras.layers.Concatenate` also support `RaggedTensor` but not `SparseTensor`. How do you think about the difference and relationship between `SparseTensor` and `RaggedTensor` along with embedding?

Sorry for the delay. 
1) Yep I think we should support sparse inputs for Embedding layer.
2) The major difference between SparseTensor and RaggedTensor is the former can represent things with missing value, i.e.:
[[0, N/A,   1  ]
 [2,    1,   N/A]]
which is usually used in structured data, while the latter cannot do that, but instead represent things with variable length, which is usually used in text.

workingloong:
> 2\. RaggedTensor

Thanks for your explanation. 
Please notify us if there is design or PR to support sparse inputs for Embedding layer. And I would like to contribute to it.

tanzhenyu:
> > 2. RaggedTensor
> 
> Thanks for your explanation.
> Please notify us if there is design or PR to support sparse inputs for Embedding layer. And I would like to contribute to it.

Of course, contribution is welcome! Can you make a PR for it?
Do note that using tf.nn.safe_embedding_lookup_sparse will probably be better than tf.nn.embedding_lookup_sparse, I have tried this when input rank is (None, None), it seems the former will return (None, None, Embedding_Dimension) while the latter will return (None, None)

ematejska:
Looks like cannot update the pull request here anymore because the original branch/repo is not available.  Will do the merge and approval in https://github.com/tensorflow/community/pull/209.  Closing this one.  Please see https://github.com/tensorflow/community/pull/209 for the notes from the design review in the RFC doc.

PR #190, opened 2019-12-18
RFC: precise tfjs

munrocket:
# Precise TFJS

| Status        | Proposed                                           |
| **RFC #**     | [190](https://github.com/tensorflow/community/issues/190)|
| **Author(s)** | Munrocket (munrocket@pm.me)                          |
| **Sponsor**   | *become my sponsor*                                  |
| **Updated**   | 2019-12-18                                           |

## Objective

TFJS version with emulated double precision that use double-word technique on GPU


munrocket:
@googlebot I signed it!

brijk7:
Hi @munrocket,
Please reach out on the tfjs@tensorflow.org discussion group to describe your proposal and find a sponsor in the TF.js team. I would also encourage you to make the RFC sections a lot more detailed once you have a sponsor. 

thanks
brijesh

munrocket:
Thanks @brijk7, I will try.

munrocket:
I guess the bigger problem was with loosing precision in RGBA packing algorithm that pack float32 -> RGBA and unpack RGBA back. With this problem we can't quickly calculate GPGPU on WebGL backend and make a polyfill to WEBGL_color_buffer_float extension. I am going to close this RFC, but if somebody need float64 support it is a good opportunity.

ematejska:
Hi @munrocket.  Were you able to reach out to get a sponsor?

munrocket:
@nsthorat do we lose some precision with packing algorithm in float16? Theoretically it is possible to extend single precision to double, but if we already losing it in float16 this trick will be useless.

nsthorat:
@munrocket packing itself shouldn't lose precision, but float16 itself will lose precision over float32 (fewer bits to represent the same value).

munrocket:
Yep, but only 6 last bit, so technically it will be float24.

Performance also can be an issue, because we need bind two textures in GPGPU chain-buffer. According to similar implementation in 10 times slower than float16.

If it will be practically useful it is a possible solution.

ematejska:
This pull request will be closed for now  in a week if a sponsor is not found.  We encourage you to find a sponsor or re-open the pull request when a sponsor is found.

munrocket:
Ty.

PR #192, opened 2019-12-23
Update SIG IO R package release instruction

terrytangyuan:
Additional details that need to be checked before release.


terrytangyuan:
cc: @tensorflow/sig-io-maintainers

PR #193, opened 2020-01-07
RFC: tf.data Snapshot

frankchn:
_Comment period is open till 1/20/2020._

# tf.data Snapshot

| Status        | Accepted                                                |
| **RFC #**     | [193](https://github.com/tensorflow/community/pull/193) |
| **Author(s)** | Frank Chen (frankchn@google.com), Rohan Jain            |
|               | (rohanj@google.com)                                     |
| **Sponsor**   | Jiri Simsa (jsimsa@google.com)                          |
| **Updated**   | 2020-02-10                                              |

## Objective

With ever faster accelerators available in Cloud and hyperparameter tuning
consuming larger chunks of accelerator time, TensorFlow users are increasingly
finding that they don’t have enough CPU resources to keep up with these
accelerators, leaving valuable accelerator resources idle.

To alleviate this problem, we are proposing a `snapshot` API within `tf.data`,
to allow users to transparently persist the output of their preprocessing
pipeline to disk, and materialize the pre-processed data on a different training
run.

This API enables repeated preprocessing steps to be consolidated, and allowing
re-use of already processed data, trading off disk storage and network bandwidth
for freeing up more valuable CPU resources and accelerator compute time.

The transformation is similar to the existing `tf.data.Dataset.cache`, with the key difference is being that, unlike `cache`, `snapshot` is intended to re-used across different executions of the same input pipelines.

### DIFF
--- a/rfcs/20200107-tf-data-snapshot.md
+++ b/rfcs/20200107-tf-data-snapshot.md
@@ -70,14 +70,10 @@ We are proposing the following API for the snapshot transformation.
 ```python
 def snapshot(path,
              compression=None,
-             reader_path_prefix=None,
-             writer_path_prefix=None,
              shard_size_bytes=None,
              pending_snapshot_expiry_seconds=None,
              num_reader_threads=None,
-             reader_buffer_size=None,
              num_writer_threads=None,
-             writer_buffer_size=None,
              shuffle_on_read=None,
              shuffle_seed=None,
              mode=None,
@@ -88,38 +84,26 @@ def snapshot(path,
 1.  `path`: Required. A directory where we want to save our snapshots and/or
     read from a previously saved snapshot.
 
-2.  `compression`: Optional. The type of compression to apply to the snapshot
+1.  `compression`: Optional. The type of compression to apply to the snapshot
     written to disk. This will support `GZIP`, `SNAPPY` or None. Defaults to
     AUTO.
 
-3.  `reader_path_prefix`: Optional. A prefix to add to the path when reading
-    from snapshots. This is useful for filesystems where configuration is passed
-    in through the path. Defaults to None.
-
-4.  `writer_path_prefix`: Optional. A prefix to add to the path when writing to
-    snapshots. This is useful for filesystems where configuration is passed in
-    through the path. Defaults to None.
-
-5.  `shard_size_bytes`: Optional. The maximum size of each data file to be
+1.  `shard_size_bytes`: Optional. The maximum size of each data file to be
     written by the snapshot dataset op. Defaults to AUTO.
 
-6.  `pending_snapshot_expiry_seconds`: Optional. How long to wait (in seconds)
+1.  `pending_snapshot_expiry_seconds`: Optional. How long to wait (in seconds)
     before the snapshot op considers a previously unfinished snapshot to be
     stale and starts writing a snapshot from scratch again. Defaults to 86400
     seconds (1 day).
 
-7.  `num_reader_threads`: Optional. Number of threads to parallelize reading
+1.  `num_reader_threads`: Optional. Number of threads to parallelize reading
     from snapshot. Especially useful if compression is turned on since the
     decompression operation tends to be intensive. If > 1, then
     this might introduce non-determinism i.e. the order in which the elements
     are read from the snapshot are different from the order they're written.
     Defaults to AUTO. 
 
-8.  `reader_buffer_size`: Optional. Maximum number of elements we can prefetch
-    reading from the snapshot. Increasing this might improve
-    performance but will increase memory consumption. Defaults to AUTO.
-
-9.  `num_writer_threads`: Optional. Number of threads to parallelize writing
+1.  `num_writer_threads`: Optional. Number of threads to parallelize writing
     from snapshot. We'll open up `num_writer_threads` files and write to them in
     parallel. Especially useful if compression is turned on since the
     compression operation tends to be intensive. If > 1, then
@@ -127,21 +111,17 @@ def snapshot(path,
     are read from the upstream iterator are different from the order they're
     written. Defaults to AUTO. 
 
-10. `writer_buffer_size`: Optional. Maximum number of pipeline elements to fill
-    up the buffer before writing them out using `num_writer_threads`. Defaults
-    to AUTO.
-
-11. `shuffle_on_read`: Optional. If this is True, then snapshot randomizes the
+1.  `shuffle_on_read`: Optional. If this is True, then snapshot randomizes the
     order in which the snapshot files are read back. This emulates shuffling
     of the input files during a training run (e.g. when `Dataset.list_files` 
     is called with `shuffle` turned on). Defaults to False.
 
-12. `shuffle_seed`: Optional. If shuffle_seed is set, the random number
+1.  `shuffle_seed`: Optional. If shuffle_seed is set, the random number
     generator used for shuffling (when `shuffle_on_read` is turned on) is seeded
     by the given seed. Otherwise, it is seeded by a random seed that differs for
     every run.
 
-13. `mode`: Optional. The mode at which snapshot should operate. Valid options
+1.  `mode`: Optional. The mode at which snapshot should operate. Valid options
     are `auto`, `read`, `write`, and `passthrough`. The default mode is `auto`,
     where the snapshot op will automatically determine what mode to operate in.
 
@@ -150,24 +130,33 @@ def snapshot(path,
         materialization currently exists. In other words, we enter the **WRITE**
         state immediately.
 
-    2.  `read` mode forces the snapshot transformation to read from the latest
+    1.  `read` mode forces the snapshot transformation to read from the latest
         version of the materialization on disk, regardless of whether the data
         stored on disk is complete and valid. In other words, we enter the
         **READ** state immediately.
 
-    3.  `passthrough` mode turns the snapshot transformation into a no-op. In
+    1.  `passthrough` mode turns the snapshot transformation into a no-op. In
         other words, we enter the **PASSTHROUGH** state immediately.
 
-    4.  `auto` retains the default behavior of snapshot. See the "Standard
+    1.  `auto` retains the default behavior of snapshot. See the "Standard
         Kernel Workflow" section for the default behavior.
 
-14. `snapshot_name`: Optional. If set, use the supplied string as a named
+1.  `snapshot_name`: Optional. If set, use the supplied string as a named
     snapshot name instead of introspecting the data pipeline and automatically
     generating a unique identifier for the specific data pipeline.
 
     1.  Instead of generating a new fingerprint of the input processing graph or
         and `run_id` (see the _Detailed Design_ section for details), we will
         use the `snapshot_name` to uniquely identify the snapshot.
+
+    1.  Multiple concurrent training jobs with the same "snapshot_name" may
+        result in concurrent write collisions and a potentially invalid snapshot
+        if the jobs tries to read from and then write to the metadata file at
+        exactly the same time. 
+
+        The user is expected to handle these cases and explicitly specify `mode`s
+        to ensure that only one run is set to `write` mode at any point if
+        collisions are a possibility.
 
 Note: `AUTO` options above indicates that snapshot will attempt to pick a 
 reasonable default that is suitable for most use cases. We will eventually add
@@ -195,10 +184,7 @@ select whether to train or preprocess on their own, which is not good.
 
 ### Performance Implications
 
-*   Do you expect any (speed / memory)? How will you confirm?
-*   There should be microbenchmarks. Are there?
-*   There should be end-to-end tests and benchmarks. If there are not (since
-    this is still a design), how will you track that these will be created?
+Benchmarks for this feature will be included as part of Dataset microbenchmarks.
 
 ### Dependencies
 
--- a/rfcs/20200107-tf-data-snapshot.md
+++ b/rfcs/20200107-tf-data-snapshot.md
@@ -1,0 +1,368 @@
+# tf.data Snapshot
+
+| Status        | Proposed                                                |
+| :------------ | :------------------------------------------------------ |
+| **RFC #**     | [193](https://github.com/tensorflow/community/pull/193) |
+| **Author(s)** | Frank Chen (frankchn@google.com), Rohan Jain            |
+:               : (rohanj@google.com)                                     :
+| **Sponsor**   | Jiri Simsa (jsimsa@google.com)                          |
+| **Updated**   | 2020-01-07                                              |
+
+## Objective
+
+With ever faster accelerators available in Cloud and hyperparameter tuning
+consuming larger chunks of accelerator time, TensorFlow users are increasingly
+finding that they don’t have enough CPU resources to keep up with these
+accelerators, leaving valuable accelerator resources idle.
+
+To alleviate this problem, we are proposing a `snapshot` API within `tf.data`,
+to allow users to transparently persist the output of their preprocessing
+pipeline to disk, and materialize the pre-processed data on a different training
+run.
+
+This API enables repeated preprocessing steps to be consolidated, and allowing
+re-use of already processed data, trading off disk storage and network bandwidth
+for freeing up more valuable CPU resources and accelerator compute time.
+
+## Motivation
+
+Large TensorFlow users have indicated that they have complicated input
+processing pipelines which saturate their CPUs before saturating their
+accelerators (TPUs in particular). Since they often experiment with
+hyperparameter tuning or tweaks to existing model without affecting their input
+pipeline, they are asking for ways to avoid similar repeated preprocessing of
+data by either saving a dataset or caching it to disk.
+
+## User Benefit
+
+Users will be able to transparently persist partially or fully processed data
+from `tf.data` input pipelines to disk or Cloud storage systems, and materialize
+the pre-processed data during subsequent runs from the same pipeline. This will
+cut down on the input pipeline processing overheads during second and subsequent
+runs.
+
+## Design Proposal
+
+We propose that we add a new `snapshot` transformation to tf.data. To illustrate
+the usage of the transformation, we can start with some sample code:
+
+```python
+dataset = Dataset.list_files("/raw/data/*").shard(num_workers, i)
+dataset = dataset.parallel_interleave(TFRecordDataset)
+dataset = dataset.map(my_preprocessing_fn)
+dataset = dataset.apply(tf.data.snapshot("/saved/data", options...))
+dataset = dataset.repeat()
+
+model = ...
+model.fit(dataset)
+```
+
+As we can see, the end user simply has to add this transformation in order to
+use this functionality. In essence, the transformation is similar to the
+existing `tf.data.Dataset.cache`, with the key difference is being that, unlike
+`cache`, `snapshot` is intended to re-used across different executions of the
+same input pipelines.
+
+### Proposed API
+
+We are proposing the following API for the snapshot transformation.
+
+```python
+def snapshot(path,
+             compression=None,
+             reader_path_prefix=None,
+             writer_path_prefix=None,
+             shard_size_bytes=None,
+             pending_snapshot_expiry_seconds=None,
+             num_reader_threads=None,
+             reader_buffer_size=None,
+             num_writer_threads=None,
+             writer_buffer_size=None,
+             shuffle_on_read=None,
+             shuffle_seed=None,
+             mode=None,
+             snapshot_name=None):
+  pass  # Implementation goes here.
+```
+
+1.  `path`: Required. A directory where we want to save our snapshots and/or
+    read from a previously saved snapshot.
+
+2.  `compression`: Optional. The type of compression to apply to the snapshot
+    written to disk. This will support `GZIP`, `SNAPPY` or None. Defaults to
+    None.
+
+3.  `reader_path_prefix`: Optional. A prefix to add to the path when reading
+    from snapshots. This is useful for filesystems where configuration is passed
+    in through the path. Defaults to None.
+
+4.  `writer_path_prefix`: Optional. A prefix to add to the path when writing to
+    snapshots. This is useful for filesystems where configuration is passed in
+    through the path. Defaults to None.
+
+5.  `shard_size_bytes`: Optional. The maximum size of each data file to be
+    written by the snapshot dataset op. Defaults to 10 GiB.
+
+6.  `pending_snapshot_expiry_seconds`: Optional. How long to wait (in seconds)
+    before the snapshot op considers a previously unfinished snapshot to be
+    stale and starts writing a snapshot from scratch again. Defaults to 86400
+    seconds (1 day).
+
+7.  `num_reader_threads`: Optional. Number of threads to parallelize reading
+    from snapshot. Especially useful if compression is turned on since the
+    decompression operation tends to be intensive. Defaults to 1. If > 1, then
+    this might introduce non-determinism i.e. the order in which the elements
+    are read from the snapshot are different from the order they're written.
+
+8.  `reader_buffer_size`: Optional. Maximum number of elements we can prefetch
+    reading from the snapshot. Defaults to 1. Increasing this might improve
+    performance but will increase memory consumption.
+
+9.  `num_writer_threads`: Optional. Number of threads to parallelize writing
+    from snapshot. We'll open up `num_writer_threads` files and write to them in
+    parallel. Especially useful if compression is turned on since the
+    compression operation tends to be intensive. Defaults to 1. If > 1, then
+    this might introduce non-determinism i.e. the order in which the elements
+    are read from the upstream iterator are different from the order they're
+    written.
+
+10. `writer_buffer_size`: Optional. Maximum number of pipeline elements to fill
+    up the buffer before writing them out using `num_writer_threads`.
+
+11. `shuffle_on_read`: Optional. If this is True, then the order in which
+    examples are produced when reading from a snapshot will be random. Defaults
+    to False.
+
+12. `shuffle_seed`: Optional. If shuffle_seed is set, the random number
+    generator used for shuffling (when `shuffle_on_read` is turned on) is seeded
+    by the given seed. Otherwise, it is seeded by a random seed that differs for
+    every run.
+
+13. `mode`: Optional. The mode at which snapshot should operate. Valid options
+    are `auto`, `read`, `write`, and `passthrough`. The default mode is `auto`,
+    where the snapshot op will automatically determine what mode to operate in.
+
+    1.  `write` mode forces the snapshot transformation to write a new
+        materialization to disk, regardless of whether a complete and valid
+        materialization currently exists. In other words, we enter the **WRITE**
+        state immediately.
+
+    2.  `read` mode forces the snapshot transformation to read from the latest
+        version of the materialization on disk, regardless of whether the data
+        stored on disk is complete and valid. In other words, we enter the
+        **READ** state immediately.
+
+    3.  `passthrough` mode turns the snapshot transformation into a no-op. In
+        other words, we enter the **PASSTHROUGH** state immediately.
+
+    4.  `auto` retains the default behavior of snapshot. See the "Standard
+        Kernel Workflow" section for the default behavior.
+
+14. `snapshot_name`: Optional. If set, use the supplied string as a named
+    snapshot name instead of introspecting the data pipeline and automatically
+    generating a unique identifier for the specific data pipeline.
+
+    1.  Instead of generating a new fingerprint of the input processing graph or
+        and `run_id` (see the _Detailed Design_ section for details), we will
+        use the `snapshot_name` to uniquely identify the snapshot.
+
+### External API Guarantees
+
+Externally, we guarantee that snapshots written by a particular version of
+TensorFlow will be readable by that specific version of TensorFlow. Eventually,
+we can also guarantee that snapshots written will be readable by all future
+versions of TensorFlow.
+
+We are not currently handling the case where workers do not go through the
+entire training set at least once.
+
+### Alternatives Considered
+
+An alternative proposal for an API would be `save()` and `load()`, where the
+saving and loading of the input pipeline would be made more explicit, avoiding
+some of the logic needed in determining whether to snapshot or read from a
+snapshot of a model.
+
+The downside here would be that the user would have to split the preprocessing
+and training into potentially different files, and users would be forced to
+select whether to train or preprocess on their own, which is not good.
+
+### Performance Implications
+
+*   Do you expect any (speed / memory)? How will you confirm?
+*   There should be microbenchmarks. Are there?
+*   There should be end-to-end tests and benchmarks. If there are not (since
+    this is still a design), how will you track that these will be created?
+
+### Dependencies
+
+No new dependencies will be introduced as part of this project to TensorFlow.
+Dependent projects may be able to use this additional op, but there should be no
+significant changes otherwise.
+
+### Engineering Impact
+
+Binary sizes increases slightly with the inclusion of this new op, and this code
+will be maintained by the `tf.data` team.
+
+### Platforms and Environments
+
+This op will work on all TensorFlow-supported platforms. We do not anticipate
+this to work on embedded systems as it is not useful in resource-constrained
+environments.
+
+### Best Practices, Tutorials and Examples
+
+A user guide for snapshot will be published to guide new users in using this
+feature.
+
+### Compatibility
+
+This introduces a new op, which will impact future backwards compatibility.
+
+### User Impact
+
+A new python function and a new op are the only user-facing changes visible.
+
+## Detailed Design
+
+### Implementation Assumptions
+
+The following implementation is based on the following assumptions that define
+the MVP this is designed for:
+
+1.  We assume that at least for one pipeline run, you can go through the entire
+    training dataset and be able to store that data on disk. Otherwise, a
+    snapshot will never get created.
+
+2.  In case there are multiple workers and the dataset is sharded across
+    workers, we assume that the number of workers remains the same from one run
+    to another. If the number changes, we’ll trigger another snapshot.
+
+3.  Any `repeat`s in the dataset should be moved to after the `snapshot` op, to
+    avoid writing large (or infinite) amounts of data during a snapshot writing
+    run.
+
+### New `SnapshotDatasetOp`
+
+To implement the transformation, we are introducing a new `SnapshotDatasetOp`
+dataset kernel that will implement all of the functionality in TensorFlow C++.
+Python code is mostly glue code to pass relevant parameters into the op kernel.
+
+### Internal Directory / File Structure
+
+Given a user directory path (e.g. `/path/to/snapshot`), the directory will look
+like:
+
+*   /path/to/snapshot
+    *   `fingerprint`/
+        *   snapshot.metadata
+        *   `run-id`/
+            *   0000000.snapshot
+            *   0000001.snapshot
+
+The `fingerprint` is a hash of the input processing graph. The `run-id` is
+unique training run ID generated.
+
+### Standard Kernel Workflow
+
+_Note: This is an implementation detail, and may change in the future. This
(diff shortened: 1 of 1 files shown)


rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+## Design Proposal
+
+We propose that we add a new `snapshot` transformation to tf.data. To illustrate
+the usage of the transformation, we can start with some sample code:
+
+```python
+dataset = Dataset.list_files("/raw/data/*").shard(num_workers, i)
+dataset = dataset.parallel_interleave(TFRecordDataset)
+dataset = dataset.map(my_preprocessing_fn)
+dataset = dataset.apply(tf.data.snapshot("/saved/data", options...))
```
jsimsa: either `dataset = dataset.apply(tf.data.experimental.snapshot(...))` or `dataset = dataset.snapshot(...)` (I would prefer the former before providing backwards compatibility guarantee for the API).

byronyi:
Thanks Frank for this RFC. We observe similar performance benefits in caching preprocessed data in multi-epoch training. 

Since we already have a ds.cache API that support file system interface, what is the unique value proposition to have another ds.snapshot API?

frankchn:
> Thanks Frank for this RFC. We observe similar performance benefits in caching preprocessed data in multi-epoch training.
> 
> Since we already have a ds.cache API that support file system interface, what is the unique value proposition to have another ds.snapshot API?
> 
> EDIT: saw that in the doc, might worth to mention that in the PR summary.

Thanks for your comments, and added the motivation and differences to `ds.cache` in the PR summary.

AakashKumarNain:
How is this different from processing data, saving them in TFRecords format on the disk and reading it later?

frankchn:
> How is this different from processing data, saving them in TFRecords format on the disk and reading it later?

Thanks for your question! You can definitely manually replicate what we are doing here by just running an input pipeline and saving the processed data to other TFRecord files.

That said, there are a couple of advantages of having this op over a manual process:

First, we can write to/read from snapshots automatically with a single additional op at the end of the input pipeline, without any new code necessary to write to/read from the snapshots. 

Second, tf.data snapshot can detect changes in the pipeline (e.g. changing JPEG image decode sizes or pointing to a new set of files) through input pipeline fingerprinting and automatically create new snapshot files with no further intervention. This minimizes potential errors when you are experimenting and changing input pipelines.

Third, the proposed file format we have is somewhat more efficient (both in terms of the file format itself, and the ability to enable gzip/snappy compression on the file level). We also try to optimize the read/write paths as much as possible, so at the limit we believe we are faster or more space-efficient (or both) than plain TFRecord reading/writing.

AakashKumarNain:
Thanks @frankchn for the detailed answer. I really liked this point
```
Second, tf.data snapshot can detect changes in the pipeline (e.g. changing JPEG image decode sizes or pointing to a new set of files) through input pipeline fingerprinting and automatically create new snapshot files with no further intervention. This minimizes potential errors when you are experimenting and changing input pipelines.
```

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+    directory is `/saved/data`, then we will attempt to enter
+    `/saved/data/f-abc123`.
+
+1.  If the snapshot directory is non-existent, empty or it doesn’t contain a
+    `metadata` file, we will enter the **WRITE** state.
+
+1.  If the snapshot directory contains a `metadata` file, we will read the
+    metadata file.
+
+    1.  The metadata file contains the following fields:
```
katsiapis: Does this suggest that metadata file needs to be modifiable after initial write? Any concurrency implications? Does it make sense to have an approach that works with immutable File artifacts?

Some enhancements that were recently made to TensorFlow Transform might be of interest here (especially their logic and rational behind them, if not the implementation).

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+The following implementation is based on the following assumptions that define
+the MVP this is designed for:
+
+1.  We assume that at least for one pipeline run, you can go through the entire
+    training dataset and be able to store that data on disk. Otherwise, a
+    snapshot will never get created.
+
+2.  In case there are multiple workers and the dataset is sharded across
+    workers, we assume that the number of workers remains the same from one run
+    to another. If the number changes, we’ll trigger another snapshot.
```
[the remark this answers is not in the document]
byronyi: I would suggest reconsidering this. Elastic training is important in our use case, to reduce resource fragmentation. Sometimes the number of workers (and along with learning rate, global batch size, etc.) is an important hyperparameter to tune by itself.

What about collective snapshot in a distributed file system? Will this be enough for easier implementation for a variable number of workers?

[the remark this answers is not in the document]
frankchn: @byronyi The current tf.data input pipeline design doesn't really allow us to coordinate between workers (e.g. there might failure cases where K worker out of N fails and the snapshot slice is invalid).

That said, I totally understand where you are coming from, and we are planning a data pipeline service (where snapshot can be integrated) so everyone would read elements off a single pipeline and having different numbers of trainers would be less of an issue. There should be a design doc RFC on that in the next few weeks so look for that.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+             shard_size_bytes=None,
+             pending_snapshot_expiry_seconds=None,
+             num_reader_threads=None,
+             reader_buffer_size=None,
+             num_writer_threads=None,
+             writer_buffer_size=None,
+             shuffle_on_read=None,
+             shuffle_seed=None,
+             mode=None,
+             snapshot_name=None):
```
karmel: This is a lot of args, especially to start off with. Is this perhaps a sign that we should be thinking of a different design pattern here?

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+    written to disk. This will support `GZIP`, `SNAPPY` or None. Defaults to
+    None.
+
+3.  `reader_path_prefix`: Optional. A prefix to add to the path when reading
+    from snapshots. This is useful for filesystems where configuration is passed
+    in through the path. Defaults to None.
+
+4.  `writer_path_prefix`: Optional. A prefix to add to the path when writing to
+    snapshots. This is useful for filesystems where configuration is passed in
+    through the path. Defaults to None.
```
karmel: Can you provide an example of 3 and 4?

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+    this might introduce non-determinism i.e. the order in which the elements
+    are read from the upstream iterator are different from the order they're
+    written.
+
+10. `writer_buffer_size`: Optional. Maximum number of pipeline elements to fill
+    up the buffer before writing them out using `num_writer_threads`.
+
+11. `shuffle_on_read`: Optional. If this is True, then the order in which
+    examples are produced when reading from a snapshot will be random. Defaults
+    to False.
```
karmel: Why is this an arg versus a call to dataset.shuffle() after the call to tf.data.snapshot? Coupling this + the following kwarg into the API spec here seems unwieldy.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+
+    1.  Instead of generating a new fingerprint of the input processing graph or
+        and `run_id` (see the _Detailed Design_ section for details), we will
+        use the `snapshot_name` to uniquely identify the snapshot.
+
+### External API Guarantees
+
+Externally, we guarantee that snapshots written by a particular version of
+TensorFlow will be readable by that specific version of TensorFlow. Eventually,
+we can also guarantee that snapshots written will be readable by all future
```
karmel: all future-- we don't guarantee that for SavedModels, and all feels strong for any API. Version N+1 is what we guarantee for SavedModel, and that may be as strong as you want to go, realistically.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+
+An alternative proposal for an API would be `save()` and `load()`, where the
+saving and loading of the input pipeline would be made more explicit, avoiding
+some of the logic needed in determining whether to snapshot or read from a
+snapshot of a model.
+
+The downside here would be that the user would have to split the preprocessing
+and training into potentially different files, and users would be forced to
+select whether to train or preprocess on their own, which is not good.
+
```
karmel: I feel like save/load fits better into existing TF paradigms, and makes the behavior more explicit. Looking at the pipeline above, it's hard to tell what is going to be run on an arbitrary run of the pipeline. You could imagine an individual pipeline doing a load first, then if no file to load, do the preproc/save. That's explicit and doesn't require two separate runners.

rfcs/20200107-tf-data-snapshot.md
karmel: How will we trigger another snapshot exactly? Does that imply attempting a read of the data, failing, and instead re-snapshotting?

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+    training dataset and be able to store that data on disk. Otherwise, a
+    snapshot will never get created.
+
+2.  In case there are multiple workers and the dataset is sharded across
+    workers, we assume that the number of workers remains the same from one run
+    to another. If the number changes, we’ll trigger another snapshot.
+
+3.  Any `repeat`s in the dataset should be moved to after the `snapshot` op, to
+    avoid writing large (or infinite) amounts of data during a snapshot writing
+    run.
```
karmel: Is this to be guaranteed (eg, raise an error)? Or is this more of a best practice that we'll have to document?

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+
+*   /path/to/snapshot
+    *   `fingerprint`/
+        *   snapshot.metadata
+        *   `run-id`/
+            *   0000000.snapshot
+            *   0000001.snapshot
+
+The `fingerprint` is a hash of the input processing graph. The `run-id` is
+unique training run ID generated.
```
karmel: How does this align/not align with existing ckpt + saved model semantics? Is there more we can reuse here?

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+_Note: This is an implementation detail, and may change in the future. This
+should not be relied upon except as a reference to the current implementation._
+
+By default, the `snapshot` operation will, upon startup, make a determination
+using the following algorithm as to whether the operation should be in the
+WRITE, PASSTHROUGH, or READ state.
+
+1.  We will compute a graph fingerprint containing all the information from the
+    Dataset preprocessing graph before the `snapshot` op. We’ll use the
+    `AsGraphDefInternal` method on DatasetBase for this.
```
karmel: For those of us less familiar-- does this work the same way in an Eager context? The graph is built regardless? On the second run in Eager, do the ops execute before we know to check for a snapshot? Presumably not, but-- how do we know to not execute anything and instead read while executing each individual preproc line?

rfcs/20200107-tf-data-snapshot.md
rohan100jain: I think right now .apply(...) is the only way to support an experimental tf.data transformation. So the plan would be to expose this as apply and then in a few months, we can promote this to a dataset.snapshot API.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+Large TensorFlow users have indicated that they have complicated input
+processing pipelines which saturate their CPUs before saturating their
+accelerators (TPUs in particular). Since they often experiment with
+hyperparameter tuning or tweaks to existing model without affecting their input
+pipeline, they are asking for ways to avoid similar repeated preprocessing of
+data by either saving a dataset or caching it to disk.
+
+## User Benefit
+
+Users will be able to transparently persist partially or fully processed data
```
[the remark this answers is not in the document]
frankchn: Yup, this is a way to serializing tf.data datasets to disk then reading them back. However, the API design is geared towards saving preprocessing time right now.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+             snapshot_name=None):
+  pass  # Implementation goes here.
+```
+
+1.  `path`: Required. A directory where we want to save our snapshots and/or
+    read from a previously saved snapshot.
+
+2.  `compression`: Optional. The type of compression to apply to the snapshot
+    written to disk. This will support `GZIP`, `SNAPPY` or None. Defaults to
+    None.
```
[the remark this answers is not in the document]
frankchn: Good point. We can have the default option be AUTO here and elsewhere, and gradually add tf.data autotuning to snapshots after (autotuning right now starts a few versions of the pipelines, which is not ideal given that we are writing to the file system).

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+    from snapshots. This is useful for filesystems where configuration is passed
+    in through the path. Defaults to None.
+
+4.  `writer_path_prefix`: Optional. A prefix to add to the path when writing to
+    snapshots. This is useful for filesystems where configuration is passed in
+    through the path. Defaults to None.
+
+5.  `shard_size_bytes`: Optional. The maximum size of each data file to be
+    written by the snapshot dataset op. Defaults to 10 GiB.
+
```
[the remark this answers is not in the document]
frankchn: Changed.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+6.  `pending_snapshot_expiry_seconds`: Optional. How long to wait (in seconds)
+    before the snapshot op considers a previously unfinished snapshot to be
+    stale and starts writing a snapshot from scratch again. Defaults to 86400
+    seconds (1 day).
+
+7.  `num_reader_threads`: Optional. Number of threads to parallelize reading
+    from snapshot. Especially useful if compression is turned on since the
+    decompression operation tends to be intensive. Defaults to 1. If > 1, then
+    this might introduce non-determinism i.e. the order in which the elements
+    are read from the snapshot are different from the order they're written.
```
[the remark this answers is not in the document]
frankchn: Changed.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+
+8.  `reader_buffer_size`: Optional. Maximum number of elements we can prefetch
+    reading from the snapshot. Defaults to 1. Increasing this might improve
```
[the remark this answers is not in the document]
frankchn: Changed

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+    compression operation tends to be intensive. Defaults to 1. If > 1, then
+    this might introduce non-determinism i.e. the order in which the elements
+    are read from the upstream iterator are different from the order they're
+    written.
+
+10. `writer_buffer_size`: Optional. Maximum number of pipeline elements to fill
+    up the buffer before writing them out using `num_writer_threads`.
+
+11. `shuffle_on_read`: Optional. If this is True, then the order in which
+    examples are produced when reading from a snapshot will be random. Defaults
```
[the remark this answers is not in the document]
frankchn: Clarified. The behavior emulates `Dataset.list_files(shuffle=True)` by randomizing the order in which the snapshot files are read back.

rfcs/20200107-tf-data-snapshot.md
frankchn: This emulates the behavior when `Dataset.list_files(shuffle=True)` and is orthogonal to using `dataset.shuffle()` after the snapshot op. Clarified the behavior of this flag.

rfcs/20200107-tf-data-snapshot.md
frankchn: This is currently a best practice. There will be a snapshot best practices guide published later.

rfcs/20200107-tf-data-snapshot.md
frankchn: Makes sense. Removed the guarantees about future stuff.

rfcs/20200107-tf-data-snapshot.md
frankchn: I don't think SavedModel supports comparisons between graphs (or in this case, parts of a graph), which is our main use case here. In general, I am not sure TensorFlow can guarantee to produce the exact same graph (i.e. with the same node names, function names, etc...), so we are utilizing the `HashGraph` function (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/kernels/data/dataset_utils.cc#L733), which computes a fingerprint while ignoring node names and other things that may vary run by run.

rfcs/20200107-tf-data-snapshot.md
frankchn: An immutable metadata file definitely makes sense. We will introduce a `metadata.final` file to indicate when snapshot has finished writing.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,368 @@
...
+    the snapshot.metadata file to determine whether it contains the same
+    training run ID.
+
+    1.  If it does, we set the complete bit to true to finalize the directory.
+    1.  If it does not, it means that someone else is concurrently writing the
+        snapshot and we lost the race to them. We delete all data in the
+        training run directory.
+
+For the current implementation, we will store the data in chunked TFRecord
+files. Eventually we may move to other more higher performance data stores or
```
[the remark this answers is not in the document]
frankchn: @katsiapis Sorry, can you clarify what sink you are referring to?

rfcs/20200107-tf-data-snapshot.md
frankchn: If the number of workers change, then (at least in the case of sharded TPU training) the `num_shards` parameter to `Dataset.shard` will change too. This results in a different graph fingerprint for each worker, and we will go into WRITE mode automatically. If there are no shards and all the workers use exactly the same input pipeline, then this won't happen. Will add something here to clarify.

rfcs/20200107-tf-data-snapshot.md
frankchn: There are internal use cases for this (e.g. readahead).

rfcs/20200107-tf-data-snapshot.md
frankchn: Yeah, I don't particularly like how many args there are here either. Perhaps reader/writer threads and buffer sizes can be collapsed into just threads and buffer (cc @rohan100jain)

rfcs/20200107-tf-data-snapshot.md
rohan100jain: Right now the use cases are primarily internal (readahead etc.) and I couldn't find any public filesystems that passed configuration through prefixes. Should we consider an internal version of the op instead then?

rfcs/20200107-tf-data-snapshot.md
rohan100jain: The issue with that is the logic that determines whether the input pipeline we want to load or not is not as simple as whether a file exists (the control aspect of the op). Following your train of thought, I think the better alternative would be splitting it up as reader, writer and control as separate ops (which as I indicated above should be the implementation for the next iteration of the op).

rfcs/20200107-tf-data-snapshot.md
rohan100jain: tf.data always builds a dataset graph for execution, whether its eager or not. So it doesn't matter much whether we're in eager mode or not [to be precise: the iterator gets created eagerly which takes in a dataset variant that contains the entire dataset graph to be executed which is what we fingerprint]

rfcs/20200107-tf-data-snapshot.md
frankchn: I think this design was chosen so that the user doesn't have to worry about changing their existing pipelines much at all, and can benefit from this just by dropping snapshot in. 

With `save()` and `load()`, users will have to manage saving and loading by themselves and this potentially introduces errors (e.g. users may want accidentally load an old version of a snapshot without realizing it).

That said, perhaps we can re-use the C++ code that implements this into more generic Dataset.save and Dataset.load ops for users who want that sort of control?

frankchn: Agree with @rohan100jain that for the next internal implementation we can expose save/load/control where appropriate while retaining this python interface for existing users who might want a simple solution.

rfcs/20200107-tf-data-snapshot.md
karmel: That is true, though I meant the file structure itself and the serialized format, not strictly the fingerprint.

rfcs/20200107-tf-data-snapshot.md
rohan100jain: Would the AUTO as default help here? I think based on our experience, we can come up with reasonable defaults reducing the amount of cognitive overhead the users will have to endure figuring out what to set.

We can also remove the buffer_size option and just set it to 2 * num_threads.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,387 @@
...
+dataset = dataset.repeat()
+
+model = ...
+model.fit(dataset)
+```
+
+As we can see, the end user simply has to add this transformation in order to
+use this functionality. In essence, the transformation is similar to the
+existing `tf.data.Dataset.cache`, with the key difference is being that, unlike
+`cache`, `snapshot` is intended to re-used across different executions of the
```
alextp: Does cache provide any advantages over snapshot?

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,387 @@
...
+             compression=None,
+             reader_path_prefix=None,
+             writer_path_prefix=None,
+             shard_size_bytes=None,
+             pending_snapshot_expiry_seconds=None,
+             num_reader_threads=None,
+             reader_buffer_size=None,
+             num_writer_threads=None,
+             writer_buffer_size=None,
+             shuffle_on_read=None,
```
alextp: Why does snapshotting include its own shuffling?

How does the snapshot shuffling interact with the shuffling performed in existing tf.data transformations like shuffle?

The way these options are presented now mean you have different behavior when loading from a snapshot than you had when not loading from a snapshot, which means the existing shuffling behavior is not guaranteed even if there is no shuffle_on_read (as reading from a snapshot cannot simulate shuffling files).

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,387 @@
...
+```
+
+1.  `path`: Required. A directory where we want to save our snapshots and/or
+    read from a previously saved snapshot.
+
+2.  `compression`: Optional. The type of compression to apply to the snapshot
+    written to disk. This will support `GZIP`, `SNAPPY` or None. Defaults to
+    AUTO.
+
+3.  `reader_path_prefix`: Optional. A prefix to add to the path when reading
```
alextp: This and writer_path_prefix feel like very specific hacks I would not like to see persisted throughout the API.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,387 @@
...
+    in through the path. Defaults to None.
+
+4.  `writer_path_prefix`: Optional. A prefix to add to the path when writing to
+    snapshots. This is useful for filesystems where configuration is passed in
+    through the path. Defaults to None.
+
+5.  `shard_size_bytes`: Optional. The maximum size of each data file to be
+    written by the snapshot dataset op. Defaults to AUTO.
+
+6.  `pending_snapshot_expiry_seconds`: Optional. How long to wait (in seconds)
```
alextp: Why do we need this?

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,387 @@
...
+        stored on disk is complete and valid. In other words, we enter the
+        **READ** state immediately.
+
+    3.  `passthrough` mode turns the snapshot transformation into a no-op. In
+        other words, we enter the **PASSTHROUGH** state immediately.
+
+    4.  `auto` retains the default behavior of snapshot. See the "Standard
+        Kernel Workflow" section for the default behavior.
+
+14. `snapshot_name`: Optional. If set, use the supplied string as a named
```
alextp: This needs to clarify what happens if two snapshots are named the same. Can they collide?

rfcs/20200107-tf-data-snapshot.md
alextp: We should consider that the semantics of save/load are much easier to explain to people specially when you consider the interaction of shuffle_on_read with other options earlier in the input pipeline.

rfcs/20200107-tf-data-snapshot.md
frankchn: After discussion, going to remove the buffer_size option.

rfcs/20200107-tf-data-snapshot.md
frankchn: Cache can be in memory, which is useful for testing and small datasets.

rfcs/20200107-tf-data-snapshot.md
frankchn: Good point, removed.

rfcs/20200107-tf-data-snapshot.md
frankchn: This is there so that if a pipeline in write mode crashes halfway without finalizing the snapshot, future runs of the pipeline can start a new writing process rather be stuck in passthrough mode indefinitely.

rfcs/20200107-tf-data-snapshot.md
frankchn: Added clarification about the possibility of collisions.

rfcs/20200107-tf-data-snapshot.md
byronyi: Cache can be on-disk as well. Its lifetime is more confined (to a specific job), and no worries about crash-safe semantics. Lifetime of snapshots could be unbounded, and I guess that's why we need explicit expiry.

rfcs/20200107-tf-data-snapshot.md
byronyi: > Would the AUTO as default help here? I think based on our experience, we can come up with reasonable defaults reducing the amount of cognitive overhead the users will have to endure figuring out what to set.
> 
> We can also remove the buffer_size option and just set it to 2 * num_threads.

Would you mind to leave it there? Buffer size is an important parameter to tune for distributed FS such as S3 or HDFS. The appropriate buffer size should be bandwidth delay product, not 2*num_threads. This option will be relevant if both the delay of accessing storage and its throughput are high.

frankchn: The buffer size configuration here is for the reading and writing thread buffers (basically the same effect as having `Dataset.prefetch(1)`) rather than file system buffers. For the buffers you are talking about, we offer configurations at the file system levels (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/platform/cloud/gcs_file_system.cc#L67) that would be more general.

byronyi: Thanks for the clarification. I am just wondering where can I pass system buffer configuration with the proposed snapshot API (and in general APIs that interact with external storage such as cache).

rfcs/20200107-tf-data-snapshot.md
rohan100jain: During earlier iterations of developing this, we considered save / load style APIs but to make that usable is quite challenging (i.e. the figuring out whether to save a new snapshot or to load). The current API serves a very concrete use case (for some significant internal users) and we feel it makes sense to expose this API to serve that use case. 

Your concerns about shuffle_on_read etc. are valid and we'll address them by allowing users to specify a reader_fn.

rfcs/20200107-tf-data-snapshot.md
rohan100jain: Thats a good point Karmel and we considered this while first developing. Here is why we ended up choosing something of our own - 

The checkpointing API is for saving the state of the model and here we're trying to store the entire output of an input pipeline and so API wise there isn't much alignment.

On the implementation side, Checkpointing uses BundleReader / BundleWriter which are designed for key / value accesses to tensors whereas here we want to sequentially read off tensors as soon as possible. We tried Bundle Reader / Writer at first but abandoned it for performance reasons.

rfcs/20200107-tf-data-snapshot.md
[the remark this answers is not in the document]
rohan100jain: I guess Gus is referring to allowing users to customize the writing data format etc. via some kind of sink configuration. I think as of now, we don't plan to allow for that level of customization because right now its good to have our own reader / writer that is optimized for performance. Its definitely possible that in the future we allow users to customize this though but I think that won't happen in the near future. Will keep this in mind though.

rfcs/20200107-tf-data-snapshot.md
frankchn: @alextp @rohan100jain I have updated the design doc to add a new `reader_fn` and remove existing reader parameters. PTAL, thanks!

rfcs/20200107-tf-data-snapshot.md
karmel: At the risk of sacrilege-- can we improve the performance of Bundle Reader / Writer instead of just building a new way of doing this?

byronyi: @karmel Same question here. I have been working on a prototype to enable external shuffled dataset with tensor bundle as its underlying storage library. Up till now I could get good performance, e.g. 4.5k image/s with 64 reader thread on a slow HDFS. 

All I did was adding the LRU cache for SSTable index (copied from the LevelDB project; apparently was removed in TF).

byronyi: Also ping @mrry here. Any chance we support externally shuffled dataset in near future? With recent tf.data [snapshot](https://github.com/tensorflow/community/pull/193 and [service](https://github.com/tensorflow/community/pull/195) RFCs it might be easier to design a unified tf.data shuffle service for dataset that is too large to fit inside memory.

byronyi: Update: with a little bit tweak I managed to get the performance from 4.5k image/s to 7.2k image/s using 128 threads. Writes are scattered into multiple shards so it wouldn't be too hard.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,396 @@
...
+
+1.  `reader_fn`: Optional. A user provided reader function to use when reading
+    the snapshot back. This allows the user to specify the concurrency and
+    randomization required when reading from the snapshot.
+
+    `reader_fn` should be a function that accepts two arguments: (1) a list of
+    snapshot file paths, and (2) a reference to a `SnapshotDataset` class.
+    The function should return a `Dataset` class.
+
+    The `SnapshotDataset` class is a `Dataset` (similar to other source datasets
```
rohan100jain: Should this be named SnapshotReaderDataset instead?

rfcs/20200107-tf-data-snapshot.md
rohan100jain: @karmel While I think its a worthy objective of unifying the Bundle Reader / Writer file format with what we're proposing, I don't want to tie them both together this early. The reason is that since the workloads are quite different, we are still early days into exactly knowing what sort of options and functionality we'll have to build in to get the maximum throughput (e.g. we've needed compression so that we don't saturate network bandwidth too quickly, different threading implementations etc.). So my proposal would be to keep it separate at the moment, gain some experience tuning this workload and then we can pull in some of the learnings into Bundle Reader / Writer.

byronyi: @rohan100jain I completely agree with you. At the end of day we will be all better off with a format that yields highest performance possible, and we will definitely learned how to build one when experimenting a new format built from scratch.

Being said that, I would also like to keep improving tensor bundle, as the current on-disk cache uses it as underlying storage. @karmel would you be interested in sponsoring an RFC regarding to that? It could serve as a primary input source (TFIndexedRecord), accelerating on-disk CacheDataset, and enhance ShuffleDataset with external shufle. We could focus on the API first. If Rohan and @frankchn designed a more performant format it could be used for these APIs as well.

rfcs/20200107-tf-data-snapshot.md
frankchn: Good point. Done.

rfcs/20200107-tf-data-snapshot.md
karmel: Thanks, @byronyi -- @rohan100jain / @frankchn , what was the conclusion from the design review as to input format + readers?

frankchn: @karmel -- We are going to re-investigate TFRecordReader performance and see if we can bring that up to equivalency with our implementation.

rfcs/20200107-tf-data-snapshot.md
```diff
@@ -0,0 +1,395 @@
...
+
+In cases like these, we can ask fingerprinting to use a fixed value for the
+fingerprint of the map function with a new `set_fingerprint`
+transformation, which asks the fingerprinting function to not compute the 
+fingerprint of the previous node but to use a user-specified value instead:
+
+```python
+dataset = ...
+dataset = dataset.map(preprocessing_fn) 
+dataset = tf.data.set_fingerprint(dataset, fingerprint="my_fixed_fp")
```
jsimsa: I would rename the transformation to `set_snapshot_fingerprint` and make it experimental.

frankchn: Changing.

frankchn:
Hi everyone,

Thank you for commenting on this design review. We've recently completed an internal round of design review meetings for this feature, and we have taken into account many of the comments and have revised the design document.

Specifically, the largest changes we have made are (1) consolidating the various options into a user-specifiable reader_fn and writer_fn, where users can customize how dataset is read and written, and (2) removing `snapshot_name` and `snapshot_mode`, and instead implementing `with_snapshot_fingerprint` for users to override fingerprinting on parts of their input pipeline graph.

We appreciate all your comments and are looking forward to implementing this feature soon!

Thanks,

Frank

byronyi:
Congratulations! Hope this to make it into 2.2.

markemus:
I wrote a simple automatic serialization module a while back that generates a "header" file from a dataset of tensors. The header file stores the datatypes and shapes of each tensor and there's a function to restore the tensors using the TFrecord and header files. Would that be a useful addition to Snapshot? I'm unclear about whether the current implementation supports this header feature already or not.

https://gist.github.com/markemus/74ba47d0b58f91d7aa7885341ed3b1b8

Obviously it would be ideal if the header could be added directly into the tfrecords file but it's currently stored separately for compatibility reasons.

frankchn:
Thanks @markemus! The implementation of Snapshot has something similar but in C++ for higher performance, but `save` and `load` is something we are actively considering for `tf.data` and we are thinking through ways of saving and loading the dtypes safely and efficiently too.

markemus:
@frankchn cool! I'm really looking forward to snapshot, this is badly needed and it's great that you guys are working on this. If any of super_serial is useful it's all yours.

byronyi:
@frankchn @aaudiber @rohan100jain I have a question regarding to use snapshot in conjunction with tf.data service. Would it be feasible to shard the snapshot into multiple tf.data service workers?

aaudiber:
@byronyi Absolutely! We plan to integrate snapshotting and tf.data service so that the tf.data service speeds up the first (normally slow) epoch. The idea is to temporarily use the tf.data service to write the snapshot quickly, then shut down the service (saving resource costs), and read from the snapshot for the remaining epochs.

PR #195, opened 2020-01-13
RFC: tf.data Service

aaudiber:
*Comment period is open till 1/27/2020.*

| Status        | Accepted                          |
| **RFC #**     | [195](https://github.com/tensorflow/community/pull/195) |
| **Author(s)** | Andrew Audibert (aaudibert@google.com) Rohan Jain (rohanj@google.com)       |
| **Sponsor**   | Jiri Simsa (jsimsa@google.com)                          |
| **Updated**   | 2019-01-30

Provide an API and implementation of a tf.data service which can process tf.data
datasets in a distributed manner. The service can be run outside the TensorFlow
cluster or be exported as a gRPC service by TensorFlow servers.

Goals:

-   Enable horizontal scaling of dataset computation to improve performance of
    input-bound dataset pipelines.
-   Improve tf.data integration with the tf.distribute API. In particular,
    support dynamic sharding of data across multiple processes.
-   Provide visitation guarantees for distributed training jobs.

### DIFF
--- a/rfcs/20200113-tf-data-service.md
+++ b/rfcs/20200113-tf-data-service.md
@@ -1,0 +1,591 @@
+# Distributed tf.data service
+
+| Status        | Proposed          |
+| :------------ | :------------------------------------------------------ |
+| **RFC #**     | [195](https://github.com/tensorflow/community/pull/195) |
+| **Author(s)** | Andrew Audibert (aaudibert@google.com) Rohan Jain (rohanj@google.com) |
+| **Sponsor**   | Jiri Simsa (jsimsa@google.com)                          |
+| **Updated**   | 2019-01-09                                              |
+
+## Objective
+
+Provide an API and implementation of a tf.data service which can process tf.data
+datasets in a distributed manner. The service can be run outside the TensorFlow
+cluster or be exported as a gRPC service by TensorFlow servers.
+
+Goals:
+
+-   Enable horizontal scaling of dataset computation to improve performance of
+    input-bound dataset pipelines.
+-   Improve tf.data integration with the tf.distribute API. In particular,
+    support dynamic sharding of data across multiple processes.
+-   Provide visitation guarantees for distributed training jobs.
+
+Non-goals:
+
+-   Process non-dataset data.
+-   Distribute datasets that rely on external / non-serializable state.
+-   Support non-graph computation (e.g. py_function).
+
+## Motivation
+
+### Host machine input pipelines can't always keep up with accelerators.
+
+Some input pipelines require significant resources to produce their data, e.g.
+due to image transformations. When the host machine isn't powerful enough to
+generate input data at the rate the attached accelerator(s) consume the data,
+the accelerator(s) will idle. This slows down training time, and also wastes
+valuable accelerator resources. The tf.data service solves this problem by using
+N input workers to feed M accelerators. The number of input workers can be
+scaled up or down as needed to keep up with the accelerators.
+
+### Distributed training requires a distribution-aware input pipeline.
+
+Today tf.data supports the tf.distribute API by providing mechanisms for
+sharding, cloning, and re-batching. The tf.distribute API uses these primitives
+to implement their own version of a distributed dataset. If distributed datasets
+become a core feature of tf.data, tf.data can provide a public API for
+tf.distribute (and users who wish to implement their own distribution) to use
+instead. This will also allow us to support feature requests that require
+cross-worker coordination, such as dynamic sharding.
+
+## User Benefit
+
+### Input-bound models
+
+Users with input-bound models can leverage the tf.data service to distribute
+input processing across horizontally-scaling compute resources. This can improve
+utilization for valuable accelerator resources, reducing total cost.
+
+### Dynamic load balancing
+
+Today, the tf.distribute API statically shards data across accelerators. This
+can lead to suboptimal utilization because some shards may contain more data
+than others. The tf.data service provides a mechanism for dynamically sharding,
+reducing the data imbalance across accelerators.
+
+### Visitation guarantees
+
+Model accuracy can often be improved when each training sample is trained on
+exactly once per epoch. The tf.data service can coordinate across workers to
+provide this guarantee.
+
+## Design Proposal
+
+The tf.data service is a master-worker system which iterates through datasets,
+producing outputs to be consumed by accelerators. The service is comprised of a
+few components:
+
+*   User-facing Python API for interacting with the tf.data service.
+*   Dataset splitting API for determining how to split up datasets for parallel
+    processing.
+*   Master and worker gRPC services.
+
+### Architecture
+
+The tf.data service is comprised of master and worker gRPC services which could
+be run in a couple of different configurations:
+
+#### Glossary
+
+**Master**: The single master coordinating the tf.data service.
+
+**Worker**: A tf.data service worker which performs dataset processing and
+provides dataset elements to consumers over RPC.
+
+**Consumer**: A machine which consumes data from the tf.data service. The
+consumer may be attached to a GPU or TPU, or use data for on-CPU training.
+
+#### Separate Cluster Architecture
+
+Each server is run on a separate host from the TensorFlow cluster. This
+configuration gives users a way to provide horizontally scaling CPU for
+processing their input pipelines and quickly feeding data to accelerators.
+
+#### Embedded Cluster Architecture
+
+without needing to provision additional compute resources. and gives all the
+benefits of the tf.data service except for horizontal scaling.
+
+#### Hybrid Architecture
+
+Users could run tf.data workers embedded in their TensorFlow cluster, and also
+run additional tf.data workers (and potentially the tf.data master) outside the
+cluster. This allows for horizontal worker scaling, while still leveraging the
+compute resources of the TensorFlow cluster for input processing.
+
+### User-facing Python API
+
+This API is how users will interact with the tf.data service from their Python
+code.
+
+```python
+def tf.data.experimental.service.distribute(address):
+  """Marks that a dataset should be processed by the tf.data service.
+
+  ds = ... # dataset to distribute
+  ds = ds.apply(tf.data.experimental.service.distribute(address))
+
+  Args:
+    address: The address of the tf.data service master.
+
+  Returns:
+    A function that can be passed to `dataset.apply()`.
+  """
+
+def tf.data.experimental.service.create_iteration(
+    dataset, num_consumers=1, num_tasks=None, deterministic=False):
+  """Begins distributed iteration over a dataset.
+
+  It is expected that the dataset contains at least one `.distribute(address)`
+  transformation, otherwise this method will print a warning and do nothing.
+
+  `create_iteration` will first register the dataset with the tf.data service
+  if it isn't already registered. It will then request the creation of
+  `num_consumers` dataset iterators which divide the dataset `num_consumers`
+  ways. The returned object can be used to read from one of the
+  iterators using
+  `tf.data.experimental.service.make_iterator(ds, obj, consumer_index)`.
+
+  ds = ... # dataset to distribute
+  ds = ds.apply(tf.data.experimental.service.distribute(address))
+  if consumer_index == 0:
+    # The iteration object is a byte array which needs to be shared among all
+    # consumers. Here we suppose there are broadcast_send and broadcast_recv
+    # method available.
+    iteration_id = tf.data.experimental.service.create_iteration(ds, address, 3)
+    broadcast_send(iteration_id)
+  else:
+    iteration_id = broadcast_recv()
+  it = tf.data.experimental.service.make_iterator(
+      ds, iteration_id, consumer_index)
+  for element in it:
+    # process element
+
+  Args:
+    dataset: The dataset to begin iteration over.
+    num_consumers: The number of consumers to divide the dataset between. Set
+      this if you require determinism. If None, a single iterator id is returned,
+      and any number of consumers can read from that iterator id. The data
+      produced by the dataset will be fed to consumers on a first-come
+      first-served basis.
+    num_tasks: The number of tasks to use for processing. Tasks run for
+      the duration of an epoch, and each worker should typically process a single
+      task. Normally it is best to leave this as None so that the master can
+      choose a reasonable number of tasks. Setting `num_tasks` is useful for
+      producing deterministic results.
+    deterministic: Whether the iteration should be performed
+      deterministically. Fully deterministic output also requires setting
+      `num_tasks` to a fixed number, and that the input dataset is itself
+      deterministic.
+
+  Returns:
+    An iteration_id which can be used to created iterators via
+      `tf.data.experimental.service.make_iterator`
+  """
+
+def tf.data.experimental.service.make_iterator(
+    dataset, iteration, consumer_index):
+  """Creates an iterator for reading from the specified dataset.
+
+  Args:
+    dataset: The dataset to read from.
+    iteration: An iteration_id object generated by
+      `tf.data.experimental.service.create_iteration`.
+    consumer_index: The consumer index within the iteration to read from. If
+      the iteration was created with `n` consumers, `consumers_index` must be
+      less than `n`.
+
+  Returns:
+    A Python iterator which iterates over the dataset elements.
+  """
+```
+
+### Dataset splitting API
+
+To parallelize dataset processing, the tf.data service needs a way to split up
+datasets. We will achieve this by adding a splitting API that allows source
+datasets to express how they can be split.
+
+Our goals for the API are
+
+*   Performance: The splitting API can be used to performantly split and process
+    datasets.
+*   Extensibility: User-defined datasets can be split as long as they implement
+    the splitting API.
+*   Minimize Surprises: Users write their datasets as though they will not be
+    split, so introducing splitting can easily lead to unexpected outcomes. To
+    mitigate this, we will be conservative about which dataset transformations
+    support splitting.
+
+The API will be used internally by the tf.data service to distribute datasets.
+It will be entirely in C++, and we don't currently have any plans to expose
+splitting through Python.
+
+The API focuses on producing and consuming `Split`s. A `Split` is a variant
+Tensor that can be subclassed to represent arbitrary types of splitting.
+
+```cpp
+class Split {
+ public:
+  virtual std::string DebugString() const = 0;
+  // Methods to support being used as a Variant tensor.
+  virtual std::string TypeName() const = 0;
+  virtual void Encode(VariantTensorData* data) const = 0;
+  virtual bool Decode(const VariantTensorData& data) = 0;
+};
+```
+
+To iterate over splits for a dataset, we will use a new
+`DatasetBase::MakeSplitGenerator()` method. This method creates a
+`SplitGenerator`, which is responsible for generating all of the splits for the
+dataset. We use an intermediate `SplitGenerator` object instead of generating
+splits directly because there could be a large number of splits, and the
+`SplitGenerator` gives us as way to tune split size in response to pipeline
+performance.
+
+```cpp
+class SplitGenerator {
+ public:
+  virtual Status GetNext(std::unique_ptr<Split>* split,
+                         bool* end_of_splits) = 0;
+  // Instructs the SplitGenerator to adjust the size of future splits by the
+  // specified percent. 100% means no change, 50% means half-sized splits, and
+  // 200% means double-sized splits. The SplitGenerator will make a best effort
+  // to incorporate the feedback when creating splits.
+  virtual void AdjustSplitSize(int percent) = 0;
+};
+```
+
+It is tempting to process each split independently, but this would cause issues
+when splits are small. tf.data pipelines need to populate internal buffers for
+shuffling, prefetching, and batching. If we use a separate pipeline to process
+each split, our shuffling will be lower quality, we will have performance jitter
+as we keep needing to refill prefetch buffers from scratching, and we will
+produce many more partial batches (each split might not even have enough data to
+fill a full batch). To avoid these issues, we use a small number of tasks, where
+each task processes many splits as a single pipeline.
+
+To enable processing of multiple splits in a dataset, we will add an optional
+`SplitProvider` field to the `IteratorContext` passed to
+`IteratorBase::Initialize`. The `SplitProvider` produces splits which tell the
+iterator what source data to iterate over. For example, if splits are
+represented by filenames, and a SplitProvider produces `["file1", "file6",
+"file11"]`, an iterator initialized by that `SplitProvider` should process those
+three files only.
+
+```cpp
+class SplitProvider {
+ public:
+  virtual Status GetNext(std::unique_ptr<Split>* split,
+                         bool* end_of_splits) = 0;
+};
+```
+
+When processing datasets, tf.data service workers will use `SplitProvider`s
+which provide splits by querying the tf.data service master for which splits to
+process. A few splits will be prefetched to hide the latency of needing to
+request a new split from the master.
+
+#### Supported Datasets
+
+Not all dataset sources and transformations are easily splittable. For example,
+`take`, `skip`, and `scan` require a global view of the dataset to produce
+correct results. Datasets which require multiple input datasets such as `zip`
+are also difficult to support, since we don't have a good way of aligning the
+splits of multiple input datasets. Users who rely on these unsupported datasets
+will need to move those datasets to come after the distributed part of their
+pipeline.
+
+Initially, we will support splitting for the following dataset sources and
+transformations:
+
+*   `batch`, `CsvDataset`, `dense_to_sparse_batch`, `filter`,
+    `FixedLengthRecordDataset`, `flat_map`, `from_tensor_slices`,
+    `group_by_window`, `ignore_errors`, `interleave`, `list_files`, `map`,
+    `range`, `repeat`, `padded_batch`, `prefetch`, `shuffle`, `SSTableDataset`,
+    `TextLineDataset`, `TFRecordDataset`, `unbatch`, `window`.
+
+### Master and worker services
+
+This section discusses the design for the master and worker services. These
+services are used by the Python API to provide distributed dataset processing,
+and these services use the splitting API as a part of their implementation.
+
+#### Master API
+
+The master is responsible for registering datasets, generating and tracking
+iteration and worker ids, and generating dataset splits for processing on
+workers.
+
+Below is a sketch of the Master API. This API is not public and is subject to
+change.
+
+```cpp
+/// ---- Methods called by consumers ----
+
+// Registers a dataset and returns an id for the dataset. If the dataset is
+// already registered, its dataset id is returned.
+int GetOrRegisterDataset(GraphDef dataset);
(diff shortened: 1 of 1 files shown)


rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+N input workers to feed M accelerators. The number of input workers can be
+scaled up or down as needed to keep up with the accelerators.
+
+### Distributed training requires a distribution-aware input pipeline.
+
+Today tf.data supports the tf.distribute API by providing mechanisms for
+sharding, cloning, and re-batching. The tf.distribute API uses these primitives
+to implement their own version of a distributed dataset. If distributed datasets
+become a core feature of tf.data, tf.data can provide a public API for
+tf.distribute (and users who wish to implement their own distribution) to use
```
alextp: This is really important, and I'd like to review the distributed data APIs independently from the service model.

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+
+## User Benefit
+
+### Input-bound models
+
+Users with input-bound models can leverage the tf.data service to distribute
+input processing across horizontally-scaling compute resources. This can improve
+utilization for valuable accelerator resources, reducing total cost.
+
+### Dynamic load balancing
```
alextp: How do we preserve determinism when we're doing dynamic load balancing?

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+
+#### Separate Cluster Architecture
+
+Each server is run on a separate host from the TensorFlow cluster. This
+configuration gives users a way to provide horizontally scaling CPU for
+processing their input pipelines and quickly feeding data to accelerators.
+
+#### Embedded Cluster Architecture
+
+without needing to provision additional compute resources. and gives all the
```
alextp: Weird sentence structure; was this an editing mistake?

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+benefits of the tf.data service except for horizontal scaling.
+
+#### Hybrid Architecture
```
alextp: I don't understand the conjunction of this and the "separate cluster architecture" above

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+  """
+
+def tf.data.experimental.service.create_iteration(
+    dataset, num_consumers=1, num_tasks=None, deterministic=False):
+  """Begins distributed iteration over a dataset.
+
+  It is expected that the dataset contains at least one `.distribute(address)`
+  transformation, otherwise this method will print a warning and do nothing.
+
+  `create_iteration` will first register the dataset with the tf.data service
```
alextp: How does create_iteration find the service?

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+  if it isn't already registered. It will then request the creation of
+  `num_consumers` dataset iterators which divide the dataset `num_consumers`
+  ways. The returned object can be used to read from one of the
+  iterators using
+  `tf.data.experimental.service.make_iterator(ds, obj, consumer_index)`.
```
alextp: Why not the usual python control flow we use in regular tf.data?

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+      `tf.data.experimental.service.create_iteration`.
+    consumer_index: The consumer index within the iteration to read from. If
+      the iteration was created with `n` consumers, `consumers_index` must be
+      less than `n`.
+
+  Returns:
+    A Python iterator which iterates over the dataset elements.
+  """
+```
+
```
alextp: An earlier discussion about this involved a public primitive to create a dataset from an iterator, which would also be useful internally when implementing create_iteration. Is that still in scope?

I ask because while mostly unrelated to the serve issue this would be very useful when dealing with the types of pipelines which are non-goals of the service project (stateful pipelines, pipelines with py_func, etc)

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+The API will be used internally by the tf.data service to distribute datasets.
+It will be entirely in C++, and we don't currently have any plans to expose
+splitting through Python.
+
+The API focuses on producing and consuming `Split`s. A `Split` is a variant
+Tensor that can be subclassed to represent arbitrary types of splitting.
+
+```cpp
+class Split {
```
alextp: This is very vague. What is stored in the variant?

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+
+When processing datasets, tf.data service workers will use `SplitProvider`s
+which provide splits by querying the tf.data service master for which splits to
+process. A few splits will be prefetched to hide the latency of needing to
+request a new split from the master.
+
+#### Supported Datasets
+
+Not all dataset sources and transformations are easily splittable. For example,
+`take`, `skip`, and `scan` require a global view of the dataset to produce
```
alextp: How do take and skip require global views?

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+are also difficult to support, since we don't have a good way of aligning the
+splits of multiple input datasets. Users who rely on these unsupported datasets
+will need to move those datasets to come after the distributed part of their
+pipeline.
+
+Initially, we will support splitting for the following dataset sources and
+transformations:
+
+*   `batch`, `CsvDataset`, `dense_to_sparse_batch`, `filter`,
+    `FixedLengthRecordDataset`, `flat_map`, `from_tensor_slices`,
```
alextp: from_tensor_slices seems like a weird fit for the service model, as that would require bouncing the tensors to remote workers and back AFAICT

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+#### Determinism
+
+Deterministic processing is a cornerstone of tf.data. Determinism is valuable
+for debugging and experimentation. This section discusses how the tf.data
+service will provide determinism.
+
+To get deterministic behavior, the tf.data service will require three things:
+
+1.  The dataset being distributed has deterministic output.
+1.  The user sets `deterministic=True` when calling
```
alextp: Why requiring determinism to be opt-in instead of making non-determinism opt-in like the rest of the tf.data API?

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+
+// Requests that the worker process the specified dataset. This will trigger the
+// worker to start requesting splits from the master using the `iteration_id`.
+void ProcessDataset(int dataset_id, int iteration_id, list<int> iterator_ids);
+```
+
+#### Visitation Guarantees
+
+When iterating over a dataset, the tf.data service will process all input data
+at least once, even in the presence of master or worker failures. If there are
```
alextp: Can you expand a bit on the tradeoffs of not enforcing "exactly once" when failures are present?

rfcs/20200113-tf-data-service.md
alextp: I ask because I can think of some not-too-expensive ways of enforcing exactly-once, like if consumers request a <split_id, batch_within_split> instead of just "next batch please" in the RPC, and things like that.

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+    the master_seqno_n file to master_seqno_n+1. If this succeeds, the master
+    will wait (N + M) milliseconds, verify that its renamed file still exists,
+    and begin acting as leader. This gives the previous leader time to notice
+    the rename.
+
+The above scheme relies on rename being atomic so that two masters don't both
+succeed at renaming the same file. Users may opt to use a filesystem that
+doesn't support atomic rename, but they do so at the (unlikely) risk of two
+concurrently running masters thinking they are leader. Common filesystems such
+as Posix and HDFS support atomic rename.
```
alextp: And does GCS support it too?

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+
+This section calls out caveats that users will need to be aware of when using
+the tf.data service.
+
+-   Due to the nature of dataset splitting, elements will not be processed in
+    the same order as they were in the pre-distributed dataset. If a dataset
+    relies on the order of the input files, the user's assumptions will be
+    violated when splitting causes each input worker to process only a subset of
+    the input files.
+-   If a dataset doesn't support splitting, it must be moved after the part of
```
alextp: I assume "dataset doesn't support splitting"  means "if a particular dataset operation doesn't support splitting" here

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,591 @@
...
+doesn't support atomic rename, but they do so at the (unlikely) risk of two
+concurrently running masters thinking they are leader. Common filesystems such
+as Posix and HDFS support atomic rename.
+
+#### Caveats
+
+This section calls out caveats that users will need to be aware of when using
+the tf.data service.
+
+-   Due to the nature of dataset splitting, elements will not be processed in
```
alextp: I think it's better to document an actual guarantee, such that the order within each split must be consistent with the original global order of the source dataset, but no promises are made around ordering across splits.

byronyi:
Thanks @aaudiber and @rohan100jain for your RFC! This is just in time when we find ourselves in need to support a better distributed data pipeline with fault tolerance and elastic scaling after benefiting from high performance distribution strategy along with single-node tf.data pipeline.

When designing the RPC service, would you mind to expose the tensor buffer information (like [here](https://github.com/tensorflow/tensorflow/blob/90f01af49ac15934aed92b8d459175336db1573b/tensorflow/core/protobuf/worker.proto#L468-L483) so high performance networking such as RDMA could use the memory address of sender and receiver to perform direct memory transfer without extra copy (from/to kernel socket buffer)? Existing networking plugins in https://github.com/tensorflow/networking could be extended to support tf.data service for improved performance.

rfcs/20200113-tf-data-service.md
aaudiber: Determinism isn't possible if the load balancing uses time-based heuristics to balance load. If deterministic mode is configured, there will be no dynamic load balancing.

rfcs/20200113-tf-data-service.md
aaudiber: Yeah, the previous two lines were deleted by accident. Fixed now, thanks

rfcs/20200113-tf-data-service.md
aaudiber: Separate, Embedded, and Hybrid are three possible options for how the service is deployed.

rfcs/20200113-tf-data-service.md
aaudiber: The service address is configured within the dataset by calling `dataset.apply(tf.data.experimental.service.distribute(address))`

rfcs/20200113-tf-data-service.md
aaudiber: In regular tf.data there is only one way to iterate over a dataset, so it's natural to make `Dataset` an `Iterable`. But with tf.data service we iterate using `N` (non-independent) consumer streams which dynamically partition a single iteration's worth of data. We need a way for the consumers to identify to the tf.data service that they all want to read from the same iteration.

rfcs/20200113-tf-data-service.md
aaudiber: Such an iterator-to-dataset operation would be useful for the implementation of `make_iterator`, but for the purpose of this design we would like to avoid publicly exposing any more API than we absolutely need to.

rfcs/20200113-tf-data-service.md
aaudiber: The interface is intentionally vague so that implementors can define their splits however makes the most sense for their dataset type.

rfcs/20200113-tf-data-service.md
aaudiber: take and skip use the global order of elements to determine which elements to produce. With splitting, the order is blurred so that it isn't clear how to compare the order of elements produced from different splits.

rfcs/20200113-tf-data-service.md
aaudiber: It would indeed require bouncing tensors around, so it isn't a great fit when the tensors are large. I could imagine from_tensor_slices being useful though if the user passes a list of filenames to `from_tensor_slices`, then does expensive per-file processing.

rfcs/20200113-tf-data-service.md
aaudiber: We want to give good out of the box performance for users using the tf.data service to speed up input processing. Achieving deterministic behavior is more expensive in a distributed environment than in a single-host environment because we need to use a fixed number of tasks, we can't dynamically balance load, and we can't sloppily interleave results. We expect the common case to be a desire for performance over determinism.

Let me think a little bit more about how to achieve exactly-once visitation with non-deterministic service execution. I'll get back to you

rfcs/20200113-tf-data-service.md
aaudiber: GCS doesn't support atomic rename. I think we could leverage GCS's strongly consistent directory listing to achieve something similar to atomic file rename, and loosen the requirement to "The filesystem supports either atomic rename or strongly consistent listing".

rfcs/20200113-tf-data-service.md
aaudiber: Correct. Updated for clarity, thanks.

rfcs/20200113-tf-data-service.md
aaudiber: We don't have such a guarantee because the dataset pipeline for each task processes multiple splits, so even the order within the split could be different than the order if the split was processed in isolation.

rfcs/20200113-tf-data-service.md
aaudiber: Does the "User-facing Python API" section address this?

rfcs/20200113-tf-data-service.md
aaudiber: I've updated the proposal to support an exactly-once guarantee even when tf.data service processing is non-deterministic.

rfcs/20200113-tf-data-service.md
aaudiber: Revised this section to support exactly once even when failures are present.

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,593 @@
...
+change.
+
+```cpp
+/// ---- Methods called by consumers ----
+
+// Registers a dataset and returns an id for the dataset. If the dataset is
+// already registered, its dataset id is returned.
+int GetOrRegisterDataset(GraphDef dataset);
+
+// Creates and returns `num_consumers` iterator ids which partition the
```
feihugis: IIUC, the Split-related classes decide how to split/partition the dataset. How does `num_consumers` work here?

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,593 @@
...
+// specified dataset. This also creates an internal `iteration_id` used to
+// track the overall dataset iteration. `num_tasks` defines how many tasks to
+// create. If `num_tasks` is -1, it is up to the master to determine how many
+// tasks to create.
+list<int> CreateIterators(int dataset_id, int num_consumers,
+                          int num_tasks);
```
feihugis: Will it be better to define `num_tasks` as the number of tasks per worker (defaults to 1)? This may be easier to set. Otherwise, users need to count how many workers in the cluster.

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,593 @@
...
+
+// Returns the list of tasks processing data for `iterator_id`. Consumers query
+// this to find which worker addresses to read data from.
+list<TaskInfo> GetWorkersForiterator(int iterator_id);
```
feihugis: Could you please give more details about how to decide which workers the consumer should read data from? Could we enable the end-users to customize the strategy?

nit: `GetWorkersForiterator` -> `GetWorkersForIterator`.

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,593 @@
...
+
+Some input pipelines require significant resources to produce their data, e.g.
+due to image transformations. When the host machine isn't powerful enough to
+generate input data at the rate the attached accelerator(s) consume the data,
+the accelerator(s) will idle. This slows down training time, and also wastes
+valuable accelerator resources. The tf.data service solves this problem by using
+N input workers to feed M accelerators. The number of input workers can be
+scaled up or down as needed to keep up with the accelerators.
+
+### Distributed training requires a distribution-aware input pipeline.
```
feihugis: The distributed dataset will be very useful for the inference as well!

rfcs/20200113-tf-data-service.md
aaudiber: `num_consumers` is needed for deterministic execution, so that workers can earmark every nth element to go to a specific consumer. The alternative is to give out elements on a first-come, first-served basis, which is more performant but nondeterministic.

rfcs/20200113-tf-data-service.md
aaudiber: Deterministic execution makes this challenging. To achieve determinism across executions, the number of tasks needs to remain the same. Since the number of workers could change, `num_tasks` can not depend on the number of workers.

rfcs/20200113-tf-data-service.md
aaudiber: Initially, the master will try to schedule one task per worker. Ideally we can make the scheduling smart enough that users don't need to worry about it, but if we can't achieve that, we can consider letting the user specify a scheduling strategy.

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,593 @@
...
+  `create_iteration` will first register the dataset with the tf.data service
+  if it isn't already registered. It will then request the creation of
+  `num_consumers` dataset iterators which divide the dataset `num_consumers`
+  ways. The returned object can be used to read from one of the
+  iterators using
+  `tf.data.experimental.service.make_iterator(ds, obj, consumer_index)`.
+
+  ds = ... # dataset to distribute
+  ds = ds.apply(tf.data.experimental.service.distribute(address))
+  if consumer_index == 0:
```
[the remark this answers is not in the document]
feihugis: Thanks for the link!

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,593 @@
...
+    # The iteration object is a byte array which needs to be shared among all
+    # consumers. Here we suppose there are broadcast_send and broadcast_recv
+    # method available.
+    iteration_id = tf.data.experimental.service.create_iteration(ds, address, 3)
+    broadcast_send(iteration_id)
+  else:
+    iteration_id = broadcast_recv()
+  it = tf.data.experimental.service.make_iterator(
+      ds, iteration_id, consumer_index)
```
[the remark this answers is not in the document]
feihugis: Gotcha!

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,593 @@
...
+
+def tf.data.experimental.service.make_iterator(
+    dataset, iteration, consumer_index):
+  """Creates an iterator for reading from the specified dataset.
+
+  Args:
+    dataset: The dataset to read from.
+    iteration: An iteration_id object generated by
+      `tf.data.experimental.service.create_iteration`.
+    consumer_index: The consumer index within the iteration to read from. If
```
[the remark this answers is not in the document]
feihugis: Yeah, the section added is helpful.

pavanky:
This looks great! I know this is not the goal of this proposal, but would it be possible (in the future) to build on this and have the training jobs consume the datasets from any a gRPC service (not necessarily one generated using tensorflow) as long as it conforms to a certain API ?

byronyi:
> This looks great! I know this is not the goal of this proposal, but would it be possible (in the future) to build on this and have the training jobs consume the datasets from any a gRPC service (not necessarily one generated using tensorflow) as long as it conforms to a certain API ?

@pavanky Checkout https://github.com/tensorflow/io/pull/206

pavanky:
> > This looks great! I know this is not the goal of this proposal, but would it be possible (in the future) to build on this and have the training jobs consume the datasets from any a gRPC service (not necessarily one generated using tensorflow) as long as it conforms to a certain API ?
> 
> @pavanky Checkout [tensorflow/io#206](https://github.com/tensorflow/io/pull/206)

Thanks!

rfcs/20200113-tf-data-service.md
aaudiber: We should consider allowing users to specify a ClusterResolver here

rfcs/20200113-tf-data-service.md
aaudiber: Updated the doc to call out that users must choose between dynamic load balancing and deterministic execution.

rfcs/20200113-tf-data-service.md
aaudiber: Updated to make it clear that the sections describe different alternatives.

rfcs/20200113-tf-data-service.md
aaudiber: Added content explaining why we don't support usual python control flow

rfcs/20200113-tf-data-service.md
aaudiber: Previously we were thinking that `distribute` would return an opaque object which can be used to create an iterator (or multiple shared iterators). Then to apply local transformations, we would need a way to go from iterator to dataset.

We decided to take an alternative path: the `distribute` transformation will return a regular `Dataset`, which the user can chain transformations onto. We no longer need to go from iterator to dataset.

rfcs/20200113-tf-data-service.md
aaudiber: Added this comment to the doc

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,608 @@
...
+
+#### Glossary
+
+**Master**: The single master coordinating the tf.data service.
+
+**Worker**: A tf.data service worker which performs dataset processing and
+provides dataset elements to consumers over RPC.
+
+**Consumer**: A machine which consumes data from the tf.data service. The
+consumer may be attached to a GPU or TPU, or use data for on-CPU training.
```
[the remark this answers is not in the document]
aaudiber: What sort of configuration are you interested in?

aaudiber:
Meeting notes from design review on 1/30/20 (thank you @rohan100jain for taking these!):

**Changes to do the doc since mailed out**
* Visitation guarantees: Even if pipeline is non-deterministic, we can provide exactly once visitation guarantee.
* Other changes are part of the comments to be discussed now.

**Distribution may change the order of elements being produced by the dataset. How do we communicate this? Current plan is documentation**
* apassos: Hopefully the elements dont change.
* andrew: the batches created could be different because we combine different examples.

**skip, take and scan might not be splittable e.g. take: do it per task and so we end up with 10 * num_worker elements**
 * andrew: one proposal is that we do it on a per task level since we already do it for distribution strategies.
* derek: can we make it clear in the API that some prefix is being distributed and then afterward  its not distributed. 
* andrew: can chain transformation after distribute for that distinction
* derek: should we prohibit this transformations in the distributed part.
* apassos: didn’t realize we can chain after distribute as well

*decision: prohibit and ask users to chain it afterwards*

**Share iteration ids between tasks**
* apassos: this is only needed for multi-client training, single graph we can just give it. Are these regular tensors?
* derek: collective broadcast? 
* apassos: we’re working on creating these collective ops and they will be available open source soon.
* anjali: how often do we need to do this? A: once per epoch

*decision: use collective ops to broadcast the iteration ids. They will be publicly available soon.*

**service.distribute takes in a ClusterResolver instead of a master address**
* derek: ClusterResolver is a python object?
* andrew: yes, and its not serializable right now
* derek: we can use a pyfunc to see the ClusterResolver.
* apassos: We might need a C++ cluster resolver thing for Kubernetes etc. 

*decision: allow ClusterResolver as input and change API arg to master_address_or_resolver*

**azaks: is the order deterministic?**
* apassos: can we guarantee a partial order that is consistent with the global order. I think we have this true for the current distributed dataset.
* jiri: if we shard the filenames, then we can have batches with different elements.
* apassos: if you had no shuffle and no batching, then if B comes after A, we can guarantee? Might be needed for something like AdBrain where strong ordering is needed.
* jiri: We might have to add more transformations to the fine print. Within a split, we can guarantee ordering. We can run the pipeline deterministically but it’ll come at a performance disadvantage.
* derek: we might be conflicting non-determinism with ordering (flat map vs. interleave). We can get most of the performance by having different ordering but deterministic execution.
* andrew: we can have some advanced documentation that describes what are the guarantees.

byronyi:
Congratulations to acceptance of this RFC! Can't wait to see it implemented in TF core.

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,648 @@
...
+    generate dataset elements. The most commonly used but unsupported datasets
+    are `from_generator` and `zip`.
+
+#### Framework Integration
+
+Many users interact with TensorFlow through a framework such as
+[TFX](https://www.tensorflow.org/tfx). A framework can make leveraging the
+tf.data service as simple as toggling a configuration boolean, triggering the
+framework to bring up tf.data service servers and add a
+`tf.data.experimental.service.distribute` transformation at the end of the
```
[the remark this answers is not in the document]
aaudiber: Tooling could be built around the tf.data service so that it can automatically scale, but I wouldn't consider that part of the core tf.data service. To enable autoscaling, the tf.data service would report whether it could use more resources, and it is up to external tooling to start more tf.data servers.

rfcs/20200113-tf-data-service.md
```diff
@@ -0,0 +1,648 @@
...
+    iteration_id = broadcast_recv()
+  it = tf.data.experimental.service.make_iterator(
+      ds, iteration_id, consumer_index)
+  for element in it:
+    # process element
+
+  Args:
+    dataset: The dataset to begin iteration over.
+    num_consumers: The number of consumers to divide the dataset between. Set
+      this if you require determinism.
```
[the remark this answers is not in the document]
katsiapis: Should the default then be None (instead of 1)?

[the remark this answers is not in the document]
aaudiber: Sorry, I thought it was `None` already but you are right! It should be `None`.

thisisandreeeee:
Would this also help address memory issues associated with caching large datasets? In a distributed tf.data architecture, is the dataset cache distributed without overlap across the workers?

rfcs/20200113-tf-data-service.md
byronyi: Agreed. A ClusterResolver would be helpful.

byronyi:
@thisisandreeeee take a look at #193.

PR #201, opened 2020-02-05
RFC: TFX Generic Trainer / Native Keras Support

1025KB:
**This RFC will be open for comment until Wednesday, Feb 19th, 2020.**
cc @karmel

# TFX Generic Trainer

| Status        | Proposed                                               |
| :------------ | :----------------------------------------------------- |
| **RFC #**      |  201 |
| **Author(s)** | Jiayi Zhao (jyzhao@google.com) |
| **Sponsor**   | Konstantinos Katsiapis (katsiapis@google.com), Zhitao Li (zhitaoli@google.com), Karmel Allison (karmel@google.com) |
| **Updated**   | 2020-01-17                                             |

## Objective

Support any TensorFlow Training loop in TFX Trainer in addition to tf.estimator, primarily focused on native Keras model.

This doc will focus on native Keras support (without model_to_estimator) in TFX.
We propose changing the user facing API to be more generic so that users can do
(single node) native Keras model training within TFX.

### DIFF
--- a/rfcs/20200117-tfx-generic-trainer.md
+++ b/rfcs/20200117-tfx-generic-trainer.md
@@ -145,6 +145,7 @@ def run_fn(args: trainer.executor.TrainerFnArgs) -> None:
   """Build the TF model and train it."""
   model = _build_keras_model()
   model.fit(...)
+  # Save model to args.serving_model_dir.
   model.save(...)
 ```
 
--- a/rfcs/20200117-tfx-generic-trainer.md
+++ b/rfcs/20200117-tfx-generic-trainer.md
@@ -237,10 +237,8 @@ def run_fn(args: trainer.executor.TrainerFnArgs) -> None:
   """Build the TF model and train it."""
   ps_strategy = tf.distribute.experimental.ParameterServerStrategy()
   with ps_strategy.scope():
-    train_dataset = _create_dataset(args.train_files, ...)
-    eval_dataset = _create_dataset(args.eval_files, ...)
     model = _build_keras_model()
-  model.fit(train_dataset, validation_data=eval_dataset, ...)
+  model.fit(...)
   if _is_chief():
     model.save(...)
 ```
--- a/rfcs/20200117-tfx-generic-trainer.md
+++ b/rfcs/20200117-tfx-generic-trainer.md
@@ -237,8 +237,10 @@ def run_fn(args: trainer.executor.TrainerFnArgs) -> None:
   """Build the TF model and train it."""
   ps_strategy = tf.distribute.experimental.ParameterServerStrategy()
   with ps_strategy.scope():
+    train_dataset = _create_dataset(args.train_files, ...)
+    eval_dataset = _create_dataset(args.eval_files, ...)
     model = _build_keras_model()
-  model.fit(...)
+  model.fit(train_dataset, validation_data=eval_dataset, ...)
   if _is_chief():
     model.save(...)
 ```
--- a/rfcs/20200117-tfx-generic-trainer.md
+++ b/rfcs/20200117-tfx-generic-trainer.md
@@ -3,7 +3,7 @@
 | Status        | Proposed                                                  |
 | :------------ | :-------------------------------------------------------- |
 | **Author(s)** | Jiayi Zhao (jyzhao@google.com)                            |
-| **Sponsor**   | Konstantinos Katsiapis (katsiapis@google.com), Zhitao Li (zhitaoli@google.com), Karmel Allison (karmel@google.com)                                                         |
+| **Sponsor**   | Konstantinos Katsiapis (katsiapis@google.com), Zhitao Li (zhitaoli@google.com), Karmel Allison (karmel@google.com) |
 | **Updated**   | 2020-01-17                                                |
 
 ## Objective
@@ -197,7 +197,7 @@ def run_fn(fn_args: executor.TrainerFnArgs):
 
 ### tf.distribute.Strategy
 
-Distribute strategy will be user module's responsibilty with the new generic
+Distribution strategy will be user module's responsibilty with the new generic
 trainer interface. To use it, user needs to modify the `run_fn()` in the module
 file, below shows the pseudo code example for single worker and multi-worker
 distribute strategy.
--- a/rfcs/20200117-tfx-generic-trainer.md
+++ b/rfcs/20200117-tfx-generic-trainer.md
@@ -1,0 +1,253 @@
+# TFX Generic Trainer
+
+| Status        | Proposed                                                  |
+| :------------ | :-------------------------------------------------------- |
+| **Author(s)** | Jiayi Zhao (jyzhao@google.com)                            |
+| **Sponsor**   | Konstantinos Katsiapis (katsiapis@google.com), Zhitao Li  |
+:               : (zhitaoli@google.com), Karmel Allison (karmel@google.com) :
+| **Updated**   | 2020-01-17                                                |
+
+## Objective
+
+### Goal
+
+*   Support any TensorFlow Training loop in TFX Trainer in addition to
+    tf.estimator, primarily focused on native Keras model.
+
+### Non Goal
+
+*   Natively support multi-worker distributed training by the system.
+*   Non-TF training that generates savedmodel.
+
+## Background and Motivation
+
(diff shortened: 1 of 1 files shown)


rfcs/20200117-tfx-generic-trainer.md
```diff
@@ -0,0 +1,252 @@
...
+hence not covered in the scope of this RFC. If the execution environment of an
+implementation of TFX Trainer has the ability to bring up the cluster of worker
+machines, and execute user funtion in the workers with correct
+[TF_CONFIG setup](https://www.tensorflow.org/guide/distributed_training#setting_up_tf_config_environment_variable),
+such as GCP AI Platform Training service via
+[extensions/google_cloud_ai_platform/trainer/executor.py](https://github.com/tensorflow/tfx/blob/r0.21/tfx/extensions/google_cloud_ai_platform/trainer/executor.py),
+the `run_fn()` would look like below:
+
+```python
+def _is_chief() -> bool:
```
omalleyt12: We have the private `strategy.extended._is_chief()`, can we work with TF Distribute team to make a canonical way of accessing this info?

rfcs/20200117-tfx-generic-trainer.md
```diff
@@ -0,0 +1,252 @@
...
+      estimator=training_spec['estimator'],
+      export_dir_base=fn_args.eval_model_dir,
+      eval_input_receiver_fn=training_spec['eval_input_receiver_fn'])
+
+  absl.logging.info('Exported eval_savedmodel to %s.', fn_args.eval_model_dir)
+```
+
+### tf.distribute.Strategy
+
+Distribute strategy will be user module's responsibilty with the new generic
```
anj-s: s/distribute/distribution

rfcs/20200117-tfx-generic-trainer.md
```diff
@@ -0,0 +1,252 @@
...
+trainer interface. To use it, user needs to modify the `run_fn()` in the module
+file, below shows the pseudo code example for single worker and multi-worker
```
anj-s: Do we want to say that the pseudo code is only applicable for single worker?

rfcs/20200117-tfx-generic-trainer.md
```diff
@@ -0,0 +1,252 @@
...
+
+```python
+def _is_chief() -> bool:
+  """Decide whether the current worker's role is chief."""
+  # Check TF_CONFIG (set by TFX when bring up the worker) in execution env.
+  ...
+
+def run_fn(args: trainer.executor.TrainerFnArgs) -> None:
+  """Build the TF model and train it."""
+  ps_strategy = tf.distribute.experimental.ParameterServerStrategy()
```
anj-s: we currently don't have support for this in TF 2.0. Can we replace this with `tf.distribute.experimental.MultiWorkerMirroredStrategy` ?

rfcs/20200117-tfx-generic-trainer.md
anj-s: Is there a reason we need to condition on the chief? If we use the multiworker strategy that I referred to below we don't need this condition I think.

rfcs/20200117-tfx-generic-trainer.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+### tf.distribute.Strategy
+
+Distribute strategy will be user module's responsibilty with the new generic
+trainer interface. To use it, user needs to modify the `run_fn()` in the module
+file, below shows the pseudo code example for single worker and multi-worker
+distribute strategy.
+
+For single worker distribute strategy, you need to create an appropriate
```
anj-s: We have 2 single worker strategies - MirroredStrategy (multiple-GPUs) and CentralStorageStrategy(variables on the CPU, computation on the GPUs) that we can maybe specify here.

rfcs/20200117-tfx-generic-trainer.md
```diff
@@ -0,0 +1,252 @@
...
+    #   6. optional custom config for passing params from component.
+    run_fn_args = self._GetRunFnArgs(
+        input_dict, output_dict, exec_properties)
+
+    run_fn(run_fn_args)
+    # Validates the existence of run_fn's output savedmodel.
+    ...
+```
+
+In module file, user needs to provide `run_fn` instead of previous `trainer_fn`.
```
lc0: From API viewpoint, `run_fn` sounds very ambiguous. At least for me, it does not convey, what is expected to happen inside given the semantics of `run`

rfcs/20200117-tfx-generic-trainer.md
```diff
@@ -0,0 +1,252 @@
...
+The `trainer_fn` was responsible for creating the model, in addition to that,
+`run_fn` also needs to handle training part and output the trained model to a
+desired location given by run args:
+
+```python
+def run_fn(args: trainer.executor.TrainerFnArgs) -> None:
```
lc0: would `args` be of type `TrainerFnArgs` or shall we update type hints here as well?

1025KB: we can reuse TrainerFnArgs as it's basically just a dict

rfcs/20200117-tfx-generic-trainer.md
1025KB: here we use the same name as internal generic trainer, it would be good to align with internal for user interface in case of future convergence

rfcs/20200117-tfx-generic-trainer.md
1025KB: if we only want to do certain thing once? for example, we only want to save model once

rfcs/20200117-tfx-generic-trainer.md
1025KB: it should also work for multi-worker on cloud

rfcs/20200117-tfx-generic-trainer.md
1025KB: how about tpustrategy and onedevicestrategy?

rfcs/20200117-tfx-generic-trainer.md
1025KB: for MultiWorkerMirroredStrategy we don't need to check chief, right? 

is there any multi-worker strategy that we only want chief to save the model? otherwise this example would be same as single worker.

rfcs/20200117-tfx-generic-trainer.md
1025KB: sg, will check

anj-s: Sure, we can work on figuring out a general way of doing this.

theadactyl:
Notes from today's design meeting at 2pm PT: 

- [Jiayi] Single worker distribution strategy supported 
- [Jiayi] Multi-worker distribution strategy is not supported yet
- [Jiayi] Need to be able to be able to see if worker is chief   
- [Jiayi] In cloud we're able to bring up the cluster to run multi-worker, but we don't have it on prem right now
- [Jiayi] Check chief for multiworkermirrored strategy? Used to have to check, but wont need to in the future 
- [azacks] For generic training, good to still have is_chief
- [azacks] Strategy itself should have to say whether is chief or not
- [Jiayi] TPU strategy or onedevice? Are these also single worker?
- [goutham] yes 
- [Jiayi] Run function name - to align with internal, we want to align on naming. Any suggestions?
- [goutham] trainer evaluate function we've used for trainer
- [goutham] As long as it's properly documented with many code example it should be fine
- [Jiayi] Trainer_fn and run_fn are both for trainer, alright reuse TrainerFnArgs dict 
- [Jiayi]After we deprecated trainer_fn it will look better for  run_fn to use TrainerFnArgs
- [azacks] when we do save, directory is part of environment. Is this already part of trainer args?
- [azacks]We should show an example to make sure people are saving to the right location and more consistent api to avoid those mistakes
- [tf ranking] To publish tfx template, should we use native keras or model estimator?
- [azacks] we're not quite there yet for fulling supporting templates in native keras
- [jiayi] Some of estimator functionality isnt supported  in native keras, there's a bug tracking that; if your use case is simple (one worker strategy), go with native keras
- [jiayi] Model estimator doesn't support tpu strategy - oss example is coming, works internally right now

PR #202, opened 2020-02-06
RFC: Standalone Keras Repository

qlzh727:
Comment period is open till 2/25/2020.

# Standalone Keras Repository

| Status        | Proposed |
| **RFC #**     | [202](https://github.com/tensorflow/community/pull/202) |
| **Author(s)** | Qianli Zhu (scottzhu@google.com), Francois Chollet (fchollet@google.com) |
| **Sponsor**   | Karmel Allison (karmel@google.com) |
| **Updated**   | 2020-02-05                         |

## Objective

Move the Keras code from the TensorFlow main GitHub repository to its own
repository, with TensorFlow as a dependency.


rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+
+The current usage of Keras from TensorFlow are:
+* Unit tests, which should be converted to integration tests, or port the tests
+to Keras repository.
+* `feature_column`.
+* Legacy `tf.layers` in v1 API.
+* legacy RNN cells.
+* TPU support code for `optimizer_v2`.
+* SavedModel.
+* TF Lite.
```
pavithrasv: I think we should add the aliased modules to this list? losses, metrics, initializers, optimizers ...

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+
+
+### Performance Implications
+
+There may be some performance implications as we move towards only using
+public TF APIs. We need to maintain a benchmark to ensure that there
+is no performance regression.
+
+### Dependencies
+
```
pavithrasv: Can you explain how namespacing will work without a circular dep? 
Keras depends on TF, will Keras continue to be under tf namespace as `tf.keras`?

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+During this conversion, we might notice that certain TF features used in Keras are
+not public. A decision should be made on a case-by-case basis:
+
+* Copy the functionality from TF to Keras.
+* Replace the usage with another alternative TF public API.
+* Make the functionality a new TF public API.
+
+**Note that the open-source community is encouraged to contribute to this effort.**
+
+### Two-stage change process
```
gabrieldemarmiesse: Thanks! Great description of the biggest problems we may encounter.

I would like us to consider targeting tensorflow stable (or even release candidates). This is the move we made in Tensorflow addons.

Some of the points to support this idea:

* As said in the meeting, changes happening in both tensorflow and keras at the same time are very rare.
* The second scenario should not happened as said in this RFC (or we should work to avoid it), because we shouldn't rely on private implementations of tensorflow.
* Pinning the version number of dependencies in the CI is considered a best practice.
* Sudden breaking/bugs in the TF side can make our build fail for a undertimined number of days (until we find the root cause) and during this period of time we can't merge any pull requests.
* For local testing, developers need to download tensorflow nightly every day.
* The build is not deterministic across time (I can't reproduce easily the tests from 8 days ago let's say)
* Since the tests are not deterministic, we can't use tools like git biscect to catch the commit which introduced a bug in the codebase.

See the move we made for tensorflow addons:
https://github.com/tensorflow/addons/pull/912

Since I understand that setting in stone the rule of targeting TF stable is very challenging for rapid developement (scenario 1 and 2), I propose a solution in between:

* We can target the stable or the nightly version, but the version must always be pinned. e.g. in the CI we have `pip install tensorflow==2.1.0` or `pip install tf-nightly==2.2.0.dev20200207`.
* When scenario 1 and 2 presented in this RFC are not happening, we target the stable version: `pip install tensorflow==2.1.0`

If scenario 1 or 2 happens:

1) Keras is targeting `tensorflow==2.1.0`
2) User makes the PR 22222 to tensorflow
3) PR 22222 gets merged
4) tf-nightly 2.2.0.dev20200207 gets released
5) PR 33333 is made to keras. We need to be able to merge this feature quickly.
6) The PR exceptionally change the pinned version of tensorflow targeted in the CI `tensorflow==2.1.0` ->  `tf-nightly==2.2.0.dev20200207`
7) The PR is merged.
8) Tensorflow 2.2.0 is out.
9) We make a PR to change the pinned version in the CI `tf-nightly==2.2.0.dev20200207` -> `tensorflow==2.2.0`.

With this, we have the best of both worlds. We target a stable version most of the time, and even when we don't, the version is pinned so the tests are deterministic.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+### Dependencies
+
+The TensorFlow pip package will auto-install the Keras package, which shouldn't make 
+any difference on the end-user side. Under the hood, Keras will be a different 
+package imported by `tf_core`, like what we do for TF estimator.
+
+### Developer experience Impact
+
+* The local build and test times should be greatly reduced, since compiling TF
+is no longer needed, and Keras is a pure-Python project.
```
gabrieldemarmiesse: Can we ditch Bazel and use pytest instead? 😃

rfcs/20200205-standalone-keras-repository.md
gabrieldemarmiesse: Good question, will we be able to do `import keras`?

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+
+With the new standalone Keras repository, external contributors
+should experience much shorter turn-around time when 
+building/testing Keras, since they don't need to build TensorFlow anymore.
+This should  have a positive impact on building a vibrant open-source
+developer community.
+
+In addition, by getting the Keras team at Google to start developing Keras
+using the same public tools and infrastructure as third-party developers,
+we make the development process more transparent and more community-oriented.
```
gabrieldemarmiesse: 👍 x1000

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Questions and Discussion Topics
+
+1. Tools for issue tracking: we can't rely on Google-internal bug tracking tool 
+since it's not publicly visible, also if managing GitHub issues across the orgs 
+is hard, we might need to find some alternatives for tracking bugs/features etc.
+2. OSS tests for TPU-related code. Since TPUs are not available during local
+testing, the verification will have to be done when the PR is mirrored to Google's 
+internal systems.
+3. Transition period for moving the Keras code from `tensorflow/tensorflow` to
```
bhack: We have already many duplicates Keras and TF.keras issue/pr. 
We will need also to cross repositories porting issue/PR that user will still going to open in the core repositories probably for months (It Is reccomended to change TF issue/PR templates to create a Keras specific note)

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+joining the other Keras SIG projects and replacing the current external Keras 
+codebase. `tf.Keras` will also replace Keras on PyPI.
+
+Also considered: `tensorflow/keras`.
+
+| keras-team/keras   | tensorflow/keras |
+:------------------- |:------------------------------------------- |
+|Under the umbrella of Keras SIG, which hosts all other Keras related projects like keras-application, KerasTuner etc.|Under the umbrella of tensorflow, which also hosts other TF related projects.|
+|Lots of existing followers on keras-team, who may not be easily migrated to TF project.|No cross org repo management cost on GitHub. Could rely on a lot of existing setup in TensorFlow.|
+|Can't easily delete keras project, which already have tons of stars and incoming reference links. Continued existence of external Keras code will create confusion ("why is there tensorflow/keras AND keras-team/keras?")|Issue/PR under the same org can be transferred easily, but not cross the different org. See here|
```
gabrieldemarmiesse: The link is missing for "see here".

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+GitHub: the code will live at [keras-team/keras](https://github.com/keras-team/keras), 
+joining the other Keras SIG projects and replacing the current external Keras 
+codebase. `tf.Keras` will also replace Keras on PyPI.
+
+Also considered: `tensorflow/keras`.
+
+| keras-team/keras   | tensorflow/keras |
+:------------------- |:------------------------------------------- |
+|Under the umbrella of Keras SIG, which hosts all other Keras related projects like keras-application, KerasTuner etc.|Under the umbrella of tensorflow, which also hosts other TF related projects.|
+|Lots of existing followers on keras-team, who may not be easily migrated to TF project.|No cross org repo management cost on GitHub. Could rely on a lot of existing setup in TensorFlow.|
```
gabrieldemarmiesse: > No cross org repo management cost on GitHub

Could we have a list of those costs? That would help in the decision process.

bhack:
Please announce this on the TF dev mailinglist(s) as usual. /cc @ewilderj

pavanky:
Will the ci still run keras tests when changes are made to the tensorflow repository ?

rfcs/20200205-standalone-keras-repository.md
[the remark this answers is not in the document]
gabrieldemarmiesse: I can merge pull requests in keras-team/keras and I think it's fair that I loose this right when the new codebase is migrate. It's a different codebase altogether, I'll need to make many pull request to understand the new codebase.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+new version of `tf-nightly`, if  certain behavior has been changed and wasn't
+caught by unit tests. We have  observed a few similar cases with
+[tf/addons](https://github.com/tensorflow/addons).
+We hope this can be reduced by stronger
+unit test coverage by Google internel systems, when both TF and Keras code are 
+tested at HEAD.
+* pip package management. Keras will now follow the `tf-estimator` approach. 
+"pip install tensorflow" should also install Keras (from PyPI) as well.
+There are more details for the pip package in the
+[Improved pip package structure](https://github.com/tensorflow/community/pull/182) RFC.
```
AakashKumarNain: Can we also consider`conda` this time? `Anaconda` is a huge part of Data Science and ML ecosystem and `tf.keras` not being a part of it doesn't feel right.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+
+* Copy the functionality from TF to Keras.
+* Replace the usage with another alternative TF public API.
+* Make the functionality a new TF public API.
+
+**Note that the open-source community is encouraged to contribute to this effort.**
+
+### Two-stage change process
+
+For any change that is affecting both TensorFlow and Keras, the change
```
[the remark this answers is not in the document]
gabrieldemarmiesse: Can you detail what is the cache you are mentioning?

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+developers, e.g. in-flight PRs during the transition period.
+* For current Keras pip package users, they will get the new TF keras package 
+when they update their pip, which should have more features than the current 
+Keras-team version.
+
+## Questions and Discussion Topics
+
+1. Tools for issue tracking: we can't rely on Google-internal bug tracking tool 
+since it's not publicly visible, also if managing GitHub issues across the orgs 
+is hard, we might need to find some alternatives for tracking bugs/features etc.
```
gabrieldemarmiesse: When you say "is hard", is it that you can't transfer issues across organisations ? Are there other pain points?

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+2. OSS tests for TPU-related code. Since TPUs are not available during local
+testing, the verification will have to be done when the PR is mirrored to Google's 
+internal systems.
+3. Transition period for moving the Keras code from `tensorflow/tensorflow` to
+`keras-team/keras`. All in-flight PRs / issues will be affected: they need
```
gabrieldemarmiesse: >  All in-flight PRs / issues will be affected

Would it be possible to do a progressive migration? Class after class and function after function or would it become too complicated?

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+has to rely on a number of private TensorFlow APIs. However, a litmus test
+of the quality of the public TensorFlow low-level APIs is that they should
+be strictly sufficient to a higher-level API like Keras.
+After splitting the repository, Keras will have to import TensorFlow and 
+rely exclusively on public APIs. If Keras still ends up using TensorFlow
+private features, it  might be an indication of tight coupling of
+implementation details. If certain private features are extensively used,
+we might want to consider exposing them  as public low level API.
+
+This design is also aligned with thee design for
```
Dref360:
```suggestion
This design is also aligned with the design for
```

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+```
+
+During this conversion, we might notice that certain TF features used in Keras are
+not public. A decision should be made on a case-by-case basis:
+
+* Copy the functionality from TF to Keras.
+* Replace the usage with another alternative TF public API.
+* Make the functionality a new TF public API.
+
+**Note that the open-source community is encouraged to contribute to this effort.**
```
gabrieldemarmiesse: Very nice description of the problem. Since there is a lot of decision making going on there, I believe it would be nice for people in charge of decisions to open issues describing what they want to do for all the unwanted imports we'll have. It's hard to help when you don't know what will be decided to resolve the problem (here option 1? 2? or 3?). 

If we don't do that, I believe the open-source community won't help much just because of a lack of guidance.

rfcs/20200205-standalone-keras-repository.md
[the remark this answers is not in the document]
Dref360: I'm also in favor of restricting members of keras-team to issues for now. 
I hope many people from the team will come back to Keras once the move is done, but it is true that the codebase is really different now.

rfcs/20200205-standalone-keras-repository.md
Dref360: If we do the same thing as we did for keras-prepro and keras-app, it should work right?

rfcs/20200205-standalone-keras-repository.md
Dref360: I think that would be a great change! 
Also return to travis/circleci.

rfcs/20200205-standalone-keras-repository.md
gabrieldemarmiesse: @Dref360 can you detail a bit? I'm sure there are people not familiar with your work (including me 😄 )

rfcs/20200205-standalone-keras-repository.md
gabrieldemarmiesse: Even better...

DOCKER BUILD ALL THE THINGS: https://github.com/tensorflow/addons/pull/1053

rfcs/20200205-standalone-keras-repository.md
qlzh727: Anna's design covers more details about the PIP package structure. https://github.com/tensorflow/community/pull/182. Currently TF package is an umbrella package, and the tensorflow_core is the package that contains all the real code. The new Keras package will depend on tensorflow_core, and TF umbrella package will depend on Keras.

This the same approach used for tensorflow_estimator, which relies on Tensorflow ops, but still exposed under tf.estimator.

qlzh727: @gabrieldemarmiesse, yes. tf.keras symbol will still work, and so as keras. Under the hood, tf.keras will load the python symbol from keras.

rfcs/20200205-standalone-keras-repository.md
yongtang: GitHub Workflow Actions is a good choice as well. From tensorflow SIGs the overall experience is that GitHub Workflow Actions is much better than TravisCI (Workflow Actions might be similar to CircleCI with respect to use experience).

rfcs/20200205-standalone-keras-repository.md
gabrieldemarmiesse: It would be nice to have 
```python
import keras
```

instead of 
```
import tensorflow_keras as keras
```
But if you prefer @qlzh727 we can move the discussion to #182

rfcs/20200205-standalone-keras-repository.md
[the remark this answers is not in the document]
bhack: https://github.com/tensorflow/community/pull/202#discussion_r376597059

rfcs/20200205-standalone-keras-repository.md
qlzh727: Yes, I had offline discussion with Anna before, and we would prefer keras instead of Keras. I think when Anna wrote her design, she was not sure about what package we will use, and she assumed we will use the similar name as tensorflow_estimator, which is why she put tensorflow_keas in her design. @annarev for #182.

gabrieldemarmiesse: Very nice, thanks! Some users won't even see the transition when upgrading their keras multi-backend ;)

rfcs/20200205-standalone-keras-repository.md
qlzh727: Thanks @gabrieldemarmiesse for the detailed suggestion. I do like the idea of pinning to a certain version which should greatly reduce the chance of breakage. I think pinning the version (nightly or stable) is a very good alternative for reproducibility and supporting the rapid development.

rfcs/20200205-standalone-keras-repository.md
qlzh727: @gabrieldemarmiesse, I think at this stage, we can't remove bazel and use pytest since it will affect too much of the Google internal builds (we rely on bazel heavily internally). Wrt to CI choice, I think we are open for that, and I will take a look how github workflow action is setup for other SIGs. 

@seanpmorgan, for tf-addons, how is your current experience with Github workflow actions? Does it meet all the need? eg showing build history, GPU testing, presubmit notification, etc?

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+cache might help speed up the follow-up builds, the initial time cost is too 
+high for regular software development workflows. Internally, Google has a
+distributed build and caching service, which Googlers heavily rely on,
+that can build TensorFlow and run all Keras tests within 5 mins. Sadly,
+we can't expose this to external contributors.
+
+Currently, any contribution to Keras code will require building all of
+TensorFlow, which is quite expensive to do for average users.
+Having a separate repository will allow the Keras package to be built
+without building TensorFlow. This should greatly improve the
```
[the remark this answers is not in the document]
qlzh727: Ack, assuming this is just a generic comment and doesn't need to be addressed.

Dref360:
@googlebot I fixed it.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Ack, will put them as list, which makes it easy to comment.

rfcs/20200205-standalone-keras-repository.md
gabrieldemarmiesse: > I think at this stage, we can't remove bazel and use pytest since it will affect too much of the Google internal builds (we rely on bazel heavily internally).

I understand it's a big task right now, especially since all the test helpers of TF are designed for unittest: https://www.tensorflow.org/api_docs/python/tf/test/TestCase I hope we can work towards being less bazel-dependant in the future, as it's very limiting for python-only depositories. Thanks for your answer!

gabrieldemarmiesse:
@googlebot I fixed it.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Design Proposal
+
+### New location of the code
+
+GitHub: the code will live at [keras-team/keras](https://github.com/keras-team/keras), 
+joining the other Keras SIG projects and replacing the current external Keras 
+codebase. `tf.Keras` will also replace Keras on PyPI.
+
+Also considered: `tensorflow/keras`.
```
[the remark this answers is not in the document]
qlzh727: Updated for the section for repo comparison. I think the we lean towards keras-team/keras since it already has a lot of uses/traffic/stars and incoming references.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+
+The TensorFlow pip package will auto-install the Keras package, which shouldn't make 
+any difference on the end-user side. Under the hood, Keras will be a different 
+package imported by `tf_core`, like what we do for TF estimator.
+
+### Developer experience Impact
+
+* The local build and test times should be greatly reduced, since compiling TF
+is no longer needed, and Keras is a pure-Python project.
+* Cross-boundary changes will require some extra handling since such changes
```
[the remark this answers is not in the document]
qlzh727: Yes, this is intended. Having said that, we don't expect this to happen a lot, since most of the TF/keras PR only touch either side, not both at the same time.

rfcs/20200205-standalone-keras-repository.md
yongtang: @qlzh727 I can help with GitHub actions if needed. You can see either the actions history either in SIG IO (https://github.com/tensorflow/io/actions) or SIG Addone (https://github.com/tensorflow/addons/actions)

The GPU testing is missing for GitHub Actions as it does not have the GPU machine resource. However, it is still possible to bring self-hosted runner and wire up with GitHub actions. In other words, it could be possible to wire up Kokoro's GPU machine with GitHub action I believe. Some additional effort is needed to make this work though.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Done.

rfcs/20200205-standalone-keras-repository.md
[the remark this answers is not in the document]
bhack: Can we make a quite common example?
I.e. I want to add a new layer for Deformable convolution https://github.com/tensorflow/addons/issues/179.
Is It Keras Is It Tensorflow?

rfcs/20200205-standalone-keras-repository.md
qlzh727: Updated with 86d2c37. I will leave this open since your comment contains more details.

qlzh727:
> Will the ci still run keras tests when changes are made to the tensorflow repository ?

It will only run tests in TF repository that relies on some Keras functionality, eg base_layers. Since Keras rely on TF as a dependency, we will not run any Keras for any PR to TF.

qlzh727:
@Dref360, could check your CLA setting? 

frederic.branchaud-charron@usherbrooke.ca | need_author_cla

rfcs/20200205-standalone-keras-repository.md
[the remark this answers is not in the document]
qlzh727: Thanks all for pointing this problem out. Since we are reusing the existing keras project with totally new code base, I think we will need to cleanup any existing PR/Issue and group since many of them maybe incorrect with new code. I will update this RFC with more details.

rfcs/20200205-standalone-keras-repository.md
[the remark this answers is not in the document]
bhack: Will this be the only "core project" outside Tensorflow/? Have we other sigs with out of Tensorflow/ repo?
Cause It will be really hard to explain why this will be the only "brand" that It is so core but still something like an external project.
I think the story telling will not work so much.

rfcs/20200205-standalone-keras-repository.md
qlzh727: @gunan from TF build team, and also @seanpmorgan who mentioned this in the SIG-addon meeting.

Dref360:
@googlebot I fixed it.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+
+* No end-user facing change for current TF users; the split would only affect
+developers, e.g. in-flight PRs during the transition period.
+* For current Keras pip package users, they will get the new TF keras package 
+when they update their pip, which should have more features than the current 
+Keras-team version.
+
+## Questions and Discussion Topics
+
+1. Tools for issue tracking: we can't rely on Google-internal bug tracking tool
```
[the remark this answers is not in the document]
qlzh727: Thanks for the notice. I will check with infra team and support team to see if they have any suggestions for improving this situation. Will leave this open for discussion in the design meeting.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Yes, this is mainly refer to the transferring the issue across the org.

rfcs/20200205-standalone-keras-repository.md
qlzh727: We want to minimize the period of transition, since during the period, there will be kind of two source of truth, which might confuse user which one to use/contribute. I will update this RFC with more details about the PR/issue transition.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,361 @@
...
+* Keras CI/presubmit build for the GitHub repo should target a stable PIP 
+version of tensorflow package as dependency. It could either be a `tf-nightly`
+with explicit version, or a release candidate version, or a stable version.
+Depend on a floating `tf-nightly` could cause CI build to be instable, which has
+been observed in other repository 
+[like tf-addons](https://github.com/tensorflow/addons/pull/912).
+* The Keras code will be mirrored to a Google-internal code repository via
+Google-internal tools within a very short time window after each change.
+The Google-internal CI tests will run on HEAD for both Keras and TF code.
+* The CI build for the repository on GitHub might break when it points to a
```
gabrieldemarmiesse: I don't believe this point will occur when we test against pinned versions of tensorflow (especially stable). I believe it can be removed.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,361 @@
...
+
+* We expect the majority of the code development/contribution from GitHub
+and the dev tools / tests / scripts should focus on the GitHub development use
+case. See below for more details.
+* Keras CI/presubmit build for the GitHub repo should target a stable PIP 
+version of tensorflow package as dependency. It could either be a `tf-nightly`
+with explicit version, or a release candidate version, or a stable version.
+Depend on a floating `tf-nightly` could cause CI build to be instable, which has
+been observed in other repository 
+[like tf-addons](https://github.com/tensorflow/addons/pull/912).
```
gabrieldemarmiesse: Can we add:  We target preferably in this order:
* a stable version
* a pre-release
* a nightly version.
Using a nightly version for testing should be motivated by the usage of a API feature not present in the stable or pre-release version.
The version should be pinned all the time for tensorflow stable or nightly.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,361 @@
...
+* The Keras code will be mirrored to a Google-internal code repository via
+Google-internal tools within a very short time window after each change.
+The Google-internal CI tests will run on HEAD for both Keras and TF code.
```
gabrieldemarmiesse: I don't believe it's a good software practice because of sudden build breaks. But it's Google's problem. I just hope it won't come back to bite the community later (like with bazel) with a justification along the lines of "we do that internally so the community has to do the same too.".

rfcs/20200205-standalone-keras-repository.md
gabrieldemarmiesse: Thanks for the answer! I hope it'll work out! Definitly not an easy task in both cases.

gabrieldemarmiesse: Thanks for the answer! I hope it'll work out! Definitly not an easy task in both cases.

rfcs/20200205-standalone-keras-repository.md
qlzh727: The build failure might still occur if we point to a newer version of TF, but probably caught by the presubmit test.

gabrieldemarmiesse: And in this case it never ends up in the master branch without the appropriate fix, so it should be fine indeed.

rfcs/20200205-standalone-keras-repository.md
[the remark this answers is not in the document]
qlzh727: Done. Updated with a new sections about existing PR/issue and permission for the repo.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+Keras team members at Google.
+
+### Best Practices, Tutorials and Examples
+
+* The new Keras repository should have a new contribution guide about how to 
+setup a local test environment and iterate based on that. A similar one in 
+tf/addons can be used as an example.
+* The existing TF docs needs to be updated to highlight that Keras code now lives
+in a different repository, with a new process for sending PRs, etc.
+* When filing an issue, people might need to consider where to send the issue,
```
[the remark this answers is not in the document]
qlzh727: Done. Updated with a new sections about existing PR/issue and permission for the repo.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Added a new section about existing PR and issue. We hope to cleanup the keras-team/keras first and bring any applicable PR/issue to TF keras first.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Done. Updated.

gabrieldemarmiesse: Thanks a lot!

gabrieldemarmiesse:
Thanks @qlzh727 for all your work! We're working you to the bone,  sorry about that. I have two questions left:

Will the git history of tf.keras be preserved? 
Can you format the code to use 4 spaces instead of 2 for indents as you copy the code? Some tools like black can do that automatically.
It would be a good time to do this since we're doing a one shot transition. It's going to be much harder to do it later.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,394 @@
...
+Pros:
+1. Under the umbrella of Keras SIG, which hosts all other Keras related projects
+like keras-application, KerasTuner etc.
+1. Lots of existing followers on keras-team, who may not be easily migrated to 
+TF project.
+1. Can't easily delete keras project, which already have tons of stars and 
+incoming reference links. Continued existence of external Keras code will create
+confusion ("why is there tensorflow/keras AND keras-team/keras?").
+
+Cons:
```
bhack: Does keras will be the only Tensorflow "core project"/sig  that is external to Tensorflow/?
What is the storytelling here?
I suppose the other one is MLIR but it is not only dedicated to Tensorflow and now it is under LLVM that is a standalone foundation so the rationale is different.

rfcs/20200205-standalone-keras-repository.md
gabrieldemarmiesse: My comment has been adressed. Could you minimize this thread for readability? Thanks!

rfcs/20200205-standalone-keras-repository.md
annarev: what @qlzh727 is right, I didn't mean to imply `tensorflow_keras` as the package name to use in my doc, just used it similarly with `tensorflow_estimator`.

gabrieldemarmiesse: Can we add it somewhere in the document to set it in stone? :)

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+[Improved pip package structure](https://github.com/tensorflow/community/pull/182) RFC.
+
+### Dependency Cleanup
+
+As the high-level API of TensorFlow, Keras should have a direct dependency on
+TF low-level APIs, but not the other way around. Unfortunately, there is some existing reverse 
+logic in the TF code that relies on Keras, which we should update/remove 
+when we split the repository.
+
+The current usage of Keras from TensorFlow are:
```
tomerk: A pretty painful one dependency on Keras from TensorFlow not listed here:
The symbolic vs eager logic for ops in TensorFlow is very tightly coupled w/ Keras due to the current implementations of the functional API and TF Op layers

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+* Unit tests, which should be converted to integration tests, or port the tests
+to Keras repository.
+* `feature_column`.
+* Legacy `tf.layers` in v1 API.
+* legacy RNN cells.
+* TPU support code for `optimizer_v2`.
+* SavedModel.
+* TF Lite.
+
+All Keras imports in integration tests can be changed to use dynamic import like below:
```
tomerk: We should be pretty careful about how we structure the integration tests & what we make them depend on to avoid this where possible.

rfcs/20200205-standalone-keras-repository.md
tomerk: The build times will be greatly reduced, but Keras is unlikely to remain pure-python forever.

We may also experience a future where Keras contains custom ops that aren't in core TF (similarly to Addons).
Right now that's very painful to sync internally & externally because the custom ops are handled in a different way, so we may end up having to work on improving that.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,351 @@
...
+* Replace the usage with another alternative TF public API.
+* Make the functionality a new TF public API.
+
+**Note that the open-source community is encouraged to contribute to this effort.**
+
+### Two-stage change process
+
+For any change that is affecting both TensorFlow and Keras, the change
+will need to be split into two, one as a PR to the TF repo,
+and the other as a PR to the Keras repo. Here are some common scenarios:
```
tomerk: This makes me extremely nervous as long as there are any cross-dependencies between TF and Keras that aren't purely public APIs... Counter-intuitively this extra process will make it *harder* to remove the cross-dependencies between TF and Keras.

How many of these cross-dependencies do we plan to resolve before the standalone Keras repo is made?

How many of the cross-dependencies do we intend to fix *before*

rfcs/20200205-standalone-keras-repository.md
gabrieldemarmiesse: This has been adressed by @annarev in https://github.com/tensorflow/community/pull/182#discussion_r377591797 @qlzh727 I believe you can minimize this thread.

rfcs/20200205-standalone-keras-repository.md
gabrieldemarmiesse: @qlzh727 It seems we all agree here and this can be discussed after the split since it's an independant subject. Could you minimize this thread? Thanks!

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,394 @@
...
+Building the open-source TensorFlow project end-to-end is an extensive exercise. 
+With a standard GCP instance, it might take more than one hour to finish the whole
+build process (it might take longer with a Mac laptop). Although the local build 
+cache might help speed up the follow-up builds, the initial time cost is too 
+high for regular software development workflows. Internally, Google has a
+distributed build and caching service, which Googlers heavily rely on,
+that can build TensorFlow and run all Keras tests within 5 mins. Sadly,
+we can't expose this to external contributors.
+
+Currently, any contribution to Keras code will require building all of
```
alextp: These requirements around build time apply to all of TF's python APIs, not just Keras.

So to me this suggests we want a more aggressive separation between python TF and C++ TF's build and test systems, not anything specifically to do with keras.

For example, if we required that every TF python test passed with the previous nightly's pywrap_tensorflow.so across the TF python surface we could dramatically simplify the development for much more than keras.

I don't mean to delay / disrupt the keras migration process, which I think is a good idea for other reasons, I just don't want us to waste the opportunity to improve the development experience.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,394 @@
...
+TensorFlow, which is quite expensive to do for average users.
+Having a separate repository will allow the Keras package to be built
+without building TensorFlow. This should greatly improve the 
+velocity of open-source developers when they contribute to Keras code.
+
+### Community Benefit
```
alextp: Isn't this the same as the build time section above?

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,394 @@
...
+should experience much shorter turn-around time when 
+building/testing Keras, since they don't need to build TensorFlow anymore.
+This should  have a positive impact on building a vibrant open-source
+developer community.
+
+In addition, by getting the Keras team at Google to start developing Keras
+using the same public tools and infrastructure as third-party developers,
+we make the development process more transparent and more community-oriented.
+
+### TensorFlow API modularity
```
alextp: That to me is the main motivation which could justify splitting the repositories (the build time one just suggests we should split the continuous integration tooling, not the repositories).

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,394 @@
...
+private features, it  might be an indication of tight coupling of
+implementation details. If certain private features are extensively used,
+we might want to consider exposing them  as public low level API.
+
+This design is also aligned with the design for
+[Modular TensorFlow](https://github.com/tensorflow/community/blob/master/rfcs/20190305-modular-tensorflow.md), 
+which splits the TensorFlow project into smaller components that are not
+tightly coupled together.
+
+
```
alextp: There are also other advantaged worth mentioning. For example a keras repository makes it easier to share some of the management with the keras SIG than the single shared TF repository.

We need to also consider drawbacks of this decision here.

Specifically:
 1. cross-repository changes will be more expensive to submit (and now most keras changes would cross the repository boundary)
 2. CI will be more complicated (we'll need to test a product of TF and keras versions as opposed to a single shared version). This is complicated further by pulling these two repositories inside google's version control system, for some of us.
 3. Previous experience splitting tf.estimator with the same type of argument justifying it has been considered a failure. How are we sure keras won't fall prey to the same issues?

alextp: Specifically, I'd like to see this proposal cover alternative approaches to the problems we've outlined, including a single repository with independent continuous integration, a tf-python repository separate from a tf-core repository, etc.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,394 @@
...
+Likewise, issue tracking and code review takes place through Google-internal tools.
+
+For Keras, since we are trying to promote community engagement, we hope to use 
+GitHub as source of truth. This will have the following implications:
+
+* We expect the majority of the code development/contribution from GitHub
+and the dev tools / tests / scripts should focus on the GitHub development use
+case. See below for more details.
+* Keras CI/presubmit build for the GitHub repo should target a stable PIP
```
alextp: This has the disadvantage of slowing down keras's adoption of new TF features and APIs.

This document should clarify when would CI against tf stable be expected to break, etc.

alextp: (also important to highlight the release process for TF)

rfcs/20200205-standalone-keras-repository.md
alextp: +1.

This document needs a plan to break this dependency before it can be seriously considered.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,394 @@
...
+
+As the high-level API of TensorFlow, Keras should have a direct dependency on
+TF low-level APIs, but not the other way around. Unfortunately, there is some existing reverse 
+logic in the TF code that relies on Keras, which we should update/remove 
+when we split the repository.
+
+The current usage of Keras from TensorFlow are:
+* Unit tests, which should be converted to integration tests, or port the tests
+to Keras repository.
+* `feature_column`, which uses Keras base layer and model.
```
alextp: These dependencies are very concerning.

Does this mean we need a third package (tf, tf-keras, and tf-stuff-that-depends-on-keras)? Where would the code for this third package live? How can we test its features?

Or are splitting all the things which depend on keras despite not being historically a part of the keras API into the keras repo?

rfcs/20200205-standalone-keras-repository.md
alextp: I have strong reservations about this type of cross-package circular dependency.

rfcs/20200205-standalone-keras-repository.md
tomerk: Chatted w/ @alextp a bit about this,

The following are all really important goals:

1. Make it easy for the community to contribute to & manage Keras (without having to build all the rest of TF to do so)
2. Layer the external TF API generation so Keras can be transitioned to using entirely public TF symbols (with actual tf.* calls)
3. Enable building a keras-only pip package separate from the standard tf package, which users will be able to use via just `import Keras`

It would be useful to explore possible solutions that don't create the same degree of development / release syncing overhead.

E.g. in addition to @alextp's suggestions above, we could explore some sort of transitional plan that partially or mostly accomplishes the above 3 goals while keeping Keras and TF in the same repo w/ a synced release cycle. 
Once all the circular dependencies have truly been fixed and Keras already uses the public TF APIs directly, we could revisit splitting Keras into a fully separate repo w/ its own release cycle and source of truth.

rfcs/20200205-standalone-keras-repository.md
gabrieldemarmiesse: It's not supposed to slow down adoption of new features compared to targeting an unpinned nightly all the time. Maybe it's not clear in the document, but if a pull request requires a new feature in tensorflow nightly, the person making the pull request can change the targeted version of tensorflow in the CI. So we don't have to wait at all.

Do I get it right or you were referring to another problem?

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,394 @@
...
+lengthy feedback loop.
+
+With the new standalone Keras repository, external contributors
+should experience much shorter turn-around time when 
+building/testing Keras, since they don't need to build TensorFlow anymore.
+This should  have a positive impact on building a vibrant open-source
+developer community.
+
+In addition, by getting the Keras team at Google to start developing Keras
+using the same public tools and infrastructure as third-party developers,
```
gabrieldemarmiesse: I would split this point into multiple sub-points, because I believe that this argument is not sufficiently highlighted.

* Github is not made to manage monorepos like tensorflow. Keras issues and pull requests get mixed up with the tensorflow core ones. 
*  The source of truth being in github is very important because it allows the community to manage the repo. Currently, some commits pop up without any public pull request or public user, making it impossible for non-googlers to understand what is the rational behind a commit, or even who to contact if it caused a bug: e.g. https://github.com/tensorflow/tensorflow/commit/9698ae1734e065b46fbb8729ad8f850386f506e7
* As long as keras is in tensorflow, it's hard for us to have any impact on the processes/tests/tools that keras use internally. Imagine for example that we want to use Isort, or pytest, or github actions... In all cases it's a lot more work (code) and a lot more people who need to review and give us a green light. With a repository with the size of tensorflow, it's just unfeasable unless we have a face-to-face meeting.

Dref360: > Keras issues and pull requests get mixed up with the tensorflow core ones.

before the split, we had the same issue. Not sure how to fix that.

rfcs/20200205-standalone-keras-repository.md
guptapriya: +1 
the limited experience with tf.estimator suggests that we should consider the drawbacks seriously and convince ourselves that they are something we would be OK dealing with on a day to day basis (or consider alternatives). For e.g. the 2 stage submitting process (with 1 have google as source of truth and one having github as source of truth) can be onerous and error prone (accounting for versions, rollbacks etc)

x10000year:
If keras is going to be moved to a standalone repo, are v1 apis (tf.get_variable(), tf.variable_scope(), etc.) for building static graph still available in the tensorflow repo? Can we build custom models without using keras? The v1 apis are really very simple and flexible to use for building static graph, while the keras api forces us to use OOP style and gives a lot of pain for industrial engineers who are familiar with static graphs and functional style.

rfcs/20200205-standalone-keras-repository.md
qlzh727: @fchollet here.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Indeed, I would also like to cleanup this dependency before we split the repository.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Done. updated wording for pure python project.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Yes, the plan is to cleanup the dependency for the private API before we split. I have the sheet to track the usage of private TF API within Keras. So far the major block of them can be converted to public API (eg, tests utils, eager context related code etc.). The rest long tail will need some rewrite.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Ack, thanks for the suggestion.

rfcs/20200205-standalone-keras-repository.md
qlzh727: I think the third package u mentioned can be avoided by the LazyLoader. For any tf code that relies on Keras, it will need keras PIP package to be installed to work. Since PIP install TF will also install Keras package, so the keras code should always be there unless user manually changed it.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Updated in the list. I agree that we should cleanup these dependency before we split the repository.

qlzh727:
> If keras is going to be moved to a standalone repo, are v1 apis (tf.get_variable(), tf.variable_scope(), etc.) for building static graph still available in the tensorflow repo? Can we build custom models without using keras? The v1 apis are really very simple and flexible to use for building static graph, while the keras api forces people to use OOP style and gives a lot of pain for industrial engineers who are familiar with static graphs and functional style.

All the public API symbols will still remain available for v1.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Done. Updated.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Had some email discussion with TF API owner. So far the API owners are OK if the TF code that relies on Keras is V1 and only use public Keras API (eg layer and model). For feature column, we might want to move them to estimator package, and let estimator depend on keras.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Done. Updated.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Done. Thanks for the suggestion. I have added a section for CI and alternative approaches. we should discuss the alternatives in the design review.

rfcs/20200205-standalone-keras-repository.md
```diff
@@ -0,0 +1,502 @@
...
+* TF Lite for keras model saving utils.
+* Aliases from tf.losses/metrics/initializers/optimizers in tf.compat.v1.
+* Keras symbolic tensor check in the ops library for tf.function.
+
+For usage like tf.layers to keras.layers, it can't be removed due to the API
+contract and guarantee. We should use LasyLoader to walk around the cyclic
+dependency issue.
+
+```python
+from tensorflow.python.util.lasy_loader import LasyLoader
```
crccw: LazyLoader?

qlzh727: Thanks for catching this. Fixed now.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Discussed this item on the design review meeting. So far this is not active target we are supporting. The meeting conclusion are:

Conda uses a different toolchain, with no ABI compatibility between TF+Conda. Because we don’t have C interfaces, any custom ops built for TF will not work for Conda, and vice versa. There is a set of people who publish TF for Conda. But tf addons, io, etc. are not published for Conda. There’s no easy way around this because it would require doubling the TF release burden entirely.
Keras in theory should be easier? Insofar as it is Python only (currently). But Keras will in the future include C(++). That should be fine as long as there is a Python interface to the exposed parts, or C parts are used internally.

rfcs/20200205-standalone-keras-repository.md
qlzh727: Ack, I think the both Francois and others on team is aware of this issue, and as we are pushing towards OSS first, we don't expect this to come back to bite the community.

qlzh727:
@ematejska, I have updated RFC with latest result from the design meeting. Would u like to have the meeting notes also being posted here?

ematejska:
@qlzh727 Yes please.

ematejska:
Also, if you could update the status of the RFC to Accepted, that would be appreciated.  Thanks.

qlzh727:
Meeting notes from the design review on Mar 03 (taken by @karmel)

Notes

- Dependencies between TF and Keras. Two ways: Keras on non-public TF; TF backwards on Keras. Majority of TF -> Keras dependencies are in higher level TF APIs (old tf.layers; feature columns). 

  - For v1 things, eg layers, we will freeze a copy of necessary APIs (eg base layer v1) in core TF. This has the advantage of also preventing accidental changes.
  - Higher cost; requires a refactor. Relying on existing unit tests to tell us if we break something.
  - Lazy loading is easier in many ways. 
  - But, deceptively easy; may lead to integration pains down the line.
  - If we can decouple future changes to Keras from tf.layers, we will be happier.
  - Can we move v1 keras layer to tf core? Ends up being tricky-- graph/eager is the split in Keras versus v1/v2. Version selecting.
  - If we make a copy, what is the scope of the copy? Strictly to break the circular dependencies. Trim down current base_layer to reduce the amount of code copied to TF.
  - So, concretely, what would be copied for tf.layers? There is a very long blaze dep list right now, but doesn’t need to be. Eg mixed precision imports optimizers. 
  - Split out legacy tf.layers into Keras? In tf we re-export that. Requires we inject into TF namespaces from other packages, but maybe that’s not so bad. 
  - Can tf.feature_column, tf.layers be separated out and depend on keras and TF?
  - tf.lite depends on some private utils from Keras saving. Some of these can be forked. They are at least public APIs in general.
  - Suggestion: move tf.feature_column to Estimator; move tf.layers to Keras (in some sort of legacy quarantine)
  - Unit tests move either to the top-most package, or into a separate integration test package. Distribution strategies is one of the largest clients here. Would be inconvenient if those tests lived in Keras (since change has to be split); moving to integration tests outside TF seems a better solution long-term. Also allows “import tensorflow as tf”, which is preferable. Can be run on nightly and head.

- Conclusions:

  - tf.feature_column to Estimator
  - RNN Cells to Keras
  - tf.layers to Keras or frozen; whichever is less painful.
  - Integration tests to new repo
  - TPU is just instance checks; resolve with tf.types
  - TF Lite either fork utils or move to tf core
  - No LazyLoader

- Keras deps on private TF symbols

  - Head cases (count >= 5) we can resolve by using different APIs in Keras
  - Long tail (count < 5). What to do with those? Many are in dataset APIs.
  - apassos: We are less concerned about these private usages; we can address over time. We will need to get owners of private APIs on board.
  - wicke: I don’t think we should split before we have a CL that gets us to where estimator is today. This list is too long; we need to narrow the list down to fewer of these before we can realistically split.
  - scottzhu: currently I am addressing the most-used of these. Generated CL from mihai’s script that converts all public. Can take care of head cases and leave longer tail. 
  - raw_ops for ops.
  - CondBranchFuncGraph-- why do we need this? Pretty ugly check. Seems to be mostly to raise a better error. Likely what this means is that some amount of FuncGraph needs to be made public and documented so that we can allow this. Or improve error messages from core; we are solving this for Keras, but other users will have similar problems. A new public API that says, I am in a conditional branch of the graph. 
  - Much of the long-tail will be case-by-case
  - How do we avoid a proliferation of experimental APIs that we depend on? That’s better than private, since at least API owners see and govern changes to experimental. We generally require global tap, and one-release deprecation before removing.

- Source of truth is Github

  - What happens if external change to Keras can’t be brought into google3 because something internal breaks? It will happen, and it will be painful. It will eat up time, and has for Swift, MLIR, etc. We must add tests to OSS keras every time this happens. But that’s the cost of having it be OSS first.
  - That’s a cost, but also a good forcing function-- our APIs should be stable.
  - We may have to have a policy that says, if you are on private APIs, we may break you on syncs.
  - Syncing should ideally be frequent, but may fail and require manual intervention.
  - When are we ready to make the split?

### apassos: no circular deps; triaging private symbols + preventing backsliding; for the gaps, we have a plan.

  - We did all this work to bring in Keras, now we’re separating it back out. In was critical for 2.0. What will we do at the next big change? 
  - In many ways, the way we did the 2.0 change was too entangled. And would have been better if it had been more structured. But now we’ve learned and grown, and we can use that going forward.

- From OSS community: conda packages?

  - Conda uses a different toolchain, with no ABI compatibility between TF+Conda. Because we don’t have C interfaces, any custom ops built for TF will not work for Conda, and vice versa. There is a set of people who publish TF for Conda. But tf addons, io, etc. are not published for Conda. There’s no easy way around this because it would require doubling the TF release burden entirely.
  - Keras in theory should be easier? Insofar as it is Python only (currently). But Keras will in the future include C(++). That should be fine as long as there is a Python interface to the exposed parts, or C parts are used internally.

- How are we going to manage issues and PRs?

  - PRs cannot span the two repos. Must be submitted separately, without breaking CI.
  - For issues, will we transfer between the repos? We can crosslink if the issue moves to TF. Issue in Keras will stay, and link to the TF issue. 
  - We will have to reset permissions for keras repo. 
  - Automatically pull in PRs to CLs?

qlzh727:
> Also, if you could update the status of the RFC to Accepted, that would be appreciated. Thanks.

Done.

bhack:
We need to handle cross repositories PR/ISSUE but we don't have this feature naturally in Github. See https://codetree.com/guides/managing-issues-across-multiple-github-repositories

bhack:
What Is the timeline to reactive the standalone repository?

PR #207, opened 2020-02-20
RFC: Tensorflow SavedModel C/C++ API

bmzhao:
This RFC will be open for comment until Friday, March 6th, 2020.

cc @gunan @martinwicke 

# Tensorflow SavedModel C/C++ API

| Status      | Proposed                                                                                      |
| :---------- | :-------------------------------------------------------------------------------------------- |
| **RFC #**     |  207 |
| **Authors** | Brian Zhao (bmzhao@google.com), Hye Soo (hyey@google.com), Paige Bailey (webpaige@google.com) |
| **Sponsor** | Gunhan Gulsoy (gunan@google.com)                                                              |
| **Updated** | 2020-02-19                                                                                    |

## Objectives

We would like to revamp the C++ saved model user journey by

1.  Examining the set of use cases we intend to support
2.  Identifying API constraints, requirements, and pain points for each use case
3.  Designing a new C++ saved model API, while being thoughtful of evolution in the surrounding space (notably TF2 and MLIR)
4.  Creating detailed documentation for C++ API and examples.


bhack:
> This RFC will be open for comment until Friday, March 6th, 2020.
Can we extend this deadline or is it done?

ematejska:
This has gone through a review period and can come to a design review meeting now. @bmzhao do you want to get yourself on the schedule for a review or do you want to extend the comment period?  If you want to extend it, just update the comment on the pull request and possibly followup with an email to developers@tensorflow.org

ematejska:
This is approved, just waiting for a couple of updates from @bmzhao before merging.

DEKHTIARJonathan:
@ematejska @bmzhao I know it's probably a bit late to comment on that PR, really sorry about this, I still have one question if you don't mind :) 

I'm maintaining TF-TRT `tensorflow/compiler/tf2tensorrt` at NVIDIA with @bixia1 from Google. the SavedModel API is critical to our operations (official way we load & export models).

This RFC looks very good to me, I see a lot of interesting new features and possibilities. However,  do you think it would be doable to keep something like `tf.legacy.SavedModel` and keep the old API accessible (for quite some time).

Let me explain the strategy, many TF-TRT users often still train models using Tensorflow 1. Until now we have been able to tell them, you can perform training in TF1.x and inference with TF2.x. It allows them to have a smooth and easy transition from the two versions of tensorflow and still benefit from the improvements in TF2.x (namely on the TF-TRT side). This whole strategy relies on the fact that we are API compatible between TF1 and TF2 for the SavedModel C/C++ API. Breaking this compatibility would have serious implications on many TF-TRT users.
I'm not asking to hold off any change you want to bring, or modify your initial idea to be "API backward compatible". I genuinely think it's fine to break things for the better ;) However if we could keep the old C / C++ API on the side (just the loading part would be good), that would help many users.

Let me know if you want to organize a quick call with @bixia1 and myself. It would be a pleasure to discuss this in more details if needed.

Sorry again for my late remarks. My bad ;) 

CC: @cliffwoolley

bmzhao:
Heya Jonathan!

No worries at all; feel free to add comments here on the PR; since these new APIs will be checked into tensorflow/c/experimental and tensorflow/cc/saved_model/experimental, getting additional feedback on them through this channel would be great. 

I don't think we have any immediate plans for *removing* the existing SavedModel C++ API, especially since we're still fleshing out the details for the new one, and the current one still has significant usage. In addition, since TF2's v1 compat layers still use [tensorflow::Session](https://cs.opensource.google/tensorflow/tensorflow/+/47a28473e4c2701463a0b1841e31a6c5385d77ab:tensorflow/core/public/session.h;l=90), eg: in [tf.compat.v1.session](https://cs.opensource.google/tensorflow/tensorflow/+/47a28473e4c2701463a0b1841e31a6c5385d77ab:tensorflow/python/client/session.py;l=701), I'm not sure if we can completely remove Session-based C++ code in TF2 anyways.

The most I think that can happen is marking the legacy SavedModel C++ API as deprecated once we have stable support for the new SavedModel API, and continue maintaining it for the lifetime of TF2?

CC: @martinwicke @alextp to confirm

Also, it'd be great to sync up with you and bixia@ to learn more about your use existing cases, so that the new SavedModelAPI can support them! Let's set up a meeting :)

martinwicke:
Our API stability guarantees prohibit removing Python APIs other than in a
major version. This includes anything in compat.v1, so all the APIs there
are safe (until TF3, which I'd call beyond the foreseeable future). This
guarantee does not formally extend to the C++ API, but we would not (and
realistically cannot) delete anything that has as widespread use as (say)
tensorflow::Session.

We can, once we have convinced ourselves of the superiority of these new
APIs, deprecate the old ones (ie., mark them as deprecated in docs, and
maybe have them print a warning on first use).

DEKHTIARJonathan:
If you plan at least to keep the possibility to load TF1 SavedModel (not save) that should be good enough for us to work properly @bixia1 @tfeher.

I'll be happy to jump on a call with you @bmzhao and the TF-TRT team (atm. we are roughly 1 Googler and 3 Nvidians). Will help us to plan ahead for the changes and have an idea of the timeline. 

So far from what I can read, it seems like we will be able to adapt our code without changing the Python API (that's a good point) and even support the old and new SavedModel API simultaneously.
I'm perfectly fine if the API goes deprecated, as long as we can keep supporting users who train in TF1 and would like to do inference with TF2 (numerous significant improvements in TF2) it sounds like a more than acceptable solution to me.

@bmzhao Feel free to email me: jdekhtiar@nvidia.com and I'll include all the relevant people (we are not many) in the email thread ;)

PR #210, opened 2020-02-26
Update @angersson to @angerson

angerson:
My username has changed and I didn't notice before that this was out of date.


ewilderj:
just a little bit less swedish

PR #214, opened 2020-03-07
Single-Client Parameter Server Training

yuefengz:
| **RFC #**      | https://github.com/tensorflow/community/pull/214 |
| **Author(s)** | Yuefeng Zhou (yuefengz@google.com), Rick Chao (rchao@google.com) |
| **Sponsor**   | Priya Gupta (priyag@google.com)                      |
| **Updated**   | 2018-03-06                                           |

## Objective
This RFC proposes a design and tentative interfaces for parameter server training in TensorFlow 2 using a single-client architecture.


rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+  for i in range(total_steps):
+    strategy.schedule(replica_fn, args=(distributed_iter,))
+# … omitted
+```
+
+
+**We will start with this kind of training loop in our first version.**
+
+
+### Fault Tolerance
```
terrytangyuan: cc ElasticDL team @wangkuiyi @LiMinghao1994 @skydoorkai @QiJune @brightcoder01 @workingloong @chunyang-wen

2sin18:
Will this approach cause a huge graph that takes a long time to compile, if there're a lot of parameter servers in the training?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+  model = create_model()
+  optimizer = tf.keras.optimizers.Adam()
+  accuracy = tf.keras.metrics.CategoricalAccuracy(name="train_accuracy")
+
+  @tf.function
+  def replica_fn(iterator):
+    x, y = next(iterator)
+    with tf.GradientTape() as tape:
+      predictions = model(x, table, training=True)
+      loss = compute_loss(y, predictions)
```
reedwm: How do you think this should work with mixed precision loss scaling?

We can either have the PS store the loss scale or have each worker have it's own loss scale. If we also want to send gradients to the PS in fp16, this will become more difficult, as then we have to send scaled gradients to the PS, so the workers would also have to send the loss scale over.

byronyi: @reedwm Interesting question there. Just curious, how does current MWMS work with the loss scale?

reedwm: Currently for multi-worker mirrored strategies, each worker has its own independent loss scale. Each worker independently scales the loss, unscales the gradients, and updates its loss scale. MWMS currently does not support allreducing in fp16, since then you must allreduce scaled gradients, but each worker may have a different loss scale. But at least it supports mixed precision for the forwards + backwards pass of the model with loss scaling.

With this proposal, extra work will have to be done even if we don't care about transferring gradients in fp16. We could have each worker have its own loss scale, like in MWMS, but this is more unnatural and difficult since all other variables are on the PS.

byronyi: What if the loss scale is also synchronized, like mean/variance in SyncBatchNormalization? Like you said, it could be aggregated over workers in MWMS or stored on PS.

reedwm: I didn't know SyncBatchNormalization existed until now, but I just read the documentation. I'm not sure how SyncBatchNormalization would work with this proposal either. The [SyncBatchNormalization documentation](https://www.tensorflow.org/api_docs/python/tf/keras/layers/experimental/SyncBatchNormalization) states it synchronizes "the global batch statistics across all devices that are training the model". But you cannot synchronizes with this proposal as it uses asynchronous training. In general, I'm not sure what happens if you call `all_reduce` with a PS Strategy.

I think aggregating the loss scale over workers is tricky since they are asynchronous, but storing it on the PS would work, I think.

byronyi: For common use case in which the batch norm statistics are updated using exponential decay, storing them on PS and updating asynchronously might work without too much harm to the accuracy.

But I agree calling AllReduce in a PS strategy is weird. Ping @anj-s who implemented SyncBN.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+    When join() is being called, it is not allowed to call `schedule`.
+    """
+    pass
+
+  def done(self):
+    """Returns True if there is no pending functions to be executed."""
+    pass
+
+  def local_results(self, future_list):
+    """Get concrete values of the future list.
```
karmel: This is not comprehensible to an outsider. The what of the what?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+    pass
+```
+
+
+#### Custom Training Loop
+
+To construct a custom training loop, users need to
+
+
+*   use `strategy.experimental_distribute_datasets_from_function` to create one dataset per worker. At this point, we recommend against using `strategy.experimental_distribute_dataset`.
```
karmel: If we recommend against, should we specifically prevent, versus soft-recommend and have to deal with the implications?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+*   create models under `strategy.scope` so variables will be assigned to parameter servers.
```
karmel: Is this, more precisely put, create models and all tf.Variables under the scope?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+*   likewise, create a Keras metric object under `strategy.scope`. We expect the metric variables to be stored on parameter servers. Each worker, within their `replica_fn`, updates the metric states.
+*   use `strategy.schedule` to schedule the `replica_fn` on one remote worker. This `replica_fn` should take an iterator and perform forward and backward computation. This `strategy.schedule` will only schedule this `replica_fn` and returns one or several `Future` objects immediately.
```
karmel: What is a Future object? Is there an API contract, and is it a public API?

karmel: "to schedule the `replica_fn` on one remote worker" -- above, the implication is that schedule() schedules on all workers. Whose job is it to iterate over the workers and schedule on each?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+*   likewise, create a Keras metric object under `strategy.scope`. We expect the metric variables to be stored on parameter servers. Each worker, within their `replica_fn`, updates the metric states.
+*   use `strategy.schedule` to schedule the `replica_fn` on one remote worker. This `replica_fn` should take an iterator and perform forward and backward computation. This `strategy.schedule` will only schedule this `replica_fn` and returns one or several `Future` objects immediately.  
+*   use `strategy.local_results` to get concrete values of results returned by `strategy.schedule`. This may be a blocking call if the result is not yet ready. With any failure that cannot be handled will be ignored and as a result some of the results may be `None`.
+*   call `strategy.join` to wait until all scheduled functions are executed.
+
+```python
+# Connect to remote servers with a user-provided `ClusterResolver` object.
+strategy = ParameterServerStrategyV2(cluster_resolver)
+
+dataset_fn = # a function that returns a dataset
```
karmel: Python function presumably? Can you provide an example here? Is this the same as an Estimator's input_fn?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+Another option from calling `join` every epoch, users can choose to schedule all steps and then asynchronously print metric values. This option doesn’t require any synchronization in epoch boundaries.
+
+
+```python
+# … omitted
+with strategy.scope():
+  # … omitted
+  for _ in range(total_steps)):
+    strategy.schedule(step_fn, args=(iterators,))
```
karmel: I don't fully understand the delta here-- what is iterators? Can you omit less?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+    print("Current accuracy: %f" % accuracy.result())
+    time.sleep(60)
+# … omitted
+```
+
+#### Error Reporting From `replica_fn`
+
+Because of the non-blocking `schedule`, any exception raised in `replica_fn` wouldn’t be returned to users immediately. Actually an exception may pollute arbitrary number of functions in flight following the culprit function. We will set the error in returned `Future` objects for the culprit function and these polluted functions and we will raise exceptions when `join` is called.
+
+Therefore the best practice for users is to avoid writing any code that may raise in `replica_fn`:
```
karmel: This seems like a strange requirement. Do you mean something like, "Try to catch and handle as many raised errors within replica_fn as possible to avoid nullifying the entire scheduled workload"? Is there a way to say, "Something went wrong, and we need to rerun this replica_fn, but not all scheduled jobs"?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+*   use repeated dataset so `OutOfRangeError` will be avoided;
+*   avoid using assertion ops or some debugging ops like `tf.debugging.check_numerics`.
```
karmel: And instead do what?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+###### When scheduling
+
+When a worker fails, our training will continue without this failed worker. Functions scheduled on a failed worker will be rescheduled on other workers. 
+
+For functions that bound to a specific worker, e.g. resource creation function, they will be queued until the worker is back.
+
+When the failed worker is back, we will update the cluster configuration with `context.update_server_def` which would also reset all the states. After resources on the restarted worker are built, we can resume scheduling functions on the worker. 
+
+
+###### When materialing a `Future` object
```
karmel: nit: materializing?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+It is possible that a function is executed but its corresponding worker fails when users try to consume its output. In this case, we will give users a `None` value and set an error in the `Future` object.
```
karmel: Can you show an example of what this would look like, and how a user should write code robust to this?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+We can mitigate the problem by eagerly materializing function outputs when they are passed to `local_results`. 
+
+We can explore mechanisms to recover these objects in the future. In the short-term, users can choose to write the results to variables on parameter servers, just like a Keras metric.
+
+
+##### Parameter server failure
+
+When a parameter server fails, the error will be propagated to the client via workers. Since the latest values of variables on the failed parameter servers are gone, there is no way for the client to recover them. Therefore the training will pause until the failed parameter server is back. The client then needs to clean up other variables on other parameter servers, rebuild all the variables and load variable values from a checkpoint. To trigger this process, the simplest method is to restart the client as well.
```
karmel: Can you give a more concrete example of how training/the user will recover here? I have two parameter servers with ten workers each. What happens to 22 (or 23? Is the single-client one of the workers?) nodes if one of the param servers goes down?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+#### Resource Management for Workers
+
+When a worker has recovered from failure, we will need to rebuild iterators, worker-local variables, lookup tables and other resources on that worker that don’t need to be read from a checkpoint. This means that the client will have to keep track of these iterators, worker-local variables and other resources.
+
+Keeping track of resources and rebuilding them will be achieved depending how users create their resources:
+
+
+*   we will record iterators created via `tf.distribute`’s API; The state of a rebuilt iterator will be lost. We can recover their states as future work.
+*   we will capture the creation of worker-local variables via variable creator scopes.
```
karmel: Who owns the var creator scopes? Where is the state and memory held in the event of worker failure?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+*   eagerly fetch the outputs of remote functions once the outputs are passed to `strategy.local_result`. In this way, we can know the status of function execution earlier.
+*   recommend users schedule only small functions. Large functions are more expensive to retry.
+
+
+### Evaluation
+
+Historically, `tf.estimator.Estimator` uses a dedicated evaluator that periodically loads from a checkpoint, and performs evaluation with evaluation data. However `tf.keras` typically evaluates in an alternating manner after every epoch of training, and this is also the case with `tf.keras` + `MultiWorkerMirroredStrategy`. 
+
+With `ParameterServerStrategyV2`, we will start with a dedicated** evaluator that runs alongside the training cluster**, **aka “sidecar evaluation”**; in this scheme, training client is required to generate checkpoints periodically, and the evaluator reads the latest checkpoint as it becomes available. The evaluation is asynchronous to the training progress. With this we provide the functionality Estimator has been able to with Keras API, which is important to attract updates from Estimator users to TF 2.0.
```
karmel: nit: the bolding is mis-spaced in the first line here.

karmel: In estimator, we auto-generated checkpoints, IIRC. In this case, is it the user's responsibility? What if they forget? How can we make this easier/part of the examples?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+With our recommendation, users should create a separate evaluation client that runs the same python binary as the training client. This python binary will contain the if-else clause as it bifurcates into two paths:
+
+
+```python
+if cluster_resolver.task_type == "chief":
+  run_training_loop()
+elif cluster_resolver.task_type == "evaluator":
+  run_evaluation_loop()
+```
```
karmel: It seems sort of sad to have this top level forking. Is there a better way?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+  while True:
+    latest_checkpoint = get_new_checkpoint()
+    checkpoint.restore(latest_checkpoint)
+    eval_result = eval_fn(iterator)  # Users can print, early stop, mark ckpt..
+```
+
+
+In the evaluation client, the user loads the checkpoints that were periodically saved into the model (by the training client), does evaluation over a full pass of eval dataset, and does whatever they want to do with eval results. Examples include exporting them to files which can be read by the training client for actions (such as reducing learning rate, early stopping, etc.)
+
```
karmel: Can we see the e2e code example with training_loop and the fork as well?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+    pass
+  
+  def _fetch(self):
+    pass
+```
+
+
+
+We can potentially merge this `Future` class with our `Tensor` class.
+
```
karmel: Note that if this is returned by a public API, we generally require that the returned type is public, or some interface to it. I don't love the idea of a generically named "Future" type in TF, especially because it is suspiciously like the two-stage-Session-running of v1, so we should consider what the contract is here and what the user needs to know about this.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+*   we will record iterators created via `tf.distribute`’s API; The state of a rebuilt iterator will be lost. We can recover their states as future work.
+*   we will capture the creation of worker-local variables via variable creator scopes.
+*   in the future we will provide users an API to create worker-local resources. We will capture these resources in the API.
+
+If users create iterators or other resources inside a function but don’t expose them as outputs, we will not rebuild them.
+
+
+#### The Unknown of Scheduled Functions
+
+For functions that have been scheduled, it is difficult for the client to know whether they have actually been executed or not when the client detects their corresponding worker failure. Therefore, in addition to inform users of this uncertainty in the case of worker failure, we should do the following to reduce this uncertainty:
```
tomerk: Depending on the consistency guarantees we make for the parameter server:
Would it be possible (or even make sense) for the parameter server to track which functions have and haven't been completed?

This may let us do things like know exactly which functions need to be 
re-executed if a parameter server fails and a checkpoint needs to be reloaded.

Then again, if people are training asynchronously I guess it's not necessarily worth trying to provide reproducibility guarantees, or guarantees that each step will be run exactly once.

rfcs/20200306-single-client-parameter-server.md
tomerk: So just to clarify: The parameter server won't be shared by the training & evaluation job. Instead, they'll do what estimator does and communicate via checkpoints & the file system?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+*   recommend users schedule only small functions. Large functions are more expensive to retry.
+
+
+### Evaluation
+
+Historically, `tf.estimator.Estimator` uses a dedicated evaluator that periodically loads from a checkpoint, and performs evaluation with evaluation data. However `tf.keras` typically evaluates in an alternating manner after every epoch of training, and this is also the case with `tf.keras` + `MultiWorkerMirroredStrategy`. 
+
+With `ParameterServerStrategyV2`, we will start with a dedicated** evaluator that runs alongside the training cluster**, **aka “sidecar evaluation”**; in this scheme, training client is required to generate checkpoints periodically, and the evaluator reads the latest checkpoint as it becomes available. The evaluation is asynchronous to the training progress. With this we provide the functionality Estimator has been able to with Keras API, which is important to attract updates from Estimator users to TF 2.0.
+
+With our recommendation, users should create a separate evaluation client that runs the same python binary as the training client. This python binary will contain the if-else clause as it bifurcates into two paths:
```
tomerk: Given it's two separate jobs, why require it be the same python binary instead of having two separate python binaries? Internally we have pretty good tooling for bringing up one binary as a training job and another as an evaluation job.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+  eval_dataset, model, eval_accuracy = ...
+  checkpoint = tf.train.Checkpoint(model=model)
+
+  @tf.function
+  def eval_fn(eval_dataset):
+    for _ in range(eval_steps):
+      # evaluation pass
+    return eval_accuracy.result()
+
+  while True:
```
tomerk: There's definitely room for non-parameter-server-specific utilities here for side car evaluation. Plenty of users (e.g. object detection) want to use dedicated evaluation jobs with mirroredstrategy, tpustrategy, & others.

tomerk: The common patterns are:
Wait until a checkpoint is present
Evaluate
Check if a new one is present, if not wait a while until it is
Finish evaluating when no new checkpoint has been written for a while

With timeouts, meaningful error messages, etc.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+    latest_checkpoint = get_new_checkpoint()
+    checkpoint.restore(latest_checkpoint)
+    eval_result = eval_fn(iterator)  # Users can print, early stop, mark ckpt..
+```
+
+
+In the evaluation client, the user loads the checkpoints that were periodically saved into the model (by the training client), does evaluation over a full pass of eval dataset, and does whatever they want to do with eval results. Examples include exporting them to files which can be read by the training client for actions (such as reducing learning rate, early stopping, etc.)
```
tomerk: So the training job and sidecar evaluation have *two-way* communication, all via the file system? Does this make sense?

It might since the training client can run asynchronously then somehow signal to workers, but it feels a little unwieldy...

How hard would it be to allow multiple clients/jobs connect to the same parameter server? Then they could communicate via reads/writes directly to the parameter server.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+
+
+*   It allows `schedule` method return immediately after pushing functions to its scheduling queue. It allows these methods to return without needing to wait for acknowledgement from workers.
+*   It serves as the container for values or errors. It would be binded with a value or an error later. When it is rebuilt, we can replace its underlying value silently.
+*   When being passed to `local_result`, we flag it to indicate that this value needs to be fetched eagerly.
+*   (Future work) It captures the lineage between functions and return values so that we can rebuild any poisoned objects. 
+
+```python
+class Future(object):
```
tomerk: So no `on_complete` asynchronous callbacks on futures? The clients check the value of futures exclusively by joining/polling?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+*   call `strategy.join` to wait until all scheduled functions are executed.
+
+```python
+# Connect to remote servers with a user-provided `ClusterResolver` object.
+strategy = ParameterServerStrategyV2(cluster_resolver)
+
+dataset_fn = # a function that returns a dataset
+
+# Clone the dataset on all workers, shuffled with different seeds.
+distributed_dataset = strategy.experimental_distribute_datasets_from_function(
```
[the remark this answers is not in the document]
yuefengz: If it is a tf dataset, it is supposed to work. But we need to verify.

yuefengz:
> Will this approach cause a huge graph that takes a long time to compile, if there're a lot of parameter servers in the training?

It will trace a function for one worker and schedule the function to all workers. So the graph size should be small as opposed to in-graph distributed training where we have a single large graph that has definition for all workers.

If there are a lot of parameter servers, the function will be partitioned into smaller functions, one for each ps. We don't see any problem so far. Please let us know if you have encountered any problem before.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+
+If we do the same thing with the `strategy.schedule` API, there are several challenges.
+
+The first challenge is we don’t know which worker the `get_next` should return to since where the `replica_fn` will be executed will be decided later. Some later-binding mechanism can be explored.
+
+The second challenge is calling `get_next` on an iterator is synchronous. This means that the training loop is not truly asynchronous. It is tricky to make `get_next` asynchronous because the client doesn’t know how many items will be in the iterator and thus doesn’t know how many functions to schedule. 
+
+
+##### Alternative: passing iterators to `strategy.schedule`
```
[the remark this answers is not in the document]
yuefengz: Yes, this is what happens in Estimator. If it is really a concern, users can choose to join before saving the model.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+### Programming Model
+
+With a single-client architecture, the programming model will be different than the multi-client architecture. All workers and parameter servers are standard TensorFlow servers, and the user-defined program will run on the client only. Generally, no matter what high-level APIs users will use, the workflow for running a step function distributedly with single-client approach includes the following steps:
+
+
+1. Connect to all remote workers and parameter servers. 
+2. Create variables on parameter servers and hold references to them.
+3. Create datasets and iterators on workers.
+4. Create the replica function that takes an iterator as input, trace it and register it on all workers.  Note: a function may create variables as well. If not specified, they will be created on parameter servers as well.
```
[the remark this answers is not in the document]
yuefengz: If variables defined inside a tf.function, these variables will be created when the function is traced. The tracing happens on the client, so the client will create these variables under the device scope of parameter servers.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+### Interfaces
+
+One of our goals is to make `ParameterServerStrategy`’s API consistent with other strategies so that users can easily switch between them. This may be challenging due to the fundamental difference between synchronous and asynchronous training. Therefore, we try to use most of the Distribution Strategy’ APIs but occasionally we will create APIs that still make sense in other strategies as well.
+
+**Note: all proposed APIs in this document are tentative and only valid for our first version. We will revisit them once we get enough feedback from users and the community.**
+
+
+#### Constraints
+
+Function is first-class citizen. Users should only schedule functions instead of running individual ops, in addition to creating variables.
```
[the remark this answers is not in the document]
yuefengz: Yes, tf.functions. We cannot support arbitrary python functions now. I'll update it.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+Users can occasionally run individual ops on the client, only for reporting purposes such as printing a metric’s value.
```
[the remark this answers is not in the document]
yuefengz: It is common for users to call metrics.result(), fetching the variable of `sum` and `count`, dividing the `sum` by the `count` happens on the client. Users can just do it without calling into `schedule`. We won't have special conditioning in our implementation.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+*   expose the non-blocking semantics to users.
+
+To enable scheduling a function on any worker, we recommend users create the same dataset, but may shuffle differently, on all workers via `strategy.experimental_distribute_datasets_from_function` API.
+
+
+```python
+class ParameterServerStrategyV2:
+  
+  def schedule(self, replica_fn, args=(), kwargs=()):
+    """Schedule the `replica_fn` on all replicas in a sync group (a worker).
```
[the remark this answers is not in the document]
yuefengz: "Sync group" is a concept of `tf.distribute` and some of our APIs depend on this concept. (You can search for sync group in the code base.) It has been defined that a sync group is one worker in PSStrategy and all workers in MWMS. So I followed this concept to be consistent with rest of the `tf.distribute` API.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+    Schedule the `replica_fn` on all replicas in a sync group that is available,    
+    returns a future of PerReplica immediately if `function` has return values.
+
+    It implements at-least-once semantics for function execution. If a worker
+    fails, it will try to reschedule the function on another replica group or throw     
+    an exception to users. So this method assumes that function execution can be 
+    out of order and function inputs are shared between sync groups. 
+
+    We don't support the cases where `args` or `kwargs` are bound to a specific
+    sync group. We will consider supporting them in the future.
```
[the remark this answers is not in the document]
yuefengz: If we schedule two functions, one function takes the input of another function, then they should schedule on the same worker. The return value of the first function (i.e. the input of the second function) is bound to that worker. Does that make sense to you?

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+    If there are barriers in `replica_fn`, it is users' responsibility to make
+    sure they won't cause deadlock.
```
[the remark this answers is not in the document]
yuefengz: For one example: users add a collective op in the function.

rfcs/20200306-single-client-parameter-server.md
yuefengz: Yes, we will throw an exception in this method. Is there anything we should do?

rfcs/20200306-single-client-parameter-server.md
yuefengz: `schedule` should only schedule to one available worker that has the least load.

rfcs/20200306-single-client-parameter-server.md
yuefengz: Users can definitely use the client to drive the evaluation on the `evaluator` (or any worker) with our current design. This is one way of running evaluation if users prefer the Estimator-style side-car eval. I think we need more paragraphs for evaluation. @rchao

rfcs/20200306-single-client-parameter-server.md
yuefengz: If users want the training cluster to read eval results, they probably shouldn't do side-car. They should do inline eval -- alternative between training and eval, preferably using the same worker pool. Our current design allows that with some limitation.

It is a future work to fully support inline eval since it needs us to think carefully on how to support visitation guarantee with a sharded dataset, especially when workers go down or when there are stragglers.

rfcs/20200306-single-client-parameter-server.md
yuefengz: I believe users can do so. @rchao

rfcs/20200306-single-client-parameter-server.md
yuefengz: This is the current design for our first cut.

rfcs/20200306-single-client-parameter-server.md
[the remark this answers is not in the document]
rchao: I think the API should follow base class' contract to the extend possible: 1) It needs to return a `tf.data.Dataset`, and 2) the Dataset should have a per-replica batch size. Like yuefengz@ said verification should be done.

rfcs/20200306-single-client-parameter-server.md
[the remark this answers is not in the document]
rchao: To the second question, since only the client is responsible for creating variables, they don't need to be synced. This is one of the benefits of single client design.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+    """
+    pass
+  
+  def join(self):
+    """Wait until all scheduled functions are finished.
+
+    Raises an error if any of the functions fails to execute. In this case,
+    there is no guarantee that non-failing functions will complete.
```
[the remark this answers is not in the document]
rchao: Based on recent discussion, we plan to revise this so that if both 1) no functions are left in the queue to be scheduled, 2) no function is being executed on remote workers, are satisfied, join() returns. Before that happens, it blocks (for now without a timeout). 

Also, we may design it in a way that if a worker fails to execute the function, the function goes back to the queue for other workers to execute. With this, there would not be the complication of user needing to handle an error (since the user may not know what to do, other than retrying, anyway).

rfcs/20200306-single-client-parameter-server.md
rchao: Yes. I think this statement is also indicating one effect of creating model is variable creation.

rfcs/20200306-single-client-parameter-server.md
rchao: A Future object possesses a field that may (once function returns result) or may not (when the function is only scheduled but not yet executed) have a concrete value. It's supposed to be used in conjunction with `local_result` with which the value of a Future is fetched.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+```python
+# … omitted
+with strategy.scope():
+  # … omitted
+  for _ in range(total_steps)):
+    strategy.schedule(step_fn, args=(iterators,))
+
+  # Print accuracy value every one minute.
+  while not strategy.done():
+    print("Current accuracy: %f" % accuracy.result())
```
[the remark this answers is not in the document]
rchao: My understanding is by design, there's no consistency guaranteed by PS training. Values at any moment can be read from PS. Can you give an example of how there can be an invalid state? My view of this is, someone might be reading while some others may be writing, and it doesn't guarantee the value read is before writing or after writing.

rfcs/20200306-single-client-parameter-server.md
rchao: I think what this meant is more like "If you raise an error within `replica_fn`, you won't be able to handle it by try-except `schedule()`, so the best thing to do is not to raise an error". I understand this is somewhat limiting, and I think this can be solved by having a mechanism of transporting error message from worker threads to client's main thread.

rfcs/20200306-single-client-parameter-server.md
rchao: I think the best solution is being able to transport error message/stack trace from worker threads to main thread (or an error handling thread) to report to user and stop the program.

rfcs/20200306-single-client-parameter-server.md
rchao: I believe currently all ps's and workers (and client) are considered the same cluster, so if any of the ps's fails, the cluster restarts (yuefengz@ to correct me if wrong).

rfcs/20200306-single-client-parameter-server.md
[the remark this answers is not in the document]
yuefengz: For the failure question, the client only distinguishes between retryable and non-retryable errors. For retryable errors such as worker preemption, the client will try to find a new worker to run. For non-retryable errors, the client will report it to users upon `join()`ing or upon materializing the function return value.

rfcs/20200306-single-client-parameter-server.md
```diff
@@ -0,0 +1,517 @@
...
+
+    Raises an error if any of the functions fails to execute. In this case,
+    there is no guarantee that non-failing functions will complete.
+
+    When join() is being called, it is not allowed to call `schedule`.
+    """
+    pass
+
+  def done(self):
+    """Returns True if there is no pending functions to be executed."""
```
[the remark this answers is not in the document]
yuefengz: Executed means a function has been executed and the client got the acknowledge. But materializing the function result is independent.

rfcs/20200306-single-client-parameter-server.md
yuefengz: Anything returned from `schedule` can be passed to this method to get the concrete value. The `schedule` just return an opaque object. This is similar to other strategies where users pass the result of `run` to `local_result` to get actual values.

Maybe I don't get your question?

rfcs/20200306-single-client-parameter-server.md
yuefengz: An example can be found [here in the tf.distribute's doc](https://www.tensorflow.org/api_docs/python/tf/distribute/Strategy#experimental_distribute_datasets_from_function).

rfcs/20200306-single-client-parameter-server.md
yuefengz: Sorry the name is confusing, I renamed (in my pending commit) it to `distributed_iter`.

rfcs/20200306-single-client-parameter-server.md
yuefengz: In addition to the pitfall of eager async execution that Rick mentioned, there is another gotcha in our eager async implementation: a failing function will pollute all other function scheduled on that executor. For some errors such as NaN errors, we shouldn't do that. We will improve that together with runtime folks. 

cc @haoyuz

rfcs/20200306-single-client-parameter-server.md
yuefengz: Users can do the assertion themselves in the function and return something indicating a failure, instead of throwing.

rfcs/20200306-single-client-parameter-server.md
yuefengz: Done (in my pending commit).

rfcs/20200306-single-client-parameter-server.md
yuefengz: Update the example in the `Custom Training Loop` section. You can see its last line.

rfcs/20200306-single-client-parameter-server.md
yuefengz: PS failure: client restart and wait until the failing ps is back.

rfcs/20200306-single-client-parameter-server.md
yuefengz: All the variable creation will be captured by distribution strategies' scope. Worker-local variables will be created on workers, the client just holds references to these variables, similar to an iterator.

rfcs/20200306-single-client-parameter-server.md
yuefengz: You can compare the future to `PerReplica` value in `MirroredStrategy`. In `MirroredStrategy`, `run` returns an opaque type which is `PerReplica` and users can only get its concrete values through `local_results`. Similarly, our `schedule` method returns `Future` (or `PerReplica` of `Future`s when we support multi-GPU), users can only get its concrete values through `local_results`

Our `Tensor` object in async eager is a future-like object as well. It doesn't have any concrete value unless we materialize it via `numpy`. Therefore, we can think of merging the logic to `Tensor` class in the future.

rfcs/20200306-single-client-parameter-server.md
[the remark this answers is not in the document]
yuefengz: In addition to what Rick said, if strong consistency is really required, users can return something from the function and read that return values. E.g. some computed metrics that doesn't depend on multiple variables on ps.

rfcs/20200306-single-client-parameter-server.md
yuefengz: We can definitely asynchronous callbacks if this is needed. Maybe you can give me some use cases?

rfcs/20200306-single-client-parameter-server.md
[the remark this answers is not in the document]
yuefengz: Do you think it is possible for Keras team to provide non-variable-based metrics? This may be useful who evaluation and for users who want strong consistency guarantee.

rfcs/20200306-single-client-parameter-server.md
rchao: Yes, if we use an async executor for evaluator's function scheduling (with a checkpoint file), it should be feasible that we have a true single client (aka inline evaluation). One implication is that the evaluator would need to be considered the same cluster as the chief and workers. The plan now is that we develop an API for sidecar eval, and provide example for how inline eval can be done (possibly some util if useful), and which to be used is at their discretion.

rfcs/20200306-single-client-parameter-server.md
rchao: Yes - we have some code sample and will update in the doc.

rfcs/20200306-single-client-parameter-server.md
rchao: With this, it sounds to me the contract is "one would need to use `local_results` to deal with whatever they get from `run()` or `schedule()`", and doesn't mention the type of it. 

Better yet, if possible, we encourage user not to rely on the return value of `schedule()`. For many things that are involved in training such as loss or metrics, they can simply be variables that get updated within the remote function call. User simply reads those values from the client.

rfcs/20200306-single-client-parameter-server.md
rchao: Our current design allows users to receive an object from `strategy.schedule()`, and when calling `local_results(object)` one gets the result. If it's not executed or not fully executed due to worker preemption, the strategy will find another worker to take over. For other kinds of errors, it is possible to design it in a way that such error would be discovered upon calling `strategy.local_results()` or `strategy.join()`. For the case where PS fails, the whole cluster would be restarted and a checkpoint previously saved is loaded.

rfcs/20200306-single-client-parameter-server.md
rchao: There's this flexibility to do either way. We were assuming the convenience brought by letting every job run the same binary, regardless of chief, worker, ps, or evaluator. But yes, user is free to have two or even more binaries for different jobs.

rfcs/20200306-single-client-parameter-server.md
rchao: Exactly and agreed. We're working on this and will update the design doc accordingly. Thanks for pointing this out.

rfcs/20200306-single-client-parameter-server.md
rchao: One option yuefengz@ and I discussed is inline-evaluation-by-single-evaluator, where one schedules an evaluation function to the dedicated evaluator job. That way, a variable (on ps) can be used for the evaluator to update, and for the client to read and act accordingly.

rfcs/20200306-single-client-parameter-server.md
rchao: For custom training loop, yes, it is their responsibility (and we should update the example accordingly). One thing we're working on is an `EvaluationLoop` API with which they gain automatic sidecar eval. If they use this, they are forced to provide a `checkpoint_dir` to `__init__`, matching what they use in the client when saving.

In the future, once we integrate with Keras `model.fit()`, users should use a callback as they would without distributed training.

byronyi: Given the large sizes of some embedding models, incremental checkpoints might ease the file system pressure if users would like to do frequent evaluation without sharing the parameter servers.

byronyi:
What's the protocol for PS/worker if the client exits (on error or not)? Do they still need manual cleanup like PS in 1.x?

rfcs/20200306-single-client-parameter-server.md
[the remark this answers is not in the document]
rchao: @yuefengz How about we remove this paragraph to avoid confusion, since we don't support it anyway? We can always add it back when it's supported.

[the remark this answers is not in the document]
yuefengz: SG, will you remove it?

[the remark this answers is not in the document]
rchao: Sg, I'll do that

ematejska:
This went through a design review on 4/9/2020 and the RFC has been updated based on the review comments. Accepting.

PR #240, opened 2020-04-30
Add C++ style guide.

aaudiber:
This guide is a starting point for documenting Tensorflow C++ style, especially where it differs from general Google C++ style.

The goal of this PR is to reflect style rules *as they are today*, so that we can be on the same page about what the style rules are, and so that we have a basis from which to discuss potential changes to the style rules. This PR itself is not the right place to discuss changing the style rules (though if the content here does not reflect the current reality, by all means please comment!).

### DIFF
--- a/governance/cpp-style.md
+++ b/governance/cpp-style.md
@@ -1,0 +1,64 @@
+# C++ Coding Style
+
+Tensorflow follows [Google C++ style](https://google.github.io/styleguide/cppguide.html),
+with a few additions.
+
+## Status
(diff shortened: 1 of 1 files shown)


aaudiber:
Would it help to remove the explanations of reasoning and simply state "Tensorflow does not use StatusOr"?

aaudiber:
@martinwicke PTAL

alextp:
@martinwicke let's merge this PR

Issue #227, opened 2020-04-14
Create a Github team for SIG mantainers


bhack:
It could be helpful to quickly involve official SIG maintainers in general purpose SIGs threads/decisions.

https://help.github.com/en/github/setting-up-and-managing-organizations-and-teams/about-teams

terrytangyuan:
Another suggestion is that SIG maintainers should be able to merge PRs that modify the folders that we maintain. This would avoid situation like in https://github.com/tensorflow/community/pull/211 where the PR has not been merged yet (after 2 weeks at the time of writing).

bhack:
Mhh. It Is not explicit but I don't think they could be classified as maintainer in this repo: https://github.com/tensorflow/community/blob/master/governance/code-and-collaboration.md#maintainers

terrytangyuan:
Perhaps we could set up something similar to what's done in Kubernetes-related projects (e.g. Kubeflow) which use [prow](https://prow.k8s.io/command-help) and `OWNERS` file so that changes to sub-folders can be approved/merged by users specified in the `OWNERS` file in the corresponding sub-folder.

bhack:
What I meant Is that our ownership governance was about review not merge. Ownership and write power is not the same role in our governance:
https://github.com/tensorflow/community/blob/master/CODEOWNERS

terrytangyuan:
Oh I see what you mean now. Anyways, the more the community (especially SIG maintainers) can get involved and help move things forward quickly, the better.

bhack:
@terrytangyuan My opinon Is that with the right level of trust could be ok. But to achieve this we need to explict this interpretation somewhere in the governance folder and that the community team agree about this policy.

Issue #226, opened 2020-04-14
Community repository minimum maintenance


bhack:
Can we expect to have a regular maintenance of this low traffic community repository? 

* Many RFCs go regularly out of the comment deadline and we don't know the public status of the process for months.
* Many very trivial fixes on docs are not merged or if approved we need to wait for months.
* Many sparse topics issues like Gsoc/Code in are not closed or redirected to the right place. 

I think that a regular 10 minutes check one or two times at weeks it is enough triage and maintain this repo in an healty and useful status as is the many communication channel between the internal team and the Tensorflow community.

bhack:
To improve the public communication It could be nice if the implementation start after the RFC it is public approved (merged) https://github.com/tensorflow/community/pull/182#issuecomment-613573220

mihaimaruseac:
Agree, but note that that commit you mention landed **after** the RFC approval, not before.

ematejska:
The problem here was with the process (I run the RFC process nowadays and take full responsibility.) The design was was reviewed but the communication was not good afterwards.  Thank you for this feedback.

bhack:
@mihaimaruseac I don't see this in our governance https://github.com/tensorflow/community/blob/master/governance/rfc-admin-guide.md#rfc-acceptance-process.
You can always open a PR to change our governance.

bhack:
@ematejska I am glad that you are the owner of the RFC. I hope that we could improve the labeling governance after the comment public deadline.

bhack:
@mihaimaruseac Don't take it as personal by downvoting. You are internal and sometimes you cannot understand that we are totally unaware of what will happen with respect to your internal threads. As a community we only know what is public.

bhack:
Can you give a pass to noise tickets?

bhack:
@ematejska I think that we could evaluate to close almost all the open issues in the repo.

ematejska:
@bhack @theadactyl and I have started to go through these on a periodic cadence.  We should get to these issues early next week at the latest.

bhack:
Thanks

theadactyl:
Got to the issues! We are working steadily on the RFC backlog as well, so closing this issue.

PR #241, opened 2020-05-05
SIG Addons: Procedure and RFC template for Addons migrations

### LINKED ISSUE #239: Tensorflow addons promotion and TF
Please take an overview of https://github.com/tensorflow/tensorflow/pull/33945#issuecomment-621117201

Can we document somewhere that new ops PR on Tensorflow duplicating TF addons implementations can pass through a small RFC proposal here (so that we include Addons removal or migration in the proposal)?

Currently we have https://github.com/tensorflow/addons/pull/774 that was approved directly inside the SIG but here we don't have any process documented on the TF Team side.

That PR was derived from a ticket discussed with @karmel @ewilderj  at https://github.com/tensorflow/addons/issues/686

seanpmorgan:
Tried to incorporate the discussion points in #239 into documentation. Closes #239 once agreed upon.

The goal of this PR is to standardize how Addons components will be migrated going forward. After the RFC template and and Procedure are agreed upon we will submit the proposal for GELU migration.

cc @karmel @alextp @bhack  for feedback

### DIFF
--- a/sigs/addons/MIGRATION_RFC_TEMPLATE.md
+++ b/sigs/addons/MIGRATION_RFC_TEMPLATE.md
@@ -24,7 +24,7 @@
 
 ## Changes to Implementation (If Needed)
 ```
-Code snippet showing proposed code migration
+Code snippet(s) showing proposed code changes
 ```
 * Discussion on the rationale for changes
 
--- a/sigs/addons/MIGRATION_RFC_TEMPLATE.md
+++ b/sigs/addons/MIGRATION_RFC_TEMPLATE.md
@@ -30,8 +30,8 @@ Code snippet(s) showing proposed code changes
 
 ## Transition Plan
 * Proposed landing place in tf-core
+* Will there be any changes to the API? (e.g. additional parameters or renaming)
 * Deprecation plan for Addons
-    * Will we be able to alias the core implementation? (parameters must be exact match)
 
 ## Relevant GitHub Issues
 
--- a/sigs/addons/MIGRATION_RFC_TEMPLATE.md
+++ b/sigs/addons/MIGRATION_RFC_TEMPLATE.md
@@ -1,0 +1,39 @@
+# Migrate XXXXX from TensorFlow Addons to TensorFlow Core
+
+| Status      | Proposed (Waiting for Sponsor)                                                                                           |
+| :---------- | :------------------------------------------------------------------------------------------------- |
+| **RFC #**   | TBD after PR |                                       |
+| **Authors** | XXXXXXXXXX   |
+| **Sponsor** | XXXXXXXXXX   |
+| **Updated** | YYYY-MM-DD   |
+| **Sponsorship Deadline** | YYYY-MM-DD   |
+
+
+## Background
+* Link to implementation in Addons:
+* What are the use cases for the addon?
+* Have there been signifiant issues reported to Addons that need to be adressed?
+* When was it implemented in Addons?
+* What is the pytest coverage of the addon?
+
+## Metrics Supporting the Migration
+* OSS usage, H5 Index, etc.
+
+## Transition Plan
+* Proposed landing place in tf-core
+* Deprecation plan for Addons
+    * Will we be able to alias the core implementation? (parameters must be exact match)
+
+## Changes to Implementation (If Needed)
+```
+Code snippet showing proposed code migration
+```
+* Discussion on the rationale for changes
+
+## Relevant GitHub Issues
+
+## Questions and Discussion Topics
+
+
+## Final Decision
+TBD
--- a/sigs/addons/MIGRATION_RFC_TEMPLATE.md
+++ b/sigs/addons/MIGRATION_RFC_TEMPLATE.md
@@ -6,23 +6,21 @@
 | **Authors** | XXXXXXXXXX   |
 | **Sponsor** | XXXXXXXXXX   |
 | **Updated** | YYYY-MM-DD   |
-| **Sponsorship Deadline** | YYYY-MM-DD   |
+| **Sponsorship Deadline** | YYYY-MM-DD (45 Days after submission)  |
 
-
-## Background
-* Link to implementation in Addons:
+## Rationale for Migration
 * What are the use cases for the addon?
-* Have there been signifiant issues reported to Addons that need to be adressed?
-* When was it implemented in Addons?
-* What is the pytest coverage of the addon?
-
-## Metrics Supporting the Migration
 * OSS usage, H5 Index, etc.
 
-## Transition Plan
-* Proposed landing place in tf-core
-* Deprecation plan for Addons
-    * Will we be able to alias the core implementation? (parameters must be exact match)
+## Historical Information
+* Have there been signifiant issues reported to Addons that need to be adressed?
+* When was it implemented in Addons?
+
+## Implementation Details
+* Link to implementation in Addons:
+* Does this include custom-op kernels?
+    * Are they CPU/GPU/TPU compatible?
+* What is the pytest coverage of the addon?
 
 ## Changes to Implementation (If Needed)
 ```
@@ -30,10 +28,14 @@ Code snippet showing proposed code migration
 ```
 * Discussion on the rationale for changes
 
+## Transition Plan
+* Proposed landing place in tf-core
+* Deprecation plan for Addons
+    * Will we be able to alias the core implementation? (parameters must be exact match)
+
 ## Relevant GitHub Issues
 
 ## Questions and Discussion Topics
-
 
 ## Final Decision
 TBD
--- a/sigs/addons/MIGRATION_TO_CORE.md
+++ b/sigs/addons/MIGRATION_TO_CORE.md
@@ -1,0 +1,32 @@
+# Migration From TF-Addons To TensorFlow Core
+
+### In-Progress & Previous Migrations:
+https://github.com/tensorflow/addons/projects/2/
+
+### Process 
+1. Create an issue in TensorFlow Addons for a candidate that you think should be 
+migrated. 
+2. The SIG will evaluate the request and add it to the `Potential Candidates` section 
+of our GitHub project.
+3. If it's agreed that a migration makes sense, an RFC needs to be written to discuss 
+the move with a larger community audience. 
+    * If the transition will impact tf-core and Keras then submit the RFC to 
+    [TensorFlow Community](https://github.com/tensorflow/community)
+    * Additions which only subclass Keras APIs should submit their migration proposals to 
+    [Keras Governance](https://github.com/keras-team/governance)
+    
+4. A sponsor from the TF/Keras team must agree to shepard the transition.
+   * If no sponsor is obtained after 45 days the RFC will be rejected and will remain 
+   as part of Addons.
+5. If a sponsor is obtained, and the RFC is approved, a pull request must move the 
+addon along with proper tests.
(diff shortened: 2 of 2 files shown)


sigs/addons/MIGRATION_TO_CORE.md
```diff
@@ -0,0 +1,32 @@
+# Migration From TF-Addons To TensorFlow Core
+
+### In-Progress & Previous Migrations:
+https://github.com/tensorflow/addons/projects/2/
```
seanpmorgan, line 4: @theadactyl This may be more visible to a larger audience if the project is housed in tf/community

sigs/addons/MIGRATION_RFC_TEMPLATE.md
```diff
@@ -0,0 +1,41 @@
...
+## Changes to Implementation (If Needed)
+```
+Code snippet(s) showing proposed code changes
+```
+* Discussion on the rationale for changes
+
+## Transition Plan
+* Proposed landing place in tf-core
+* Deprecation plan for Addons
+    * Will we be able to alias the core implementation? (parameters must be exact match)
```
bhack, line 34: I think that the upstreamed version could, under a review request, acquire a parameter superset. But if the additional paramas are optional I think that it is still ok if we have the same output with the parameter subset.

sigs/addons/MIGRATION_RFC_TEMPLATE.md
```diff
@@ -0,0 +1,41 @@
...
+| **Authors** | XXXXXXXXXX   |
+| **Sponsor** | XXXXXXXXXX   |
+| **Updated** | YYYY-MM-DD   |
+| **Sponsorship Deadline** | YYYY-MM-DD (45 Days after submission)  |
+
+## Rationale for Migration
+* What are the use cases for the addon?
+* OSS usage, H5 Index, etc.
+
+## Historical Information
```
bhack, line 15: I don't know how it is easy to have a quick query but for stability we could ask how many bugs was handled in the last (x) month(s).

sigs/addons/MIGRATION_TO_CORE.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Process
```
bhack, line 6: I think this could be quite complicated for cases like `ImageProjectiveTransformV2`.
As we could have an upstream request where the sponsor is already in the TF core team I don't think this double pass is plausible for this specific case.

sigs/addons/MIGRATION_TO_CORE.md
```diff
@@ -0,0 +1,32 @@
...
+of our GitHub project.
+3. If it's agreed that a migration makes sense, an RFC needs to be written to discuss 
+the move with a larger community audience. 
+    * If the transition will impact tf-core and Keras then submit the RFC to 
+    [TensorFlow Community](https://github.com/tensorflow/community)
+    * Additions which only subclass Keras APIs should submit their migration proposals to 
+    [Keras Governance](https://github.com/keras-team/governance)
+    
+4. A sponsor from the TF/Keras team must agree to shepard the transition.
+   * If no sponsor is obtained after 45 days the RFC will be rejected and will remain
```
bhack, line 19: I am ok to put a number here but I think that we need a realistic evaluation by the TF team. Is 45 days something realistic for the internal process?

sigs/addons/MIGRATION_TO_CORE.md
```diff
@@ -0,0 +1,32 @@
...
+   * If no sponsor is obtained after 45 days the RFC will be rejected and will remain 
+   as part of Addons.
+5. If a sponsor is obtained, and the RFC is approved, a pull request must move the 
+addon along with proper tests.
+6. After merging, the addition will be replaced with an alias to the core function 
+if possible. If an alias is not possible (e.g. large parameter changes), then a deprecation 
+warning will be added and will be removed from TFA after 2 releases. 
+
+
+### Criteria for Migration
```
bhack, line 28: This ok for me on the SIG side. I just want to wait if TF team could put some quantitative metrics here.
Cause the sponsor define the bandwidth of the maintainership but having some quantitative metrics will make the process more transparent (if possible).

sigs/addons/MIGRATION_TO_CORE.md
```diff
@@ -0,0 +1,32 @@
...
+2. The SIG will evaluate the request and add it to the `Potential Candidates` section 
+of our GitHub project.
+3. If it's agreed that a migration makes sense, an RFC needs to be written to discuss 
+the move with a larger community audience. 
+    * If the transition will impact tf-core and Keras then submit the RFC to 
+    [TensorFlow Community](https://github.com/tensorflow/community)
+    * Additions which only subclass Keras APIs should submit their migration proposals to 
+    [Keras Governance](https://github.com/keras-team/governance)
+    
+4. A sponsor from the TF/Keras team must agree to shepard the transition.
```
alextp, line 18: shepherd the transition and take over maintenance of the new API

sigs/addons/MIGRATION_TO_CORE.md
alextp, line 28: ...and someone on the TensorFlow team agrees to take over maintenance of the API

sigs/addons/MIGRATION_TO_CORE.md
gabrieldemarmiesse: It may be problematic since the Addons maintainers don't have write access to it, it may be possible that it's not updated often. If we have control over it, we can take over the responsability of updating it.

sigs/addons/MIGRATION_RFC_TEMPLATE.md
gabrieldemarmiesse: We shouldn't alias it, but use the tf-core function in the body of our implementation. Otherwise we can't add a deprecation warning somewhere. As such it's ok if the parameters differs as long as it offers the same features/functionalities.

sigs/addons/MIGRATION_TO_CORE.md
bhack: @gabrieldemarmiesse I agree with you cause we had not a really good historic records for this repo (see https://github.com/tensorflow/community/issues/226).
I think that the upstreaming/promotion process will enter soon or later in "submarine mode".
I suppose this will happen just after we have surpassed all the requirements/thresholds on the SIGs side and so as soon as the RFC could be formally opened here cause it is complaint with the bare minimal template thresholds/constrains (I hope also quantitative) that we are trying to define together in this review. 
We will have not so much control over the Kanban states cause the process at some point will be almost internal to the TF/core team (internal evaluations, internal activities looking for a sponsors etc.) so the ownership will be of the TF/core team and we can just suppose here that we could have some discrete "emergent" info about this phases of the process that could impact an update of a Kanban states here and/or  RFC labels handling. 
So I suppose we could manage a simplified Kanban on our side that is limited to our ownership of the process and so the last state before the final accepted/rejected could be just something like  "RFC submitted to community repo". 
From this point we need to track the process here cause we are totally dependent from discrete/sparse info emerging from the TF internal process and so if and when the TF team is able to release new info I think it could also handle the ownership of the state in Kanban and labels update.
We could think differently if the process will be totally public so we are on par about all the info related to the process or SIG members could have some different role here in the repo but as I understand with the current repo polcies IMHO this is not the case (see https://github.com/tensorflow/community/issues/227)

sigs/addons/MIGRATION_TO_CORE.md
```diff
@@ -0,0 +1,32 @@
...
+1. Create an issue in TensorFlow Addons for a candidate that you think should be 
+migrated. 
+2. The SIG will evaluate the request and add it to the `Potential Candidates` section 
+of our GitHub project.
+3. If it's agreed that a migration makes sense, an RFC needs to be written to discuss 
+the move with a larger community audience. 
+    * If the transition will impact tf-core and Keras then submit the RFC to 
+    [TensorFlow Community](https://github.com/tensorflow/community)
+    * Additions which only subclass Keras APIs should submit their migration proposals to 
+    [Keras Governance](https://github.com/keras-team/governance)
```
bhack, line 16: The new Keras website is online. I don't know if we need to duplicate/fork this template also in Keras governance and toadd something to the [Keras SIG new webpage](https://keras.io/governance/#governance-the-keras-sig). 
We are in the special case here where a SIG is an upstream of another SIG.

sigs/addons/MIGRATION_RFC_TEMPLATE.md
alextp: I'd like to leave the core implementation the freedom to drop parameters that are not frequently used, and rename other parameters which might have better names.

seanpmorgan: Thanks, yeah even if it were a parameter match we'd still want to call the core function inside of our call with deprecation as mentioned. Updated the discussion points to keep them more general.

sigs/addons/MIGRATION_TO_CORE.md
seanpmorgan: Agreed I think this is more sustainable if the Addons team is able to manage the project.

sigs/addons/MIGRATION_TO_CORE.md
seanpmorgan: Updated.

sigs/addons/MIGRATION_TO_CORE.md
seanpmorgan: Hmm so I think this is a bit of a chicken and egg problem. The sponsor will often be determined after the RFC has been created. I've renamed this as `Criteria for Migration RFC` PTAL and let me know if you think that makes more sense.

sigs/addons/MIGRATION_TO_CORE.md
bhack: I think we could align to 1  month. See https://github.com/tensorflow/community/blob/master/governance/TF-RFCs.md#rfc-process

seanpmorgan: Hmmm, would like to align here but 45 days ensures we overlap with a TFA monthly meeting and lets us get help from TF internal team members to help find sponsors. @theadactyl @ewilderj is there anyway we could make 45 days the standard amount of time for RFC sponsorship? I imagine this would be useful for multiple SIGs that have monthly meetings.

bhack: Yes it is the same if we could extend the standard to 45 days is ok

sigs/addons/MIGRATION_TO_CORE.md
bhack: @seanpmorgan We have other new guidelines for the Keras upstream: https://github.com/keras-team/governance/blob/master/keras_api_design_guidelines.md

sigs/addons/MIGRATION_TO_CORE.md
tanzhenyu: This seems to be helping external contribution for the migration only, not internal contribution for the migration

bhack: @tanzhenyu Do you mean that you want a different process for Google internals?

seanpmorgan:
Per SIG discussion with TF team, we've determined that it is unfeasible for an external SIG to handle the migration process. In the future core TF and Keras member will handle the migration / deprecations:
https://github.com/tensorflow/addons/pull/2156

PR #244, opened 2020-05-13
Transactions extension to Filesystem Plugins RFC

samikama:
In this RFC, we propose extending Filesystem Plugins RFC with addition of Transactions to enable better integration of non-traditional persistent storage systems.


samikama:
Sorry, wrong branch :(

PR #246, opened 2020-05-20
RFC: CSRSparseMatrix

penpornk:
This RFC will be open for comment until Friday, June 12th, 2020.
| **Author(s)** | Penporn Koanantakool (penporn@google.com)            |
| **Sponsor**   | Rasmus Larsen (rmlarsen@google.com), Tatiana Shpeisman (shpeisman@google.com)|
| **Updated**   | 2019-05-28                                           |

## Objective

Support the [compressed sparse row (CSR)](https://en.wikipedia.org/wiki/Sparse_matrix#Compressed_sparse_row_(CSR,_CRS_or_Yale_format)) sparse matrix format in TensorFlow. Implement linear algebra and other common neural networks operations on both CPU and GPU devices for seamless user experience.


ematejska:
@penpornk Looks good from a process point of view.  The once section I might augment is if you plan to have some documentation/tutorial somewhere.  Did you already email developers@tensorflow.org to announce the two week review period?

penpornk:
@ematejska Thank you for taking a look! I've added a short paragraph on future documentation/tutorial. Just sent an announcement to developers@tensorflow.org.

jackd:
Would there be benefit in implementing indices as a `RaggedTensor`, rather than `col_indices`, `row_pointers`, `batch_pointers`? I'm not sure if `RaggedTensor`s are a thing in the c++ API, but I imagine this would make a lot of things easier in the python API, e.g. batching?

Padarn:
Maybe it is totally obvious, but I would appreciate some more details on the expected sparsity in Tensorflow that you're trying to address: `It is generally suitable for the sparsity levels and patterns observed in TensorFlow workloads.`.

jackd:
@Padarn tf-graphics has some use-cases with graph neural networks, though they're currently implemented with `tf.gather` + `tf.reduce_sum` . Performance benefits using COO are inconsistent. See [this PR](https://github.com/tensorflow/graphics/pull/155) for equivalent implementations and the discussion for rough benchmarks.

Padarn:
@jackd Thanks. Thats interesting, I naively assumed the scatter+gather appraoch would be esentially equivalent.

penpornk:
Thank you for the comments!

<br/>

> Would there be benefit in implementing indices as a RaggedTensor, rather than col_indices, row_pointers, batch_pointers? I'm not sure if RaggedTensors are a thing in the c++ API, but I imagine this would make a lot of things easier in the python API, e.g. batching?

@jackd All the processing of the CSR sparse matrix happens in the C++ layer. The Python layer won't see any of the array components (i.e., `batch_pointers`, `row_pointers`, `col_indices`, and `values`) so `RaggedTensor` won't be able to help with batching. 

<br/>

> Maybe it is totally obvious, but I would appreciate some more details on the expected sparsity in Tensorflow that you're trying to address: `It is generally suitable for the sparsity levels and patterns observed in TensorFlow workloads.`.

@Padarn My bad. COO is only efficient when tensors are extremely sparse (e.g., > 99.9% sparse, or have less than one nonzero per fiber/row/column). When tensors get denser, formats that compress the tensor indices in some way would be more efficient (save space, easier traversal, etc). Off the top of my head, weight/activation tensors in deep neural networks usually have 50-90% (or 95%, at best) sparsity. Sparse linear algebra applications could have 90-99.9% sparsity, but will still mostly benefit from CSR more than COO.

I've also added a shorter explanation to the RFC.

Padarn:
Looks great @penpornk. Thanks for the addition.

jackd:
@penpornk is there a reason this isn't exposed as a CompositeTensor?

Separately, can I request the following ops are considered/exposed in python?
* constructor that takes components (`values, column_indices, row_pointers, batch_pointers`)
* properties that return components
* `CSRMatrix.with_values(new_values)` that creates a new matrix with the same sparsity but with `values` replaced with `new_values`, similar to `RaggedTensor.with_values`

penpornk:
@jackd Great question. One of the questions in my discussion section is also about whether it is worth converting `CSRSparseMatrix` to be based on `CompositeTensor`.  

In principle, I agree it should be a `CompositeTensor`. However, The majority of the `CSRSparseMatrix` code ([C++ ops](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/kernels/sparse/), and [Python class](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/python/ops/linalg/sparse/sparse_csr_matrix_ops.py;drc=57395f70dbc24d3ac5f97626520194f6773303e8;l=315)) had been written before `CompositeTensor` existed. (It just got open-sourced after `CompositeTensor`.) So I'm not sure if it's worth the effort converting what we have now to `CompositeTensor`, given that TensorFlow is transitioning to a new backend based on [TFRT](https://blog.tensorflow.org/2020/04/tfrt-new-tensorflow-runtime.html) and [MLIR](https://www.tensorflow.org/mlir) soon, and I don't know how much of the code here can be reused there. If it turns out to be worth it, I'll make it a `CompositeTensor`. 

> Separately, can I request the following ops are considered/exposed in python?

Of course!

> * constructor that takes `components (values, column_indices, row_pointers, batch_pointers)`

Definitely adding this.

> * properties that return components

Right now it's an op called `csr_sparse_matrix_components`, but I can make it a property too (if it passes the eventual API review). [[Example](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/python/kernel_tests/linalg/sparse/csr_sparse_matrix_ops_test.py;drc=815fa1866caa88727d2d5288c028f0e610c4bda3;l=157-159)][[Spec](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/api_def/base_api/api_def_CSRSparseMatrixComponents.pbtxt?q=csrsparsematrixcomponent&ss=tensorflow%2Ftensorflow)]

> * `CSRMatrix.with_values(new_values)` that creates a new matrix with the same sparsity but with `values` replaced with `new_values`, similar to `RaggedTensor.with_values`

Sounds great. Sure!

Padarn:
Congrats. Looking forward to it.

penpornk:
@Padarn Thank you! :)

@jackd It doesn't seem to take too much effort to convert `CSRSparseMatrix` to `CompositeTensor`. I'll make it a `CompositeTensor`.  I've also updated the RFC to include the features you requested.

penpornk:
**Design review notes**

CompositeTensor
* Ed Loper: Making `CSRSparseMatrix` a `CompositeTensor` shouldn't require much work.
  * Define a new `CSRSparseMatrixSpec` class (a subclass of `TypeSpec`). 
    * It would have properties `shape` and `dtype`. 
    * `_to_components(self, value)` would return `value._csr_matrix`. 
    * `_from_components(self, components)` would return `CSRSparseMatrix(components)`. 
    * `_component_specs` would return `TensorSpec([], dtypes.variant)`. 
    * `_serialize(self)` would return `(self.shape, self.dtype)`.
  * Add `CompositeTensor` as a base class to `CSRSparseMatrix`.
  * Add a `_type_spec` property to `CSRSparseMatrix` which returns an appropriate `CSRSparseMatrixSpec`.
* Ed Loper: `CompositeTensor` will make it more straightforward to use `CSRSparseMatrix` with control flow ops (`tf.cond`, `tf.while_loop`), `tf.function`, `SavedModel`, Keras, etc.

TensorFlow op dispatcher
* Ed Loper: Any reason not to use [TensorFlow op dispatcher](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/python/util/dispatch.py;l=46) to allow standard TensorFlow ops to take `CSRSparseMatrix`?
* Rasmus Larsen: We think making the CSR representation live in a "walled API garden" with well-defined entry and exit points was preferable to making some subset of TF ops transparently support the new type.
  * Pros: It makes it easy to
    * Control the API and underlying kernels.
    * See which operations and variables are sparse.
    * Understand performance trade-offs.
  * Cons: It enlarges the API surface of TF as a whole.
* Alex Passos: I think using op dispatcher in this case would give strictly worse user experience.
* Eugene Brevdo: Some ops change sparsity pattern and we have to be careful on what ops to support.

`tf.function`
* Alex Passos: `tf.function` should be able to take `DT_VARIANT` type.

Naming
* Ed Loper: Why `Matrix` and not `Tensor`?
* Penporn K.: CSRSparseMatrix is essentially a batch of CSR matrices.
* Eugene Brevdo: `CSRSparseMatrix` is part of a family of `SparseMatrix` classes - the first one - and the generic class name is `SparseMatrix` (though it hides the underlying type). We support only matrix operations right now, and for the foreseeable future.
* Rasmus Larsen: We do allow a batch dimension, but only operate on the inner most matrices pairwise. For example, it can't do generic tensor contractions or 3D convolutions.

BSR
* Yunlu Li: Would the BSR format have the same API as CSR?
* Penporn K.: Yes. There would be a `BSRSparseMatrix` class which shares the same `SparseMatrix` superclass as `CSRSparseMatrix`. Ops would be able to take `BSRSparseMatrix` too.

Connection with TFLite
* Raziel Alvarez: It the short term, it would be great if this can allow TFLite to express block sparsity in the current pruning API. Currently we have to pass two separate TF tensors for one sparse TFLite tensor.
* Penporn K.: Noted and will try to see if we can augment `CSRSparseMatrix` somewhat to accommodate this. If we can't, at least `BSRSparseMatrix` would be able to do it.

penpornk:
@ematejska Please don't merge this PR yet. I'd like to add a reference (that will be available next Monday) to the PR. I'll let you know once it's ready. Thank you!

penpornk:
@ematejska The PR is ready for merging. Thank you very much! :)

tabergma:
@penpornk Question regarding `CSRSparseMatrix` (using Tensorflow 2.2): If I create a `CSRSparseMatrix` via 
```
from tensorflow.python.ops.linalg.sparse import sparse_csr_matrix_ops as csr

A_indices = np.array([[0, 0], [2, 3], [2, 4], [3, 0]])
A_values = [1.0, 5.0, -1.0, -2.0]
A_dense_shape = [5, 6]

A_csr = csr.sparse_tensor_to_csr_sparse_matrix(A_indices, A_values, A_dense_shape)
```
the resulting tensor `A_csr` does not have any shape information or dtype set. When I would create a `SparseTensor`, I would see the shape, dense_shape, indices, and values. Is there a plan to add this kind of information?
Also you could distinguish between `tf.Tensor` and `tf.SparseTensor` by checking `isinstance(tensor, tf.SparseTensor)` in python. Is there a way to distinguish a `CSRSparseMatrix` from a normal dense tensor?

Thanks!

penpornk:
@tabergma 

> `A_csr `does not have any shape information or dtype set. When I would create a `SparseTensor`, I would see the shape, dense_shape, indices, and values. Is there a plan to add this kind of information?

Yes, I plan to make `CSRSparseMatrix` a subclass of `CompositeTensor` in the future. `CompositeTensor` provides these properties. (`SparseTensor` and `RaggedTensor` are both `CompositeTensor`s.)

Currently, there is also a helper function that extracts shape and dtype info, but it only works in graph mode (i.e., v1, non-eager mode).
```python
import tensorflow.compat.v1 as tfv1
tfv1.disable_v2_behavior()
# The call to create CSRSparseMatrix must be in graph mode, 
# or dense_shape_and_type would raise a 'missing handle data' error.
A_csr = csr.sparse_tensor_to_csr_sparse_matrix(A_indices, A_values,
                                               A_dense_shape)
shape_and_type = csr.dense_shape_and_type(A_csr)
print(shape_and_type.shape)
print(shape_and_type.dtype)
```

> Is there a way to distinguish a CSRSparseMatrix from a normal dense tensor?

Yes. The following snippet works in both graph and eager mode. :) 
```python
isinstance(A_csr, csr.CSRSparseMatrix)
```

tabergma:
@penpornk Thanks for the answer! Looking forward to the update. Just one follow up question:
The following snippet 
```
from tensorflow.python.ops.linalg.sparse import sparse_csr_matrix_ops as csr

A_indices = np.array([[0, 0], [2, 3], [2, 4], [3, 0]])
A_values = [1.0, 5.0, -1.0, -2.0]
A_dense_shape = [5, 6]

A_csr = csr.sparse_tensor_to_csr_sparse_matrix(A_indices, A_values, A_dense_shape)

print(isinstance(A_csr, csr.CSRSparseMatrix))
```
has the output
```
False
```
So the `isinstance` check does not seem to work. Any idea why? Any particular version I need to run? Thanks for your help!

penpornk:
@tabergma Oops. Sorry about that! I had tested that snippet in a script that did many things and misread the output. Please try this instead. I've verified that it works (in both eager and graph modes):
```python
print(isinstance(A_csr, (tf.Tensor, csr.CSRSparseMatrix)))
```

PR #247, opened 2020-05-21
RFC: TensorFloat-32 support in TensorFlow

reedwm:
This RFC will be open for comment until Wednesday, June 3rd, 2020.

| **RFC #**     | [247](https://github.com/tensorflow/community/pull/247) |
| **Author(s)** | Reed Wanderman-Milne (reedwm@google.com)             |
| **Sponsor**   | Sanjoy Das (sanjoy@google.com)                 |
| **Updated**   | 2020-05-20                                           |

## Objective

Allow [TensorFloat-32](https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format) to be used in TensorFlow to improve performance.

### DIFF
--- a/rfcs/20200520-tensor-float-32.md
+++ b/rfcs/20200520-tensor-float-32.md
@@ -47,7 +47,7 @@ tf.config.tensor_float_32_execution_allowed()
 
 Since TF32 only affects Ampere GPUs, moving an op to a GPU can affect numerics. Grappler and other graph optimizations will not consider this, and will freely move ops between devices without regard to numeric stability. As a result, explicitly putting an op on the CPU does not ensure it will use the full float32 precision instead of TF32.
 
-Since TensorFlow 2.3 will not support CUDA 11, which is required for TF32, this API will first be exposed in TensorFlow 2.4. However, Google Cloud will likely cherrypick CUDA 11 and this API into their version of 2.3, so they can offer TF32 support to their customers who use TensorFlow 2.3.
+Since TensorFlow 2.3 will not support CUDA 11, which is required for TF32, this API will first be exposed in TensorFlow 2.4. However, downstream repackagers of TensorFlow (such as Google Cloud) are encouraged to cherrypick CUDA 11 and this API into their version of 2.3, so they can offer TF32 support to their customers who use TensorFlow 2.3.
 
 
 ### Turning TF32 on by default
--- a/rfcs/20200520-tensor-float-32.md
+++ b/rfcs/20200520-tensor-float-32.md
@@ -1,6 +1,6 @@
 # TensorFloat-32 in TensorFlow
 
-| Status        | Proposed                                             |
+| Status        | Accepted                                             |
 :-------------- |:---------------------------------------------------- |
 | **RFC #**     | [247](https://github.com/tensorflow/community/pull/247) |
 | **Author(s)** | Reed Wanderman-Milne (reedwm@google.com)             |
--- a/rfcs/20200520-tensor-float-32.md
+++ b/rfcs/20200520-tensor-float-32.md
@@ -1,0 +1,80 @@
+# TensorFloat-32 in TensorFlow
+
+| Status        | Proposed                                             |
+:-------------- |:---------------------------------------------------- |
+| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)|
+| **Author(s)** | Reed Wanderman-Milne (reedwm@google.com)             |
+| **Sponsor**   | Sanjoy Das (sanjoy@google.com)                 |
+| **Updated**   | 2020-05/20                                           |
+
+## Objective
+
+Allow [TensorFloat-32](https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format) to be used in TensorFlow to improve performance.
+
+## Motivation
+
+NVIDIA Ampere, an upcoming generation of NVidia GPUs announced at GTC 2020, introduces a new numeric format called TensorFloat-32, or TF32 for short.
+TF32 has the range of float32/bfloat16 (i.e. 8 bits of exponent) and the precision of fp16 (i.e. 10 bits of mantissa).
+For the most part, it is not an in-memory format, but tensor cores natively support it as a computation format.
+TF32 should not be thought of as an in-memory dtype but instead a computation mode that increases performance and decreases numeric precision for certain float32 operations.
+Nvidia has not found any cases where TF32 reduces the convergence of deep learning models.
+
+Upcoming versions of cuDNN, cuBLAS, and other CUDA libraries will expose a mode of execution that has float32 inputs and outputs, but internally truncates float32 to TF32 and uses tensor cores.  This is expected to be sufficiently accurate to reach the same convergence as the “full” float32 mode of execution but significantly faster.  Each element still takes four bytes, so there is still a memory and performance penalty compared to using float16 or bfloat16.
+
+As TF32 is only usable by tensor cores, it can only be used for matrix multiplications and other ops implemented in terms of matrix multiplications, such as convolutions. It is not used for pointwise ops or reductions.
+
+TF32 will benefit users who run float32 models on Ampere GPUs, so we need an API to allow these users to enable TF32. 
+
+## Design Proposal
+
+In TensorFlow, TF32 can be enabled for supported ops on Ampere GPUs with the following call:
+
+```python
+tf.config.allow_tensor_float_32_execution(True)
+```
+
+The word "allow" emphasizes only certain devices (Ampere GPUs) and ops (such as matmuls and convolutions) will be affected. Once enabled, all local and remote Ampere GPUs use TF32 for supported float32 ops.
+
+Passing `False` to `allow_tensor_float_32_execution` will disable TF32 if already enabled.
+
+We call the function "allow_tensor_float_32_execution" instead of the more concise "allow_tf32_execution" because people may mistakenly interpret the phrase "tf32" to refer to TensorFlow instead of TensorFloat. 
+
+The following can be used to query whether TF32 is enabled. The function returns a bool.
+
+```python
+tf.config.tensor_float_32_execution_allowed()
+```
+
+Since TF32 only affects Ampere GPUs, moving an op to a GPU can affect numerics. Grappler and other graph optimizations will not consider this, and will freely move ops between devices without regard to numeric stability. As a result, explicitly putting an op on the CPU does not ensure it will use the full float32 precision instead of TF32.
+
+Since TensorFlow 2.3 will not support CUDA 11, which is required for TF32, this API will first be exposed in TensorFlow 2.4. However, Google Cloud will likely cherrypick CUDA 11 and this API into their version of 2.3, so they can offer TF32 support to their customers who use TensorFlow 2.3.
+
+
+### Turning TF32 on by default
+
+Numerical studies by NVIDIA covering many common models suggest that TF32 is numerically robust for deep learning applications. In order to take advantage of these new accelerations in Ampere hardware for float32 models, we would like to enable TF32 by default. However, since the TensorFlow 2.4 release is still months away and we intend to use that time to further test and evaluate TF32, it is too early to decide in this RFC whether TF32 execution will be enabled or disabled by default. Here we begin a discussion by listing the most likely scenarios. Comments are also welcome. The scenarios are:
+
+1. Turn it on by default in 2.4, the first release with the TF32 API.
+2. Turn it on by default in 2.5, the second release with the TF32 API.
+3. Do not turn it on by default.
+
+
+The advantage of (1) is that all Ampere float32 users get the performance benefit unless they opt out. Additionally, Ampere numerics will not be loosened in a new release: TensorFlow 2.4 will be the first release with Ampere support, and it will immediately default to TF32 being enabled. The disadvantage is that we cannot collect as much feedback from users before defaulting to TF32, because no stable version of TensorFlow will support TF32 but not have it enabled by default.
+
+The advantage of (2) is that it allows users to test and give feedback on TF32 with a stable version of TensorFlow before we decide whether it should be default. The disadvantage is it’s possible we break Ampere users who relied on the full float32 precision in 2.4 when they upgrade to 2.5
+
+The advantage of (3) is that a user’s model will never break due to using reduced precision, even if they upgrade from an earlier GPU to Ampere. The disadvantage is that many Ampere users would not get the performance benefit from TF32 as they would not know about the API to enable it.
+
+Another advantage of turning on TF32 by default is that it makes TensorFlow’s behavior with GPUs more consistent with TPUs. TPUs internally use lower precision for float32 matmuls and convolutions, similar to how Ampere GPUs will use lower precision for float32 matmuls and convolutions if TF32 is enabled.
+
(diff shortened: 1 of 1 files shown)


rfcs/20200520-tensor-float-32.md
```diff
@@ -0,0 +1,80 @@
...
+| **Sponsor**   | Sanjoy Das (sanjoy@google.com)                 |
+| **Updated**   | 2020-05/20                                           |
+
+## Objective
+
+Allow [TensorFloat-32](https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format) to be used in TensorFlow to improve performance.
+
+## Motivation
+
+NVIDIA Ampere, an upcoming generation of NVidia GPUs announced at GTC 2020, introduces a new numeric format called TensorFloat-32, or TF32 for short.
```
sanjoy, line 16: Link to announcement?

rfcs/20200520-tensor-float-32.md
```diff
@@ -0,0 +1,80 @@
...
+TF32 has the range of float32/bfloat16 (i.e. 8 bits of exponent) and the precision of fp16 (i.e. 10 bits of mantissa).
+For the most part, it is not an in-memory format, but tensor cores natively support it as a computation format.
```
sanjoy, line 18: Why "For the most part"?

rfcs/20200520-tensor-float-32.md
```diff
@@ -0,0 +1,80 @@
...
+TF32 should not be thought of as an in-memory dtype but instead a computation mode that increases performance and decreases numeric precision for certain float32 operations.
+Nvidia has not found any cases where TF32 reduces the convergence of deep learning models.
```
sanjoy, line 20: I believe this should be NVIDIA as per the [NVIDIA Brand Guidelines](https://www.nvidia.com/content/dam/en-zz/Solutions/about-us/NVIDIA-Brand-Guidelines-for-NVIDIA-Partner-Network-v03-3-5-19.pdf).

rfcs/20200520-tensor-float-32.md
```diff
@@ -0,0 +1,80 @@
...
+
+The following can be used to query whether TF32 is enabled. The function returns a bool.
+
+```python
+tf.config.tensor_float_32_execution_allowed()
+```
+
+Since TF32 only affects Ampere GPUs, moving an op to a GPU can affect numerics. Grappler and other graph optimizations will not consider this, and will freely move ops between devices without regard to numeric stability. As a result, explicitly putting an op on the CPU does not ensure it will use the full float32 precision instead of TF32.
+
+Since TensorFlow 2.3 will not support CUDA 11, which is required for TF32, this API will first be exposed in TensorFlow 2.4. However, Google Cloud will likely cherrypick CUDA 11 and this API into their version of 2.3, so they can offer TF32 support to their customers who use TensorFlow 2.3.
```
sanjoy, line 50: The "However, Google Cloud will likely cherrypick CUDA 11 and this API into their version of 2.3..." sentence should not be part of the RFC I believe.

rfcs/20200520-tensor-float-32.md
```diff
@@ -0,0 +1,80 @@
...
+
+1. Turn it on by default in 2.4, the first release with the TF32 API.
+2. Turn it on by default in 2.5, the second release with the TF32 API.
+3. Do not turn it on by default.
+
+
+The advantage of (1) is that all Ampere float32 users get the performance benefit unless they opt out. Additionally, Ampere numerics will not be loosened in a new release: TensorFlow 2.4 will be the first release with Ampere support, and it will immediately default to TF32 being enabled. The disadvantage is that we cannot collect as much feedback from users before defaulting to TF32, because no stable version of TensorFlow will support TF32 but not have it enabled by default.
```
sanjoy, line 62: > because no stable version of TensorFlow will support TF32 but not have it enabled by default.

I don't buy this: the models that TF32 targets use FP32 today, so I'd expect users to notice a regression even if 2.4 enables it by default, which they can corroborate further by comparing the accuracy with disabling it explicitly.

rfcs/20200520-tensor-float-32.md
reedwm: Why not? It provides information on when the API will be available. And also motivation for why we need to have the RFC so early despite TF 2.4 not coming out for months.

rfcs/20200520-tensor-float-32.md
reedwm: I don't fully understand your argument. We'd like to have a release where users can try TF32 and give us feedback before we decide to whether to turn it on by default. If we immediately turn it on by default in 2.4, users can still give feedback, but it will be too late: we will have already made our decision.

rfcs/20200520-tensor-float-32.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Remote devices
+
+Enabling TF32 will affect remote Ampere GPUs in addition to local Ampere GPUs. In particular, it will affect devices on hosts connected to via [`tf.config.experimental_connect_to_host`](https://www.tensorflow.org/api_docs/python/tf/config/experimental_connect_to_host) or [`tf.config.experimental_connect_to_cluster`](https://www.tensorflow.org/api_docs/python/tf/config/experimental_connect_to_cluster). The initial, unexposed version of the function in TensorFlow 2.3 may only support local devices, not remote devices, if we do not have time to implement remote device support.
```
byronyi, line 74: Any specific additional efforts needed here to support remote devices?

rfcs/20200520-tensor-float-32.md
```diff
@@ -0,0 +1,80 @@
...
+As TF32 is only usable by tensor cores, it can only be used for matrix multiplications and other ops implemented in terms of matrix multiplications, such as convolutions. It is not used for pointwise ops or reductions.
+
+TF32 will benefit users who run float32 models on Ampere GPUs, so we need an API to allow these users to enable TF32. 
+
+## Design Proposal
+
+In TensorFlow, TF32 can be enabled for supported ops on Ampere GPUs with the following call:
+
+```python
+tf.config.allow_tensor_float_32_execution(True)
```
byronyi, line 33: Why not using the Keras mixed precision policy API?

rfcs/20200520-tensor-float-32.md
reedwm: I haven't worked this out yet, which is why I state this might not be done for TensorFlow 2.3. This will likely be done by adding a field to the [CreateContextRequest](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/protobuf/eager_service.proto;l=85;drc=8588e0aab8c1ef6a4214bcc2f7d0bb61578a88b3) proto to indicate whether TF32 is used.

rfcs/20200520-tensor-float-32.md
reedwm: This affects ops outside Keras, so it shouldn't be under `tf.keras`. In a sense, TF32 is a form of mixed precision, as some ops use TF32 and others use float32. We could put it under `tf.mixed_precision`, but I think `tf.config` is better since tf32 should be though of as a mode, not a dtype.

rfcs/20200520-tensor-float-32.md
sanjoy: > but it will be too late: we will have already made our decision.

Is the assumption that disabling tf32 by default (if users report problems after we enable it by default in 2.4) is more of a breaking change than enabling it by default (if users try it with 2.4 and don't report problems)?

rfcs/20200520-tensor-float-32.md
alextp: Also the mixed precision API mostly changes the dtype of tensors, while tf32 doesn't affect tensor dtype (afaict) just the dtype of accumulators inside ops.

rfcs/20200520-tensor-float-32.md
alextp: Replacing "google cloud will likely" with "downstream repackagers of tensorflow (such as google cloud) are encouraged to" will make this read better

rfcs/20200520-tensor-float-32.md
```diff
@@ -0,0 +1,80 @@
...
+
+The advantage of (3) is that a user’s model will never break due to using reduced precision, even if they upgrade from an earlier GPU to Ampere. The disadvantage is that many Ampere users would not get the performance benefit from TF32 as they would not know about the API to enable it.
+
+Another advantage of turning on TF32 by default is that it makes TensorFlow’s behavior with GPUs more consistent with TPUs. TPUs internally use lower precision for float32 matmuls and convolutions, similar to how Ampere GPUs will use lower precision for float32 matmuls and convolutions if TF32 is enabled.
+
+**If you know of any models whose accuracy may be impacted by TF32, please comment on this RFC.** Note that TF32 is equivalent to float32 except it has 10 bits of mantissa instead of 23 bits. It will initially be used only for matmuls and convolutions, but may be used for other ops in the future if they are implemented in terms of a matmul. Once TensorFlow 2.4 is released, you will be able to test the impact of TF32 on your models if you have Ampere GPUs. You will be able to test earlier if you use Tensorflow nightly packages, and even earlier if you build from source with CUDA 11 support.
```
alextp, line 70: You might want to indicate a way to receive private feedback about this too.

rfcs/20200520-tensor-float-32.md
reedwm: Done.

rfcs/20200520-tensor-float-32.md
reedwm: No, enabling tf32 is probably more of a breaking change. However, we only want to make such a change at most once. After enabling tf32, I don't think we should subsequently disable it.

rfcs/20200520-tensor-float-32.md
reedwm: Do you think it's likely someone would be willing to share with us but not publicly? I could recommend emailing me for private feedback, but I would rather people post feedback publicly since I want to be transparent about why we make whatever decision we make.

byronyi: +1 for the transparency.

rfcs/20200520-tensor-float-32.md
```diff
@@ -0,0 +1,80 @@
...
+
+In TensorFlow, TF32 can be enabled for supported ops on Ampere GPUs with the following call:
+
+```python
+tf.config.allow_tensor_float_32_execution(True)
+```
+
+The word "allow" emphasizes only certain devices (Ampere GPUs) and ops (such as matmuls and convolutions) will be affected. Once enabled, all local and remote Ampere GPUs use TF32 for supported float32 ops.
+
+Passing `False` to `allow_tensor_float_32_execution` will disable TF32 if already enabled.
```
karmel, line 38: What are the use-cases for toggling between these? Are there potentially issues with moving between the two in a single program?

rfcs/20200520-tensor-float-32.md
```diff
@@ -0,0 +1,80 @@
...
+
+The advantage of (3) is that a user’s model will never break due to using reduced precision, even if they upgrade from an earlier GPU to Ampere. The disadvantage is that many Ampere users would not get the performance benefit from TF32 as they would not know about the API to enable it.
+
+Another advantage of turning on TF32 by default is that it makes TensorFlow’s behavior with GPUs more consistent with TPUs. TPUs internally use lower precision for float32 matmuls and convolutions, similar to how Ampere GPUs will use lower precision for float32 matmuls and convolutions if TF32 is enabled.
+
+**If you know of any models whose accuracy may be impacted by TF32, please comment on this RFC.** Note that TF32 is equivalent to float32 except it has 10 bits of mantissa instead of 23 bits. It will initially be used only for matmuls and convolutions, but may be used for other ops in the future if they are implemented in terms of a matmul. Once TensorFlow 2.4 is released, you will be able to test the impact of TF32 on your models if you have Ampere GPUs. You will be able to test earlier if you use Tensorflow nightly packages, and even earlier if you build from source with CUDA 11 support.
```
karmel, line 70: Are there users that NVIDIA can help us find directly?

rfcs/20200520-tensor-float-32.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Design Proposal
+
+In TensorFlow, TF32 can be enabled for supported ops on Ampere GPUs with the following call:
+
+```python
+tf.config.allow_tensor_float_32_execution(True)
+```
+
+The word "allow" emphasizes only certain devices (Ampere GPUs) and ops (such as matmuls and convolutions) will be affected. Once enabled, all local and remote Ampere GPUs use TF32 for supported float32 ops.
```
karmel, line 36: Should an error be raised (or a warning) if allow=True and the device does not support TF32? One could imagine users being surprised that no complaint is raised when when this mode is requested. I guess in that case the flag would be "use_tensor_float_32_execution" instead of allow... but maybe explicit is preferable here?

rfcs/20200520-tensor-float-32.md
sanjoy: We don't have to _recommend_ it, but not everyone may be at freedom to talk about what they're working on in a public forum.  So mentioning a private channel seems like a good idea.

rfcs/20200520-tensor-float-32.md
reedwm: Nvidia will probably directly collect feedback and tell us. They have already tested themselves on many models.

rfcs/20200520-tensor-float-32.md
byronyi: Does the UpdateContext RPC support this particular use case? I imagine one could use the `cluster_device_attributes` field to pass on the TF32 mode toggle, but it looks like it will update the remote session when `cluster_device_attributes` is not empty in current codebase. Not sure if setting TF32 worth carrying out such a heavy(?) operation.

rfcs/20200520-tensor-float-32.md
byronyi: So it will not be part of the cluster_device_attributes, but a new field? (Also see my comment below on updating remote context)

rfcs/20200520-tensor-float-32.md
reedwm: Right now UpdateContext is only  used to add/remove machines/devices. But we can add a field to allow it to also update whether TF32 is used. We can altneratively add an option to [QueueItem](https://github.com/tensorflow/tensorflow/blob/6f22fa93766d18f45e2baac140bded0bc3e5e681/tensorflow/core/protobuf/eager_service.proto#L57) in case there are ordering concerns between enabling/disabling TF32 and executing an op.

I asked internally, and `cluster_device_attributes` is only useful for propogating device information to other machines in the cluster. It is not intended for use in communicating information about how ops should run, but only fundamental information about the devices themselves.

See [this post](https://github.com/tensorflow/tensorflow/pull/39764#issuecomment-635055025) for some more context.

> Not sure if setting TF32 worth carrying out such a heavy(?) operation.

Yeah, this will be a heavy operation. But users should only set/unset TF32 very rarely: only at the beginning of the model and between running one model and the next (or one test and the next). So I think it's OK.

I updated the RFC based on this discussion. I added a paragraph to the "Remote Devices" section and added a new paragraph to "Alternatives considered"

rfcs/20200520-tensor-float-32.md
reedwm: Yes, we will add a new field. See my reply to your other comment.

PR #249, opened 2020-05-26
RFC: Elasticity and Fault tolerance in Multi-WorkerMirroredStrategy

sboshin:
# Elasticity and Fault tolerance in Multi-WorkerMirroredStrategy (MWMS)

|Status	|Proposed|
|---	|---	|
|**RFC #**	|[NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)	|
|**Author(s)**	|Samuel Oshin ([sboshin@amazon.com](mailto:me@example.org))|
|**Sponsor**	|A N Expert ([whomever@tensorflow.org](mailto:whomever@tensorflow.org))	|
|**Updated**	|2020-05-25	|

## Objective

Elastic training will enable distributed training to leverage dynamic clusters. This will help users leverage pre-emptible instances, allows users to start training faster, and potentially reduce cost to train. 
The goal is to enable elastic training in the Multi-Worker Mirrored Strategy.


rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
```diff
@@ -0,0 +1,439 @@
...
+
+Currently in TF, keras has implemented a fault tolerance mechanism. When a node gets pre-empted and is able to restart, keras will wait for the node to come back online, and [everyone will restart from the last saved checkpoint](https://www.tensorflow.org/tutorials/distribute/multi_worker_with_keras#fault_tolerance). 
+
+### **_Design Summary_**
+
+This design aims to augment the Multi-Worker Mirrored Strategy to add elasticity while avoiding start/initialization times. In the event of cluster shrinkage or expansion, we don’t want to restart the process/session, but want session to adapt to the cluster changes, avoiding re-compilation, re-optimization, and other usual first time costs. 
+
+
+### **_CAVEAT: Model Determinism_**
```
byronyi, line 105: The local/global batch size might change, and the model optimizer needs to take that into account. It’s a little bit more than just “model indeterminism”, and I suggest you to take that into account in this RFC.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
```diff
@@ -0,0 +1,439 @@
...
+
+To add elasticity and fault tolerance to MWMS, the system needs to be fault tolerant, and expandable. For fault tolerance, certain faults need to be identifiable and handled explicitly such as communication errors, user error, node errors. These faults will have certain impacts to the system which can reduce the system size and force the system to recover a previously known state. To expand, the system needs to be able to identify new nodes and modify the system to use the new nodes.
+
+Currently in TF, keras has implemented a fault tolerance mechanism. When a node gets pre-empted and is able to restart, keras will wait for the node to come back online, and [everyone will restart from the last saved checkpoint](https://www.tensorflow.org/tutorials/distribute/multi_worker_with_keras#fault_tolerance). 
+
+### **_Design Summary_**
+
+This design aims to augment the Multi-Worker Mirrored Strategy to add elasticity while avoiding start/initialization times. In the event of cluster shrinkage or expansion, we don’t want to restart the process/session, but want session to adapt to the cluster changes, avoiding re-compilation, re-optimization, and other usual first time costs.
```
byronyi, line 102: Membership changing adds overhead, too. I’m not convinced that the performance gain outweighs the extra design complexity.

It would be better to add some preliminary experiment results and/or a prototype of this design here.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
```diff
@@ -0,0 +1,439 @@
...
+* Benchmarks comparing dedicated nodes vs nodes with planned failure to measure impact of elasticity
+* End to end tests will utilize the Multiprocess runner and run small training jobs with planned failures. These tests will be added to collective_all_reduce_strategy_test.py
+
+### Tutorials and Examples
+
+* The run script to specify min/max nodes will be a new cli
+    * tensorflowrun —min_nodes X —max_nodes Y —discovery_script discover.sh —hostfile hostfile python training_script.py
+* Alternative
+* `multiworker_strategy = tf.distribute.experimental.MultiWorkerMirroredStrategy(min_nodes=X, max_nodes=Y)`
+* A tutorial on creating discovery scripts
```
byronyi, line 427: TF ClusterResolver supports ClusterSpec update and propagation. Is there any other specific features we need to support using a discovery script, and if there is, could we extend the current TF ClusterResolver?

byronyi:
Thanks for the proposal! It would a valuable feature for distributed model training. 

However, I’m not sure current RFC is sufficient for a feasible implementation of your design. Given the complexity it adds to the TF distributed runtime, some preliminary prototypes might be helpful.

Some discussion before: https://github.com/tensorflow/tensorflow/issues/18880#issuecomment-390705270

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
```diff
@@ -0,0 +1,439 @@
...
+
+
+## _**Adding Elasticity**_
+
+To add elasticity and fault tolerance to MWMS, the system needs to be fault tolerant, and expandable. For fault tolerance, certain faults need to be identifiable and handled explicitly such as communication errors, user error, node errors. These faults will have certain impacts to the system which can reduce the system size and force the system to recover a previously known state. To expand, the system needs to be able to identify new nodes and modify the system to use the new nodes.
+
+Currently in TF, keras has implemented a fault tolerance mechanism. When a node gets pre-empted and is able to restart, keras will wait for the node to come back online, and [everyone will restart from the last saved checkpoint](https://www.tensorflow.org/tutorials/distribute/multi_worker_with_keras#fault_tolerance).
```
byronyi, line 98: I might miss something here, but it seems to me that the proposed recovery mechanism is similar to an in-memory save/restore of checkpoints?

sboshin: Yes, in memory checkpoint, this was more of a performance concern depending on how often we save states.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
sboshin: One concern for complete session tear down and bring up is the first pass. There is an issue which has stated 3mins for startup is in the expected range, which is a huge perf impact. 

That sounds good, ill start working on getting some pre-lim numbers together.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
sboshin: Do you mean adjusting the learning rate as the local batch size changes? The current plan was to keep the global batch size constant, by augmenting the strategy's method of distributing the dataset (autosharding).

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
sboshin: Do you mean to bake in the discovery script into the TF Cluster resolver? That sounds good to me. 
By update and propagation, could you point me to where this is? I wasn't able to find this feature. Also do you mean update and propagate the cluster spec or the underlying grpc server?

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
byronyi: Then you might have layers such as BatchNorm dealing with different local batch size.

byronyi: Not all global batch size are divisible by scaled number of workers. And for XLA compiled kernels, adjusting local batch size will trigger re-compilation as well. That being said, we might come up with some remedies for these problems.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
byronyi: It would be great if we could go deeper and analyze which parts of the initialization overhead are essential, and which parts could be delayed or optimized out. The analysis might help even we decide to go with saving/restoring in-memory checkpoints.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
byronyi: I saw that being discussed in alternative design 1. Thanks!

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
```diff
@@ -0,0 +1,439 @@
...
+When a new process joins the run call, the dataset of the new nodes and the old nodes will be out of sync. Each dataset operator internally has an iterator object which can be restored through MakeIteratorFromCheckpoint. The chief will send its serialized iterator to be restored on the new nodes. This will bring the new nodes to the last iterated state of the dataset. 
+
+
+### _Membership Change Process_
+
+The membership change process happens when the cluster needs to be modified. This will take place when new nodes are being added to the cluster, or there is a NetworkException raised. The membership change process will attempt to figure out failed nodes, new nodes, and send out cluster spec updates to each worker. 
+
+* Everyone ping chief
+    * timeout for chief to respond
+    * If chief doesn’t respond ping next chief candidate
```
yuefengz, line 145: It seems difficult for all workers to agree on the failure state of the chief without any coordination.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
```diff
@@ -0,0 +1,439 @@
...
+* broadcast new cluster spec
+* if number of nodes doesn’t meet minimum threshold, restart membership process
+
+### _Update Cluster_
+
+The update cluster process happens when a node receives a new cluster spec. This cluster spec needs to modify the worker node to discover or blacklist nodes. Currently TensorFlow doesn’t allow for dynamic cluster sizes. There are several identified things that need to be changed in order to allow for dynamic sizes. 
+
+* Collective algorithms will need to be dynamic
+    * There are a node def attributes that need to become inputs that can be set at runtime
+        * **group_size** and **group_key** will need to be Inputs
```
yuefengz, line 157: You can just rebuild the graph/function without needing to passing new group sizes or keys.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
byronyi: That’s also my concern. We probably need a consensus algorithm, e.g. Paxos or Raft, if we wanted to do leader election in a proper way.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
byronyi: Does rebuilding graph/function costs a lot if group size changes? I could imagine that the ring reduce algorithm might need a different packing scheme for different number of workers.

I’m not sure if that’s part of the 3 minutes initialization mentioned above...

sboshin: I think more concrete numbers will be needed to understand the impact of rebuilding the graph. I've been going off other issues which have stated up to 3mins is normal. I will look into getting accurate measurements on several models.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
sboshin: > Thank you for your RFC! Do you think your proposal can be a layer above the existing MWMS, so that we can experiment with it and allow users to use it first while we think of ways to let it graduate it into TensorFlow core? I guess this is possible since your `discovery_script` is not coupled with any of TF's components. You can also extend MWMS for dataset-related changes.

Thank you!
Is the idea a separate repository or just a separate Strategy? This current idea has Core changes implied. A separate repo can be done once the graph rebuilding costs are understood.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
yuefengz: We've optimized graph building time and it should be pretty fast these days.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
yuefengz: I think most of the changes can be in a separate repository (we've been thinking of creating a new repo). It is reasonable if some changes are needed in the TF core. But you can start with a prototype or a demo using monkey-patch, linking new kernels etc.

yuefengz: Re consensus algorithm, I am not sure whether it is a good an idea to require users to configure these services. Another option is similar to [the single-client PS design](https://github.com/yuefengz/community/blob/master/rfcs/20200306-single-client-parameter-server.md) which uses a single coordinator for the whole training cluster. But it assumes the client won't fail, otherwise distributed consensus is still needed. These elastic resources (e.g. elastic dataset) may be better fit in the client layer (we'll separate the client and the strategy class).

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
sboshin: Just trying to gather some data on a variety of models. This isn't indepth. 
| Strategy | Model | setup (sec) | step 1 (sec) | step 2 (sec) | step 3 (sec) | step 4 (sec) |
|----------|:-----:|:------:|:------:|:------:|:------:|:-----:|
| mirrored | mnist| 1 g4dn 4 gpus| 9.31 | 0.25 | 0.02 | 0.01|
| mirrored | resnet ctl| 1 g4dn 4 gpus| 30.53 | 0.13 | 0.13 | --|
| mirrored | resnet keras| 1 g4dn 4 gpus| 39.44 | 1.43 | 0.78 | 0.78|
| mirrored | efficientnet| 1 g4dn 4 gpus| 46.58 | 0.91 | 0.29 | 0.28|
| multi-worker | resnet| 2 g4dn 4 gpus| 36.62 | 2.14 | 0.95 | 0.94|
| multi-worker | efficientnet| 2 g4dn 4 gpus| 43.17 | 0.95 | 0.35 | 0.33|

Ill be looking at getting more models like bert and albert to add to this list.

yuefengz: Does step 1 here include gpu initialization, variable creation and dataset creation time? I think you can just measure the time of function tracing to measure cost of rebuilding graphs.

sboshin: This is using the TimeHistory keras callback with log_steps=1. 
Which only records on "on_train_batch_begin"
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/engine/training.py#L1089

So GPU, and dataset should be created by now. 
But this should include variable creation, as that is part of the model creation. 
If we had to rebuild the session, won't we have to recreate variables?

yuefengz: Whether to build your variables depends on how you implement you solution I think. But I think it is simpler to rebuild variables. According to how you instructment it, it seems to me that your step 0 contains dataset creation/initialization time as well, which could be quite expensive. Rebuilding the dataset in case of worker pool change is necessary if you want to reshard your dataset.

Here we are talking about the necessity to add `group_size` and `group_key` to collective ops. In my opinion, simply recreating the graph should be good enough. So we can just see how expensive it is to recreate the graph.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
sboshin: So this was actually a design that was considered. Using an "orchestrator". If the orchestrator fails, you will lose all elasticity and possibly fault tolerance. Another reason why I didn't pursue this was it will change the idea of MWMS. But it seems because adding elasticity is a huge feature, this will be encapsulated in its own strategy, So this can be introduced again.

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
yuefengz: Anyway, changing the collective ops is acceptable if rebuilding the graph turns to be really expensive. My point was that you can start building a quick prototype on top of the existing MWM. This quick prototype won't require changing collective ops if you rebuild the graphs. This prototype would allow us to know better about the potential problems and necessary changes. This also allows potential users to try it out earlier.

sboshin: I understand, thanks!

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
sboshin: Adding some more data. 
Variable initialization is throughout the whole process This is Dataset and Function Tracing.
From my understanding of eager, each function is traced twice, one with Variable creation, and another without.
Any more models i should get data for?
| Strategy | Model | setup | Function Tracing (sec) | Variable Initialization (sec) | Dataset Initialization (sec) |
|----------|:-----:|:------:|:------:|:------:|:------:|
| multi-worker | efficientnet| 2 g4dn 4 gpus| 36.54 | 6.77 | 6.06 |
| multi-worker | resnet| 2 g4dn 4 gpus| 20.35 | 4.89 | 1.43 |
| multi-worker | transformer| 2 g4dn 4 gpus| 38.08 | 14.07 | 0.08 |
| multi-worker | bert| 2 g4dn 4 gpus| 55.32 | 24.5 | 0.28 |
| multi-worker | albert| 2 g4dn 4 gpus| 38.08 | 14.07 | 0.08 |

rfcs/20200525-elasticity-and-fault-tolerance-in-multiworkermirroredstrategy.md
sboshin: @yuefengz @byronyi  So i have been working on a prototype, Its very raw and its just there to show possible functionality. 
Currently there are two aspects that are working. 
1. Adding a new worker: To add a new worker, you can partially reinitialize the strategy by calling the enable_collective_ops function in the context. Then modify all functions that are loaded in the context with the new group size. (This is only for autographed functions). All ops created in functions should then have the same collective keys, and since they are updated it will have the correct group size. 
2. Resharding Iterators: This involves creating a new API for an elastic iterator. We need to update the iterator internally with a new dataset. This is accomplished by creating a new Operator, this operator saves the old multi-device iterator, creates a new multi-device iterator with a new dataset, then restores the iterator. The iterator restore doesn't change the dataset, only the iterator members. 

The example working right now is in the context of Keras. 
Took the Mnist examples in tensorflow/models, run the model for a long epoch time, and add a new node mid-epoch. This works. Next step is to validate the batch size in the keras context, and update the batch size so the keras logging updates the ETA. 

Are there any other specifics needed, or should i update the RFC with the new findings?

sboshin:
Spoke with Yuefeng, and this will be closed for now.

PR #250, opened 2020-05-28
Updated "optional arguments" section.

dynamicwebpaige:
As discussed in the TF APIs Owners meeting on 27 May.

### DIFF
--- a/governance/api-reviews.md
+++ b/governance/api-reviews.md
@@ -151,13 +151,18 @@ imperfections into new APIs for the sake of consistency with old APIs.
 
 ### Optional arguments with default values
 
-Many APIs have optional arguments with a default value. Our recommendation is to
-use `None` as the default value of any optional arguments and have the
-implementation be responsible for handling it as opposed to using a default
-value that directly represents the behavior (e.g. `aggregate='sum'`).  The
-latter prevents the implementation from distinguishing between the caller not
-setting the argument vs. the caller setting the argument to the default value,
-which may be needed when the default behavior is changing.
+Many APIs have optional arguments that assign a default value (for example, `activation = None`
+in [`tf.keras.layers.Dense`](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dense)).
+
+Our recommendation is to use `None` as the default value for _any optional arguments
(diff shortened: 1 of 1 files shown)


governance/api-reviews.md
```diff
@@ -151,13 +151,18 @@ imperfections into new APIs for the sake of consistency with old APIs.
...
+
+Our recommendation is to use `None` as the default value for _any optional arguments
+that may be adjusted or changed over time_, and have the implementation be responsible
+for handling the value, as opposed to using a default value that directly represents 
+the behavior (e.g. `aggregate='sum'`).  The latter prevents the implementation from
+distinguishing between the caller not setting the argument vs. the caller setting the
+argument to the default value, which may be needed when the default behavior is changing.
+
+If the optional argument is _backwards incompatible to change_, however, its default should
+reflect the actual default value when possible.
```
karmel, line 165: Can we include @fchollet 's Milk example as well? I rather like that as a way to differentiate the two. In code would be ideal, so we can show what we mean by letting the implementation set the value.

alextp: The activation function or aggregation is a good example of something backwards incompatible and should be here

governance/api-reviews.md
```diff
@@ -151,13 +151,18 @@ imperfections into new APIs for the sake of consistency with old APIs.
...
-latter prevents the implementation from distinguishing between the caller not
-setting the argument vs. the caller setting the argument to the default value,
-which may be needed when the default behavior is changing.
+Many APIs have optional arguments that assign a default value (for example, `activation = None`
+in [`tf.keras.layers.Dense`](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Dense)).
+
+Our recommendation is to use `None` as the default value for _any optional arguments
+that may be adjusted or changed over time_, and have the implementation be responsible
+for handling the value, as opposed to using a default value that directly represents 
+the behavior (e.g. `aggregate='sum'`).  The latter prevents the implementation from
```
alextp, line 160: Maybe mention num_threads=10 instead of aggregate=sum? That's more in keeping with stuff that can obviously change (changing the aggregation is incompatible)

dynamicwebpaige: Changed!

Issue #242, opened 2020-05-07
Tensorflow to Sig downstreaming


bhack:
As we are already improving the upstreaming/promotion process with RFC template at https://github.com/tensorflow/community/pull/241 it could be useful to have something similar for downstreaming so that we could cover cases like https://github.com/tensorflow/addons/issues/1781

bhack:
I think that we could also add an extra bullet point in https://github.com/tensorflow/community/blob/master/governance/api-reviews.md#experimental-apis before the deprecation option.

bhack:
https://github.com/tensorflow/addons/pull/1944

bhack:
@seanpmorgan I suppose that we can close this a we are not handling this anymore. We just accept downstreaming Issue/PR from TF teams

PR #260, opened 2020-06-17
RFC: Multihead Attention and EinsumDense on Keras

saberkun:
# RFC: Multihead Attention and EinsumDense on Keras

| Status        | Proposed         |
| **RFC #**     | [260](https://github.com/tensorflow/community/pull/260) |
| **Author(s)** | Hongkun Yu (hongkuny@google.com), Mark Omernick (momernick@google.com)   |
| **Sponsor**   | Francois Chollet (fchollet@google.com)                  |
| **Updated**   | 2020-06-16                                              |

## Objective

Introduce the MultiHeadAttention layer and EinsumDense layer to tf.keras.

### DIFF
--- a/rfcs/20200616-keras-multihead-attention.md
+++ b/rfcs/20200616-keras-multihead-attention.md
@@ -302,7 +302,7 @@ No dependencies.
 
 *   We will first introduce the layer as
     `tf.keras.layers.experimental.MultiHeadAttention` and
-    `tf.keras.layers.experimental.MaskedSoftmax`.
+    `tf.keras.layers.experimental.EinsumDense`.
 
 ## Detailed Design
 
--- a/rfcs/20200616-keras-multihead-attention.md
+++ b/rfcs/20200616-keras-multihead-attention.md
@@ -134,7 +134,7 @@ class MultiHeadAttention(tf.keras.layers.Layer):
   >>> target = tf.keras.Input(shape=[8, 16])
   >>> source = tf.keras.Input(shape=[4, 16])
   >>> mask_tensor = tf.keras.Input(shape=[8, 4])
-  >>> output_tensor, weights = layer([target, source])
+  >>> output_tensor, weights = layer([target, source], attention_mask=mask_tensor)
   >>> print(output_tensor.shape), print(weights.shape)
   (None, 8, 16)  (None, 2, 8, 4)
 
--- a/rfcs/20200616-keras-multihead-attention.md
+++ b/rfcs/20200616-keras-multihead-attention.md
@@ -1,6 +1,6 @@
 # RFC: Multihead Attention and EinsumDense on Keras
 
-| Status        | Proposed         |
+| Status        | Accepted         |
 | :------------ | :------------------------------------------------------ |
 | **RFC #**     | [260](https://github.com/tensorflow/community/pull/260) |
 | **Author(s)** | Hongkun Yu (hongkuny@google.com), Mark Omernick (momernick@google.com) |
@@ -342,8 +342,8 @@ method. For example, we implemented
 [TalkingHeadAttention](https://github.com/tensorflow/models/blob/master/official/nlp/modeling/layers/talking_heads_attention.py)
 introduced by ["Talking-Heads Attention "](https://arxiv.org/abs/2003.02436)
 paper. Using the keras Attention layer as another example, since it supports the
-basic single-head case 1-D attention, we can use it inside `_build_attention`
-and `_compute_attention`.
+basic single-head case 1-D attention, we can use it inside `build_attention`
+and `compute_attention`.
 
 ## Questions and Discussion Topics
 
--- a/rfcs/20200616-keras-multihead-attention.md
+++ b/rfcs/20200616-keras-multihead-attention.md
@@ -3,7 +3,7 @@
 | Status        | (Proposed / Accepted / Implemented / Obsolete)          |
 | :------------ | :------------------------------------------------------ |
 | **RFC #**     | [260](https://github.com/tensorflow/community/pull/260) |
-| **Author(s)** | Hongkun Yu (hongkuny@google.com), Mark Omernick (momernick@google.com)    |
+| **Author(s)** | Hongkun Yu (hongkuny@google.com), Mark Omernick (momernick@google.com) |
 | **Sponsor**   | Francois Chollet (fchollet@google.com)                  |
 | **Updated**   | 2020-06-16                                              |
 
@@ -83,7 +83,7 @@ test_layer = MultiHeadAttention(
     num_heads=2, key_size=2, return_attention_scores=True)
 target = np.array([[[0.1, 0.2], [0.0, 0.0]]])
 source = np.array([[[0.1, 0.2], [3.0, 1.0]]])
-output, scores = test_layer([target, source])
+output, scores = test_layer(query=target, value=source)
 scores = tf.math.reduce_sum(scores, axis=1) # shape = (1, 2, 2)
 ```
 
@@ -96,7 +96,7 @@ mask_shape = [2, 3, 4, 3, 2]
 query = 10 * np.random.random_sample(query_shape)
 value = 10 * np.random.random_sample(value_shape)
 mask_data = np.random.randint(2, size=mask_shape).astype("bool")
-output = test_layer([query, value], mask_data)
+output = test_layer(query=query, value=value, attention_mask=mask_data)
 ```
 
 ### Interface
@@ -134,7 +134,8 @@ class MultiHeadAttention(tf.keras.layers.Layer):
   >>> target = tf.keras.Input(shape=[8, 16])
   >>> source = tf.keras.Input(shape=[4, 16])
   >>> mask_tensor = tf.keras.Input(shape=[8, 4])
-  >>> output_tensor, weights = layer([target, source], attention_mask=mask_tensor)
+  >>> output_tensor, weights = layer(query=target, value=source
+  ...                                attention_mask=mask_tensor)
   >>> print(output_tensor.shape), print(weights.shape)
   (None, 8, 16)  (None, 2, 8, 4)
 
@@ -142,7 +143,7 @@ class MultiHeadAttention(tf.keras.layers.Layer):
 
   >>> layer = MultiHeadAttention(num_heads=2, key_size=2, attention_axes=(2, 3))
   >>> input_tensor = tf.keras.Input(shape=[5, 3, 4, 16])
-  >>> output_tensor = layer([input_tensor, input_tensor])
+  >>> output_tensor = layer(query=input_tensor, value=input_tensor)
   >>> print(output_tensor.shape)
   (None, 5, 3, 4, 16)
 
@@ -167,7 +168,7 @@ class MultiHeadAttention(tf.keras.layers.Layer):
     bias_constraint: Constraint for dense layer kernels.
   """
 
-  def call(self, inputs, attention_mask=None):
+  def call(self, query, value, key=None, attention_mask=None):
     """Implements the forward pass.
 
     Size glossary:
@@ -180,10 +181,9 @@ class MultiHeadAttention(tf.keras.layers.Layer):
       * Value (source) attention axes shape (S), the rank must match the target.
 
     Args:
-      inputs: List of the following tensors:
-        * query: Query `Tensor` of shape `[B, T, dim]`.
-        * value: Value `Tensor` of shape `[B, S, dim]`.
-        * key: Optional key `Tensor` of shape `[B, S, dim]`. If not given, will
+      query: Query `Tensor` of shape `[B, T, dim]`.
+      value: Value `Tensor` of shape `[B, S, dim]`.
+      key: Optional key `Tensor` of shape `[B, S, dim]`. If not given, will
           use `value` for both `key` and `value`, which is the most common case.
       attention_mask: a boolean mask of shape `[B, T, S]`, that prevents
         attention to certain positions.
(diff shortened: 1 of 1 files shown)


rfcs/20200616-keras-multihead-attention.md
```diff
@@ -0,0 +1,340 @@
...
+
+### Key Features
+
+*   Returns multi-headed attention scores, which is commonly useful for
+    attention visualization and analysis.
+*   Supports query (Q), key (K), value (V) tensors as individual inputs and
+    supports projecting Q, K, V to different dimensions.
+*   Final outputs projects to user specified dimensions.
+*   Using tf.einsum to express high-dimensional computation and adopts
+    [tf.keras.layers.experimental.EinsumDense](https://www.tensorflow.org/api_docs/python/tf/keras/layers/experimental/EinsumDense)
```
terrytangyuan, line 38: What's the relationship between `tfa.layers` and `tf.keras.layers.experimental`? cc @seanpmorgan @karmel @bhack

terrytangyuan: In particular, relation to the existing [tfa.layers.MultiHeadAttention](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/layers/multihead_attention.py).

rfcs/20200616-keras-multihead-attention.md
```diff
@@ -0,0 +1,340 @@
...
+*   Attention caching for decoding. Implemented in
+    [Flax](https://github.com/google/flax/blob/master/flax/nn/attention.py#L301).
+    The caching is a special treatment for inference and we noticied that
+    different treatments are required for dynamic or static shape programs.
+    Thus, subclassing as a
+    [CachedAttention](https://github.com/tensorflow/models/blob/master/official/nlp/modeling/layers/attention.py)
+    layer is the solution inside the model garden.
+*   [MultiHeadAttention](https://github.com/tensorflow/addons/blob/master/tensorflow_addons/layers/multihead_attention.py)
+    keras layer is also implemented in TF-Addons. The design in this doc covers
+    the features in TF-addons implementation but generalizes to more use cases.
```
seanpmorgan, line 260: So this was just released in TFA [~1month ago](https://github.com/tensorflow/addons/releases/tag/v0.9.0). How can we improve the communication so there is less wasted effort by contributors and reviewers?

Is there a future roadmap so we can anticipate features that may be superceeded?

rfcs/20200616-keras-multihead-attention.md
seanpmorgan: IIUC layers that go in `tf.keras.layers.experimental` are slated to land in keras, but the API is not set in stone.

Addons would be a place for contributions whos broad applicability is not yet clear, or it is mostly used by a smaller subset of the community (Per charter). This gets tricky though because MultiHeadAttention has proven its applicability but there was no installable implementation in the TF ecosystem. Perhaps we should bring all situations like this up to the Keras team beforehand? That adds another subjective step though so a roadmap would be preferable.

saberkun: The addons implementation is a very good reference. We incorporate the features inside the  addons version and generalize more to fit the emerging needs. For this design, it should cover the tfa.layers.MultiHeadAttention. We hope this common layer can be inside tf.keras directly. The implementation started in model garden back to last year.

rfcs/20200616-keras-multihead-attention.md
saberkun: Hi Sean,
Added a section: "TFA MultiHeadAttention Deprecation and Re-mapping". The proposed implementation should cover TFA use cases. Please review and suggest how we should help TFA consolidate. There are minor differences on argument names and mask shape checking. How would be a common procedure to re-map a symbol in TFA? We can work on it once the layers and changes are finished in tf.keras. Thank you

bhack:
I don't understand here what Is the plan for Model gardens official models.
I have a strong feeling that at some point we will have the model gardens official models own implementation, TFA implementation and this one that will be mainly going to be used to develop [keras-nlp](https://github.com/keras-team/keras-nlp).

rfcs/20200616-keras-multihead-attention.md
bhack: `.experimental` was defined and approved in https://github.com/tensorflow/community/blob/master/governance/api-reviews.md#experimental-apis.
Sometimes when an experimental namespace doesn't get too much traction it could has an opportunity to be [downstreamed to tf.addons](https://github.com/tensorflow/community/issues/242) as an alternative to jubbe removed (but this downstreaming step is not defined in the RFC).

saberkun: For the MultiHeadAttention layer, it is relatively clear that we should put it inside core keras finally.

saberkun:
> I don't understand here what Is the plan for garden official models.
> I have a strong feeling that at some point we will have the model gardens official models own implementation, TFA implementation and this one that will be mainly going to be used to develop [keras-nlp](https://github.com/keras-team/keras-nlp).

Hi @bhack, the model garden implementation of the MultiHeadAttention will be moved to tf.keras.experimental in theory in the same commit. We have done the migration for EinsumDense. The model garden does not keep duplicated implementation whenever there is a reliable source with full features needed. @rachellj218

rfcs/20200616-keras-multihead-attention.md
```diff
@@ -0,0 +1,353 @@
...
+  Examples:
+
+  Performs 1D cross-attention over two sequence inputs with an attention mask.
+  Returns the additional attention weights over heads.
+
+  >>> layer = MultiHeadAttention(num_heads=2, key_size=2,
+  ...                            return_attention_scores=True)
+  >>> target = tf.keras.Input(shape=[8, 16])
+  >>> source = tf.keras.Input(shape=[4, 16])
+  >>> mask_tensor = tf.keras.Input(shape=[8, 4])
```
qlzh727, line 136: The mask_tensor seems not used in this code block.

rfcs/20200616-keras-multihead-attention.md
```diff
@@ -0,0 +1,353 @@
...
+  >>> output_tensor, weights = layer([target, source])
```
qlzh727, line 137: @tomerk, since we started the effort of treat all tensor input to be the same and not specialize for the first call arg, should we encourage user to not pass in list of tensor, and have individual kwargs for tensor inputs? like call(query, key, value=None, mask=None)?

I think the explicit kwargs will be more readable, since the plain list doesn't carry any information of the individual tensor.

rfcs/20200616-keras-multihead-attention.md
```diff
@@ -0,0 +1,353 @@
...
+    Garden applications.
+
+### User Impact
+
+*   We will first introduce the layer as
+    `tf.keras.layers.experimental.MultiHeadAttention` and
+    `tf.keras.layers.experimental.EinsumDense`. When the APIs are stable and
+    functionalities are fully verified, the next step is to
+    graduate as core keras layers by removing `experimental` scope.
+
```
qlzh727, line 317: @fchollet, for the existing keras.layers.Attention and keras.layers.AddictiveAttention, I think we should add some clarification to not confuse user between them and the new multihead attention. I could expect most of the user will use multihead attention.

rfcs/20200616-keras-multihead-attention.md
saberkun: I am ok with that as Tom has fixed that. But yeah, I think we should test real models with mix precision policy.
If we use individual kwargs, the `Attention` layer interface may need to be consistent, which takes a list. Using inputs as a list is intentionally to be consistent with `Attention` and TFA implementation. Do we want to change that?

rfcs/20200616-keras-multihead-attention.md
saberkun: Fixed. Thanks

rfcs/20200616-keras-multihead-attention.md
qlzh727: I think the first input tensor was specially handled before Tomer's change, which means you had to use list of tensors if there were multiple inputs. Now you can have individual tensors as kwargs. I think the kwargs will be more readable, and could set as an example for user when there are multiple inputs. @fchollet and @tomerk, WDYT?

AakashKumarNain: IMO `kwargs` is much better not only for readability but it also removes the burden on the user of checking the consistency of positional arguments passed

saberkun: Sounds good. I also like kwargs. Will add this to proposal and discuss.
Will also test it on our end.

rfcs/20200616-keras-multihead-attention.md
```diff
@@ -0,0 +1,353 @@
...
+layer call method takes an optional argument, `mask`, which requires two
+tensors, `q_mask` and `v_mask`. They are following keras framework requirements
+with (batch_size, target_length) and (batch_size, source_length) as shapes. This
+limits the flexibility of masking and `MultiHeadAttention` layer generalize the
+attention mask to be (batch dims, target dims, source dims). To be consistent,
+we would like to introduce an optional argument `attention_mask` for
+`tf.keras.layers.Attention`. In the reduced case of `tf.keras.layers.Attention`,
+the shape is (batch_size, target_length, source_length). Whenever
+`attention_mask` is specified, the `mask` argument is OK to be skipped.
+
```
karmel, line 244: I find this confusing. Can you be more explicit about the differences between the two attention layers? If this is a generalization, should there be an inheritance relationship here? If MultiHead is a generalization, why not just update the existing Attention layer and update to handle this case? We will need to have clear guidance for users as to when to use the one versus the other.

rfcs/20200616-keras-multihead-attention.md
```diff
@@ -0,0 +1,353 @@
...
+    
+    
+
+## Detailed Design
+
+The layer has been implemented as the
+[MultiHeadAttention](https://github.com/tensorflow/models/blob/master/official/nlp/modeling/layers/attention.py#L116)
+inside TensorFlow Model Garden.
+
+First, as we rely on `tf.eisum` to define projections and attention computation,
```
karmel, line 326: nit:einsum

rfcs/20200616-keras-multihead-attention.md
```diff
@@ -0,0 +1,353 @@
...
+batch dimensions and attention softmax can be performed on multiple axes, we
+need to track the batch axes and attention axes inside einsum notations. We use
+a vector of chars and use two local methods to generate einsum notations for
+projections and attentions.
+
+Second, the layer by default implements the most common dot-product attention.
+There are various ways to implement the attention computation, so we modulize it
+as two methods `_build_attention` and `_compute_attention`. Thus, users may be
+able to just override them to get a new keras layer with a novel attention
+method. For example, we implemented
```
karmel, line 338: If these are underscored, they are not backwards-compatibility-guaranteed, and we should not encourage users to overwrite. If we want these to be overwritable, we should make them public methods.

rfcs/20200616-keras-multihead-attention.md
saberkun: The MultiHeadAttention layer contains projection layers to Q, K, V inputs and outputs. The attention computation is multi-headed dot-product attention. The proposal follows the same scope as other ML libraries, please checkout. The module is commonly used in NLP and new Vision research.
The current Keras Attention layer is the attention computation for single-head and not using einsum. It can be used to implement the attention computation part of MultiHeadAttention.

rfcs/20200616-keras-multihead-attention.md
saberkun: Updated.

rfcs/20200616-keras-multihead-attention.md
saberkun: We will clarify the Attention and AddictiveAttention are really for attention computation. They are subsets of MultiHeadAttention commonly used in NLP/transformer-based vision models which shares the same scope of other ML libraries.

rfcs/20200616-keras-multihead-attention.md
bhack: Is `tf.keras.layers.Attention` design flexible enough to support the research proliferation on dense attention alternatives?

saberkun: bhack has good comment. We will meet with the team and check the plan with keras.Attention layer. 
For NLP research, people may prefer direct access to attention computation with ops.

saberkun:
Notes from the review meeting
Q1:  Is TF Addons team ok with the RFC?
A: Yes, they are on board. Sean from TF Addons is ok with this. We will send a PR to TFA to deprecate or rewire to this new layer after it goes to tf nightly.
We will keep engaging with the TF Addons community. Consider giving them early updates.

Q2: The relation between the existing Attention layer and MultiHeadAttention. Is it possible to make Attention be part of MultiHeadAttention? or make MultiHeadAttention inherit from Attention?
A: The existing Attention layer is only for the attention computation. The Attention could take the inputs after the projections inside MHAttention and then do the computation. However, it doesn’t fully support computation needed in MHAttention. Also it doesn’t use einsum or support high-dimensional attention. To make it usable in MHAttention, we need to rewrite the attention layer. MHAttention contains more logics for transformer architecture, not a superset of Attention. It’s hard to deprecate Attention. 
Will articulate clearly in RFC that the difference between attention and multi headed attention, with docstrings and code examples. Users will be able to use Attention/AdditiveAttention inside the `compute_attention` method once we add the `attention_mask` argument.

Q3 Do we agree to use Keyword arguments for tensor inputs?
A: yes, we should use this for better usability. This will only work with the latest TF version though.
Follow with keras team to make sure keyword-argument layers are treated as the first-class citizen in the TF ecosystem.

ematejska:
@saberkun  Are we ready to merge or did you want to check anything else before I push all the merge buttons?

saberkun:
> @saberkun Are we ready to merge or did you want to check anything else before I push all the merge buttons?

Yes, we are ready to merge. I have chat with the team to follow up with AIs after the review. It is clear now. Thanks

PR #266, opened 2020-07-13
RFC: TFRT Kernel Fallback

annarev:
| Status        | (Proposed / Accepted / Implemented / Obsolete)          |
| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) |
| **Author(s)** | Anna Revinskaya (annarev@google.com), Jeremy Lau (lauj@google.com) |
| **Sponsor**   | Jeremy Lau (lauj@google.com)                            |
| **Updated**   | 2020-07-12

## Objective

This proposal focuses on getting majority of “well-behaved” ops running in
[TF Lite](https://www.tensorflow.org/lite) by skipping current eager runtime and
calling kernels directly in [TFRT](https://github.com/tensorflow/runtime) (a new
TensorFlow runtime).

Note that there is an effort to call existing kernels by delegating to
TensorFlow eager runtime instead. This approach is called Runtime Fallback and
corresponding RFC will be published soon. The goals of the two fallback
mechanisms are as follows:

*   Runtime Fallback aims to reuse all current TensorFlow kernels in TFRT.
*   Kernel Fallback (focus of this document) aims to get a large number of
    existing kernels working in TFRT while reducing binary size to support
    mobile devices.


rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,641 @@
+# Title of RFC
+
+| Status        | (Proposed / Accepted / Implemented / Obsolete)          |
+| :------------ | :------------------------------------------------------ |
+| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) |
+| **Author(s)** | Anna Revinskaya (annarev@google.com), Jeremy Lau        |
+:               : (lauj@google.com)                                       :
```
fdxmw, line 7: This syntax does not seem supported by github's markdown renderer

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,641 @@
...
+Corresponding .cc file then registers the kernel using the correct kernel and
+context classes. For example, this is how we register `AddN` kernel with TFRT:
+
+```cpp
+REGISTER_KERNEL_FALLBACK_KERNEL( "AddN", AddNOp<CPUDevice, int32>);
+```
+
+## Calling kernel
+
+We add a new TFRT BEF kernel called `tfd.kernel_fallback`. This kernel directly
```
zhangqiaorjc, line 292: Might be good to rename this legacy dialect name. It's no longer "TF Delegate" dialect.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,641 @@
...
+    `float`).
+*   Inputs and outputs have type `tensorflow::Tensor`. We will provide BEF
+    kernels to construct these from BEF data (for e.g. constant values).
+
+Example of invoking Conv3D kernel:
+
+```
+%tft_c = "tfd.kernel_fallback"(%tft_a, %tft_b) {
+    _op_name = "Conv3D", attr1_name="data_format",
+    attr1_value="string$NDHWC", attr2_name="strides",
```
zhangqiaorjc, line 336: Might be good to mention that TFRT orders attributes by name, alphabetically, and thus the decision to do attrN_value and attrN_name pair.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Thank you for catching! Fixed. Thats my editor being eager to autowrap and format.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Should I just change it to tf. or tff for (tf fallback)?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,641 @@
...
+    pointers on the C API surface (they get cast to `TF_OpKernelContext` and
+    `TF_OpKernelConstruction` aliases).
+1.  Most of the functions just provide accessors into
+    `OpKernelContext`/`OpKernelConstruction` types.
+
+Given current API structure, we can consider two approaches going forward:
+
+1.  TFRT fully supports all functionality available in the C API. This way any
+    kernel defined using the C API would be automatically available using either
+    full TensorFlow or the TFRT-to-TF forwarding delegate.
```
fdxmw, line 391: s/the TFRT-to-TF forwarding delegate/TFRT kernel fallback/ ?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,640 @@
...
+`OpKernelBase` instead of `OpKernel`. This lets us remove runtime-specific
+information from kernel subclasses and lets us support both current and new
+TensorFlow runtime.
+
+However, TensorFlow runtime assumes that kernel extend `OpKernel` and support
+all of its functionality. In other words we want kernels to extend
+`OpKernelBase` but be added to existing TensorFlow registry as `OpKernel`
+objects.
+
+It seems easiest to me to wrap OpKernelBase some class that extends OpKernel (I
```
fdxmw, line 591: "with some class" ?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,640 @@
...
+`OpKernelContextInterface` in terms of TFRT data structures. Example how
+`TFRTOpKernelConstruction` might look like:
+
+```cpp
+class TFRTOpKernelConstruction : public OpKernelConstructionInterface {
+ public:
+  explicit TFRTOpKernelConstruction(AttrMap attributes);
+  ~TFRTOpKernelConstruction() override {};
+
+  Status GetAttr(StringPiece attr_name, int32* value) const override;
```
aselle, line 239: Consider a mechanism that allows delegated string interning to allow these attribute matches to ammortized by every invocation rather than multiplied by every invocation.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,641 @@
...
+| **Author(s)** | Anna Revinskaya (annarev@google.com), Jeremy Lau        |
+:               : (lauj@google.com)                                       :
+| **Sponsor**   | Jeremy Lau (lauj@google.com)                            |
+| **Updated**   | 2020-07-06                                              |
+
+## Objective
+
+This proposal focuses on getting majority of “well-behaved” ops running in
+[TF Lite](https://www.tensorflow.org/lite) by skipping current eager runtime and
+calling kernels directly in [TFRT](https://github.com/tensorflow/runtime) (a new
```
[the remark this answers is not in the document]
aselle: I would suggest a diagram that shows what this change really looks like in system architecture. It will also elucidate "directly" means. Show a box for the eager api, the opkernelcontext, etc.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Done

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Done.

rfcs/20200712-tfrt-kernel-fallback.md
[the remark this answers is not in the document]
annarev: Good point. updated

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,641 @@
...
+
+*   100-200k overhead to call TF kernels
+*   20k / kernel marginal size
+
+To get closer to the size restrictions we plan to define a call path from TFRT
+to TensorFlow kernels that minimizes the amount of generated code.
+
+## User Benefit
+
+Running more kernels on mobile devices would allow TensorFlow users to implement
```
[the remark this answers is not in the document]
annarev: Updated.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,641 @@
...
+custom
+[OpHandler](https://github.com/tensorflow/runtime/blob/3c7a1ea02c87325f1b47aebb24b3ca6e84e7e7e7/include/tfrt/core_runtime/op_handler.h#L47)).
+
+## Non-goals
+
+*   Supporting all existing ops. `OpKernelContext` surface is quite large and
+    implementing all of it would require a significant amount of time. Instead,
+    we will start by adding most common and easy functionality. If certain
+    functionality is only used by a handful of kernels, it might make sense to
+    implement TFRT native kernels instead. One notable example is
```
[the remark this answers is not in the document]
annarev: Makes sense, updated.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,641 @@
...
+*   Proposed implementation order for these methods:
+    [spreadsheet](https://docs.google.com/spreadsheets/d/10u6tcTE9PAi45A04nxSz61whSnwhscRLlSUNRPJugIY/edit?usp=sharing)
+
+Based on these estimates, we can support >= 423 kernels. Note that this number
+is just based on the `OpKernelConstruction`/`OpKernelContext` coverage that we
+can provide. It doesn't take into consideration other issues we might face.
+
+### TFRT Integration Goals
+
+We want to support executing a [BEF](https://github.com/tensorflow/runtime) file
```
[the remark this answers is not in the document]
annarev: Nice! That looks like a much better link to include :)

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Good suggestion, thank you!

rfcs/20200712-tfrt-kernel-fallback.md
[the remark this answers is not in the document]
annarev: Good suggestion to add diagrams. I minimal side-by-side diagrams for Kernel Fallback and Runtime Fallback.
I also included a more detailed diagram under Design Proposal.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Changed to tfrt_fallback.kernel_fallback.

bhack:
Can you introduce why we are talking about fallback? The no fallback are codegen? What Is the native kernel concept in TFRT?
Is there an impact on custom c++ ops in SIGs repositories? There will be a way to run that on mobile?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,641 @@
...
+and
+[OpKernelContext](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/framework/op_kernel.h;l=584?q=OpKernelContext&ss=tensorflow%2Ftensorflow)
+classes. Number of supported kernels will depend on the surface we manage to
+customize. (Note that I have already started prototyping the implementation that
+includes a few common methods such as `input`, `output`. The spreadsheet below
+consideres these methods to be already *supported*).
+
+*   List of kernels and `OpKernelConstruction`/`OpKernelContext` methods they
+    require:
+    [spreadsheet](https://docs.google.com/spreadsheets/d/18bOu2gJQnZtCRGPZ4yerEAKUHgp1V429dPdCuzoCSkU/edit?usp=sharing)
```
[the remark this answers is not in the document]
annarev: I added supplemental markdown docs with the tables instead. Thank you for reminder!

rfcs/20200712-tfrt-kernel-fallback.md
annarev: We currently lookup attributes in TFRTs [OpAttrs](https://cs.opensource.google/tensorflow/runtime/+/master:include/tfrt/core_runtime/op_attrs.h;l=114?q=OpAttrs).
Depending how kernel fallback will be called from TFRT, `OpAttrs` might be passed to us as an argument and we save time by not creating a different data structure. That being said, these `OpAttrs` would only contain explicitly set attributes. We still need inferred attribute values.

So, may be a different data structure would make sense in the future and we can support string interning there.

annarev:
> Can you introduce why we are talking about fallback? The no fallback are codegen? What Is the native kernel concept in TFRT?

"Native kernels" are kernels implemented in the new runtime (as opposed to existing TensorFlow runtime). See: https://github.com/tensorflow/runtime/blob/62695daf59a51774709fac081165f36611dfe67c/documents/tfrt_host_runtime_design.md#low-level-abstractions-asyncvalue-kernel-threadpool-allocator . @fdxmw let me know if there is anything I am missing.

Since TFRT is a new runtime, it only supports a small subset of kernels. At the same time, existing TensorFlow runtime supports many kernels (probably around 1000). We want to define a way for TFRT to call these existing TensorFlow kernels and we call it "fallback".

> Is there an impact on custom c++ ops in SIGs repositories?

Impact is optional. We are thinking of having two types of "fallback": Runtime fallback and Kernel fallback. Runtime fallback should support everything out of the box, including custom C++ ops. Kernel fallback aims to support a subset of kernels with a subset of functionality. Custom ops can choose to support kernel fallback by extending `OpKernelBase` instead of `OpKernel` and taking `OpKernelConstructionInterface`/`OpKernelContextInterface` instead of `OpKernelConstruction/`OpKernelContext` as inputs. Depending on the kernel, this might or might not be feasible.

> There will be a way to run that on mobile?

Main concern is binary size. It might be that Runtime fallback can run on mobile as well - with larger size demands, which would automatically support all kernels. At the same time, the focus of this doc - Kernel fallback aims to support a subset kernels but minimize binary size. So it would be a better option for models that just use this subset of kernels.

bhack:
> Since TFRT is a new runtime, it only supports a small subset of kernels. At the same time, existing TensorFlow runtime supports many kernels (probably around 1000). We want to define a way for TFRT to call these existing TensorFlow kernels and we call it "fallback".

@annarev So is this just a more or less long temporary solution waiting for native kernel population in TFRT?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+findings in the following spreadsheets. As we describe in
+[Design Proposal](#design-proposal) below, Kernel Fallback depends on
+customizing
+[OpKernelConstruction](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/framework/op_kernel.h;l=256?q=OpKernelConstruction)
+and
+[OpKernelContext](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/framework/op_kernel.h;l=584?q=OpKernelContext&ss=tensorflow%2Ftensorflow)
+classes. Number of supported kernels will depend on the surface we manage to
+customize. (Note that I have already started prototyping the implementation that
+includes a few common methods such as `input`, `output`. The spreadsheet below
+consideres these methods to be already *supported*).
```
mhong, line 59: considers

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+
+*   Supporting all existing ops. `OpKernelContext` surface is quite large and
+    implementing all of it would require a significant amount of time. Instead,
+    we will start by adding most common and easy functionality. If certain
+    functionality is only used by a handful of kernels, it might make sense to
+    implement TFRT native kernels or rely on runtime fallback instead. One notable example is
+    [ResourceMgr](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/framework/resource_mgr.h;l=152?q=ResourceMgr).
+    We might support it later, but it is definitely not first priority due to
+    extra effort required.
+*   Gradients would not be supported by the first iteration of Kernel Fallback,
```
mhong, line 92: can you elaborate on what this means? IIUC gradient kernels are just like other normal kernels, but there is extra metadata that associate them with the forward-pass ops/kernels, which the fallback mechanism need not care.

So I imagine there could be gradient kernels that are also "well behaved" for kernel fallback purposes.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+[limited set of ops](https://www.tensorflow.org/lite/guide/ops_compatibility).
+As the range and variety of applications grows, it becomes essential to grow the
+pool of available ops as well, ideally supporting everything that fully-fledged
+TensorFlow supports now.
+
+However, supporting TensorFlow ops on mobile devices presents some challenges.
+Specifically, binary size on mobile platforms should be restricted. TF Lite team
+provided us with the following *ideal* numbers:
+
+*   100-200k overhead to call TF kernels
```
mhong, line 111: do we have size measurements from the actual impl?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+
+ private:
+  llvm::StringMap<std::vector<TFRTOpKernelReg>> factories_;
+};
+
+extern llvm::ManagedStatic<TFRTOpKernelFactories>
+    tfrt_forwarding_kernel_factories;
+```
+
+Similar to current TensorFlow kernel registartion, we will introduce a
```
mhong, line 168: typo

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+[OpKernelContext](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/framework/op_kernel.h;l=584?q=opkernel).
+We want to provide custom implementations of these two classes in terms of data
+we get from TFRT (for e.g. inputs, attributes).
+
+There are two main approaches to customize class implementations:
+
+*   Use inheritance and define common interfaces.
+*   Use templates.
+
+We ran multiple benchmarks to get an idea of the trade offs between inheritance
```
mhong, line 210: it's hard to follow the discussion here, since we have not described what these 2 approaches do.

one option is to defer the trade-off analysis and proposal here, till after we've described how inheritance works, and also sketched the templates based solution as well.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+case in the real world.
+
+To use inheritance, we will define `OpKernelConstructionInterface` and
+`OpKernelContextInterface` interfaces. Ideally, these interfaces should be pure
+virtual. However, we will have one exception - templated `eigen_device` method
+that calls per-device pure-virtual implementations.
+
+We will then introduce `TFRTOpKernelConstruction` and `TFRTOpKernelContext`
+subclasses that implement `OpKernelConstructionInterface` and
+`OpKernelContextInterface` in terms of TFRT data structures. Example how
```
mhong, line 236: Here's an example of how ...

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+
+  void CtxFailure(const Status& s);
+  void CtxFailureWithWarning(const Status& s);
+  void CtxFailure(const char* file, int line, const Status& s);
+  void CtxFailureWithWarning(const char* file, int line, const Status& s);
+  ...
+};
+```
+
+When forwarding, we instantiate the kernel interfaces with TFRT’s lightweight
```
mhong, line 256: avoid this term throughout the doc, and use "fallback" consistently

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+```
+
+(For details how extending from `OpKernelBase` instead of `OpKernel` would work
+with current TensorFlow runtime see [Appendix 1](#appendix-1))
+
+Corresponding .cc file then registers the kernel using the correct kernel and
+context classes. For example, this is how we register `AddN` kernel with TFRT:
+
+```cpp
+REGISTER_KERNEL_FALLBACK_KERNEL( "AddN", AddNOp<CPUDevice, int32>);
```
mhong, line 293: to confirm, will we be adding these extra symbols to TF, where a CPU kernel like "AddN" for fallback is different from the original CPU kernel for "AddN", or will we be upgrading the existing symbol with fallback capability? (I think it's the latter)

For that CPU kernel, will we still need to use the original REGISTER_KERNEL, or just the new  REGISTER_KERNEL_FALLBACK_KERNEL? (I hope it's the latter)

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+
+extern llvm::ManagedStatic<TFRTOpKernelFactories>
+    tfrt_forwarding_kernel_factories;
+```
+
+Similar to current TensorFlow kernel registartion, we will introduce a
+registration macro that adds a kernel to `TFRTOpKernelFactories`.
+
+```cpp
+#define REGISTER_KERNEL_FALLBACK_KERNEL(name, ...) \
```
mhong, line 172: shall we spell this as REGISTER_FALLBACK_KERNEL for simplicity?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+
+We will be using the following conventions (essentially, these are based on
+Runtime Fallback work that will probably have RFC coming soon):
+
+*   Attributes are passed as key-value pairs, where both key and value are
+    represented as strings.
+*   Types have a specific string representation. We are trying to use names
+    consistent with BEF syntax as much as possible (for e.g. `f32` represents
+    `float`).
+*   Inputs and outputs have type `tensorflow::Tensor`. We will provide BEF
```
mhong, line 334: which ops have this, other than tf.const (which we might want to just implement natively in TFRT)?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+    kernels to construct these from BEF data (for e.g. constant values).
+
+Example of invoking Conv3D kernel:
+
+```
+%tft_c = "tfd.kernel_fallback"(%tft_a, %tft_b) {
```
mhong, line 340: the term "d" stands for delegate, which is now obsolete.

shall we change it to just `tfrt.kernel_fallback`?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+    _op_name = "Conv3D",
+    attr1_name="data_format", attr1_value="string$NDHWC",
+    attr2_name="strides", attr2_value="list(i32)$1,1,1,1,1",
+    attr3_name="dilations", attr3_value="list(i32)$1,1,1,1,1",
+    attr4_name="padding", attr4_value="padding$SAME"}: (!tfd.tensor, !tfd.tensor) -> !tfd.tensor
+```
+
+For example, `dilations` attribute here has a value of `[1, 1, 1, 1, 1]`.
+Note: TFRT orders attributes by name, alphabetically, which is why we use `attrN_value` and `attrN_name` pattern pair.
```
mhong, line 349: @cky9301 @fdxmw Is this true, and is it documented somewhere?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+
+Given current API structure, we can consider two approaches going forward:
+
+1.  TFRT fully supports all functionality available in the C API. This way any
+    kernel defined using the C API would be automatically available using either
+    full TensorFlow or the TFRT kernel fallback.
+1.  Certain functionality is only available with TF backend. TFRT C API
+    implementation falls back to full TensorFlow in these cases.
+
+I recommend that we prioritize option 1 and try to get it working (i.e. support
```
mhong, line 402: I'd like to understand more about this subject, and also how we should coordinate with kernel registration for native TFRT kernels.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+Performance | Lower overhead        | Higher overhead
+Binary size | Lower (no TF runtime) | Higher
+
+### Performance Implications
+
+*   Slow down due to adding inheritance for `OpKernelContext` and
+    `OpKernelConsturction`.
+*   Speed up for lighter weight kernel calls.
+
+We will run benchmarks to check performance numbers as we work on the
```
mhong, line 540: do we plan to share some numbers (binary size, and performance) as part of the RFC?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+implementation.
+
+### Dependencies
+
+No new dependencies.
+
+### Engineering Impact
```
mhong, line 547: Can you also provide a summary of the current eng status, and the planned/upcoming work, to help us understand the scope of the project, and how to measure success?

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+| **Author(s)** | Anna Revinskaya (annarev@google.com), Jeremy Lau (lauj@google.com) |
+| **Sponsor**   | Jeremy Lau (lauj@google.com)                            |
+| **Updated**   | 2020-07-14                                              |
+
+## Objective
+
+This proposal focuses on getting majority of “well-behaved” ops running in
+[TF Lite](https://www.tensorflow.org/lite) by skipping current eager runtime and
+calling kernels directly from [TFRT](https://github.com/tensorflow/runtime) (a new
+TensorFlow runtime).
```
raziel, line 15: This wording is a bit confusing.
Maybe, ....

"This proposal focuses on getting a majority of "well-behaved" TensorFlow ops running efficiently on
mobile devices by removing the need to execute them via the TensorFlow eager runtime, instead
calling kernels directly from the new [TFRT](https://github.com/tensorflow/runtime) TensorFlow runtime."

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+
+*   100-200k overhead to call TF kernels
+*   20k / kernel marginal size
+
+To get closer to the size restrictions we plan to define a call path from TFRT
+to TensorFlow kernels that minimizes the amount of generated code.
+
+## User Benefit
+
+Running more kernels on mobile devices would allow TensorFlow Lite users to implement
```
raziel, line 119: I'd also add that a reduced binary size will be beneficial to users that currently use TensorFlow Lite's experimental TensorFlow Select ops (https://www.tensorflow.org/lite/guide/ops_select), or users that do not use the experimental feature because of that reason.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+
+  // Creates a kernel with the given name and passes op_kernel_construction
+  // to kernel constructor.
+  // Returns the constructed kernel on success.
+  // In case of failure, returns a nullptr. Kernel creation can fail in one
+  // of the following cases:
+  //   1. Kernel with the given name is not found.
+  //   2. Attributes in op_kernel_construction don't match type constraints
+  //      for any of the kernels with this name.
+  //      Note that we consider a constraint to be "not matched" if attribute
```
raziel, line 154: s/if attribute/if the attribute

rfcs/20200712-tfrt-kernel-fallback.md
raziel: +1

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+                                  -> std::unique_ptr<TFRTOpKernel> {     \
+          return std::make_unique<__VA_ARGS__>(construction);            \
+        }));                                                             \
+    return true;                                                         \
+  }();
+```
+
+## Op registration
+
+To support type specification, we will also provide a minimal Op registry and
```
raziel, line 191: This could benefit from an example of the use case.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+
+*   Use inheritance and define common interfaces.
+*   Use templates.
+
+We ran multiple benchmarks to get an idea of the trade offs between inheritance
+and templating approaches. Key findings are summarized below:
+
+*   Time difference negligible for full model benchmarks.
+*   A simple scalar op benchmark with Kernel Fallback (runs scalar
+    multiplication, division, addition) was only 0.3% slower on mobile with
```
raziel, line 215: worth stating the mobile architecture you run the benchmark on.

rfcs/20200712-tfrt-kernel-fallback.md
bhack: Other then inside the RFC can we propose to publish these numbers as the output of CI? We are still waiting to find a way to publish regular results with current runtime (see https://github.com/tensorflow/build/issues/10)

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+    `TF_OpKernelConstruction` aliases).
+1.  Most of the functions just provide accessors into
+    `OpKernelContext`/`OpKernelConstruction` types.
+
+Given current API structure, we can consider two approaches going forward:
+
+1.  TFRT fully supports all functionality available in the C API. This way any
+    kernel defined using the C API would be automatically available using either
+    full TensorFlow or the TFRT kernel fallback.
+1.  Certain functionality is only available with TF backend. TFRT C API
```
raziel, line 398: how would this "fall back to full TF..." work for the mobile case?

bhack:
Will these fallback kernels maintained in the Tensorflow repo?

rfcs/20200712-tfrt-kernel-fallback.md
annarev: After talking to @zhangqiaorjc and Kuangyuan and they recommended to use `tfrt_fallback.kernel_fallback`.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: We have currently estimate of 1MB for framework and 100k per kernel per type. I will look into getting a better OSS number.

I would say the numbers here are the limit we aim to get to. In general, the plan is to get as close to them as possible.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Updated.

rfcs/20200712-tfrt-kernel-fallback.md
raziel: My main point being that the second option that relies on current TF runtime as fallback is likely not feasible given the goals of kernel fallback for mobile, and/or would require the runtime fallback work to target mobile as well (which is an expressed non-goal at this point).

Because of this, and general cleanliness, I agree with your suggestion below that option 1 should be preferred.

rfcs/20200712-tfrt-kernel-fallback.md
raziel: Agree.
Ideally this should be part of the work towards defining (i.e. creating and registering) kernels for TFRT.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: This paragraph is talking about a choice that can be made at compile time:

* For mobile, we can build with kernel fallback.
* For non-mobile usecase, we can build with full TF.

In more detail: C API will have 2 implementations selected at compile time. Therefore, every kernel implemented using the C API will automatically be supported with kernel fallback (first bullet point above) or full TensorFlow (second bullet point).

rfcs/20200712-tfrt-kernel-fallback.md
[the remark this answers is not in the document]
annarev: Updated.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+
+High level goals of the project:
+
+*   Call existing kernels from new TensorFlow runtime
+*   Reduce size and overhead to make this a feasible option for mobile
+
+We address the first goal by implementing a new fallback mechanism that directly
+calls TensorFlow kernels without going through Eager runtime first. We plan to
+address the second high level goal by trimming down dependencies, switching to
+more compact proto representation, etc..
```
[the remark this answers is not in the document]
annarev: Good to learn something new! Updated.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Updated.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,646 @@
...
+  TF_SetStatus(status, TF_OK, "");
+}
+```
+
+## TFRT integration
+
+Current preferred direction would generate a
+[BEF](https://github.com/tensorflow/runtime/blob/master/documents/binary_executable_format.md) file in advance and then run that
+file on a mobile device. Generated BEF file would have to call either native, TF
+Lite, runtime fallback or kernel fallback kernels and provide any glue logic in
```
raziel, line 478: I think it would be good to clarify here that when the reference to "TF Lite" is the current TF Lite runtime via TFLite Select, as opposed to TFRT running on mobile which is the preferred longer term path.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,646 @@
...
+between (such as tensor conversions).
+
+We also need to consider how kernel or runtime fallback will be selected. This
+could be a parameter at BEF file creation step. It might also be good to package
+both runtime and kernel fallback implementations in a BEF file to be selected at
+runtime.
```
raziel, line 484: I think this should be part of the TF stack that offline will look at the graph and registers to decide which kernel to use.
For example, long term we want to have mobile specific kernels defined (i.e. implemented and registered) via TF authoring APIs, and then we will prefer those to any pre-existing kernels that will be enabled by Kernel Fallback.

I'm not sure we would want to package kernels to be selected ad-hoc at runtime. Maybe I'm not understanding why you mean by it, or having a sample use-case would be useful.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,646 @@
...
+[tensorflow::StringPiece](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/platform/stringpiece.h;l=33;drc=af7fd02ca40f362c4ac96dd064d6a2224b65d784)
+for example) to make it easier to replace all references to save binary size.
+
+@gunhan is also starting an effort to define a library of STL utilities that
+helps us cut down on binary size.
+
+## Selecting which kernels to register
+
+We want to add a script to the TF Lite build setup that can determine required
+kernels based on a model. We would then only build these kernels. For now, we
```
raziel, line 515: I suggest being specific on the parts that are specific to the current TFL runtime vs new TFRT stack.
I think this bit will be mostly (entirely?) relevant to the current TFL runtime stack.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Benchmark setup:

* Pixel 3
* ABI: arm64-v8a
* SDK version: 29

I added this info to the doc as well.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,646 @@
...
+### Alternatives Considered
+
+The main alterantive to TFRT Kernel Fallback is TFRT Runtime Fallback. TFRT
+Runtime Fallback will call TensorFlow Eager C API (corresponding RFC should be
+published soon). Main trade offs between the two fallbacks are described in the
+table below:
+
+Property    | TFRT Kernel Fallback  | TFRT Runtime Fallback
+----------- | --------------------- | ---------------------
+Generality  | Support subset of ops | Support all ops
```
raziel, line 529: I suggest highlighting here important things like resources, which I believe are the main shortcoming of kernel fallback vs runtime fallback.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Good questions. I will try to come up with better way to collect these numbers and share exact details in the next ways.
Sorry for being sloppy about this.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Basically, I didn't look into what it will take to support gradients and decided to postpone this investigation till later.

Based on further discussion, seems like gradient kernels are just normal kernels but there might be further work in registering and calling them. For now, I am just keeping wording in this doc the way it is. We can revisit gradients later.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: I want to understand the difference in supporting one or the other better. Ideally, we want to support both.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: I removed reference to `TF Lite` here. We would want a way to selectively register kernels in TFRT as well.
So, overall, details here still need to be worked out because they have dependencies on TFRT decisions that we don't fully know yet.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Updated.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: I updated Engineering Impact to include this information.

annarev:
@bhack see replies below:
> @annarev So is this just a more or less long temporary solution waiting for native kernel population in TFRT?
Correct.

> Will these fallback kernels be maintained in the Tensorflow repo?
Yes, we just update existing kernel code instead of adding new ones. We will add separate registration macros, but these will also be in TensorFlow repo.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: I added Appendix 2 now. Does it help clarify things? (I actually looked at the two approaches in detail before sending this doc, but excluded details to reduce size of this doc)

This is the most important point of this doc. Once we decide on templates vs inheritance, we get unblocked on migrating more kernels.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Updated.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: We will be upgrading existing kernel code, but we will need 2 separate registrations:

* `REGISTER_KERNEL_FALLBACK_KERNEL`: register kernel with Kernel Fallback.
* `REGISTER_KERNEL_BASE_BUILDER`: register a kernel with current TensorFlow runtime. Note this is different from `REGISTER_KERNEL_BUILDER` that TensorFlow currently uses. We are using a separate kernel registration factory due to reasons in Appendix 1 (we wrap `OpKernelBase` with `OpKernel`).

Reason for 2 registrations:
Current TensorFlow registration mechanism is considered part of *runtime* and we are avoiding dependency on runtime to reduce code size. I suppose we could still have one macro that calls both registrations but we need to make sure it doesn't call TF registration if *only* using Kernel Fallback.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,647 @@
...
+Script details still need to be worked out.
+
+### Alternatives Considered
+
+The main alterantive to TFRT Kernel Fallback is TFRT Runtime Fallback. TFRT
+Runtime Fallback will call TensorFlow Eager C API (corresponding RFC should be
+published soon). Main trade offs between the two fallbacks are described in the
+table below:
+
+Property    | TFRT Kernel Fallback  | TFRT Runtime Fallback
```
zhangqiaorjc: Does it make sense to also compare the "scalability" aspect of both approaches?

For Runtime Fallback, there's no additional per-op or per-kernel work needed after one time setup.

iuuc, for Kernel Fallback, every kernel that we wants to support requires code changes inside Tensorflow? We need to change the kernel interface to take OpKernelContextInterface, and also need explicit registration for each converted kernel. Is that the case?

rfcs/20200712-tfrt-kernel-fallback.md
annarev: I added details for size benchmark in Appendix 3.
To do so, I put opensource prototype of Kernel Fallback here: https://github.com/annarev/tensorflow/tree/kernel_fallback/tensorflow/core/tfrt_fallback/kernel

Based on the opensource prototype, updated size measurements are as follows:

* 900k framework
* 200k per kernel per type

Currently, this is not a propertly setup continuous benchmark, but rather a one-off. Setting up a continuous one would definitely be useful and something we should do once Kernel Fallback is properly committed to TensorFlow.

Performance numbers are harder to compare since this proposal is adding more kernels to run on mobile which we currently don't run on mobile. So, we don't have a good baseline.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: I added a scalability section to the table.

rfcs/20200712-tfrt-kernel-fallback.md
raziel: TFLite and TFRT are separate pieces of infra that, as you mention, will benefit from your work here.
TFLite is the current TF mobile solution, whereas we're working on making TFRT a viable solution for mobile as well.
  
I just mean that to the reader not familiar with these details may get confused by the references to TFLite, and TFRT, and mobile.

Maybe clarifying this further in the Goals section would be good.

rfcs/20200712-tfrt-kernel-fallback.md
```diff
@@ -0,0 +1,837 @@
...
+### Dependencies
+
+No new dependencies.
+
+### Engineering Impact
+
+*   Build / startup time / binary size will be impacted by additional code added
+    to implement Kernel Fallback. At the same time one of the goals of Kernel
+    Fallback is to provide a lower-binary-size way to run existing TensorFlow
+    kernels in TF Lite.
```
raziel, line 582: This is related to my point above. The mobile related goals span also TFRT:
Short term we want to run kernels on TFLite Select, but also run them on mobile via TFRT.

rfcs/20200712-tfrt-kernel-fallback.md
raziel: @miaout17 
Also, we're not planning on BEF calling TFLite kernels.

annarev:
## Design review notes:

*   `tfrt_fallback` as preferred dialect prefix
*   String interning less important now, will consider later
*   Why called fallback? Because we fallback from TFRT
*   Gradients op comment is not super relevant, can be removed
    *   Can kernel fallback be used for training?
        *   Kernels are available
*   Size measurement
    *   Clarified: 200k per kernel per dtype
    *   Should compare TFLite select (Flex) delegate with Kernel Fallback
        *   Is the main saving from the framework or kernel side?
    *   Ideal size number: 200k framework, 20k for kernels
*   Kernel fallback
    *   Short term usable for TFLite select and also TFMRT, to get 400 of kernels while we develop proper native kernels
*   What's required per kernel?
    *   Add base classes, change TF kernels to take interfaces, add registrations
    *   Unexpected kernel change work, because some kernel use special base class
*   TFRT attribute names are sorted alphabetically
    *   Long standing issue, there are some solution, but not committed to any solution yet
*   Kernel registration not related to native kernel
*   Subset of kernels: how many?
    *   Ideally ~400 based on estimate
    *   Will get a better idea of whether we can convert all 400 kernels
*   How to fallback to full TF for kernels that use C API?
    *   Depends on which platform you are using, select at compile time
*   Do we expect external users to use Kernel Fallback to fallback to their custom kernels?
    *   We release two binaries, one with full TF, and the other with kernel fallback
    *   One kernel implementation can work with both
*   Want to use TFRT kernel fallback for both TFLite Select and also TFMRT
    *   TFMRT won't use the same API as TFLite
*   Should be able to package both runtime, kernel fallback
    *   If we do that, we won't get binary size benefits
    *   In the future, may consider kernel fallback to be used together with runtime fallback for non-mobile
*   Correct, this is a temporary solution until we have native kernels
    *   Yes, will maintain them in TF repo
*   ENG cost (not scalability) for the kernel fallback:
    *   Anna is responsible for adding those new kernels
    *   Will provided an ENG estimate in Q4
*   Whether we want to use template or inheritance?
    *   We need to convert a lot of kernels? How to convert kernels? 
    *   On latency, inheritance is slower, due to vtable lookup for small kernels
    *   `basic_ops_benchmark` call methods that are used by all TF kernels
        *   Regression meant lots of more GCU used in TF
    *   Will cause regression on existing TF => this inheritance approach doesn't seem feasible
        *   Could #ifdef stuff we don't want to use; we already tried, but really hard to make it work
        *   Have a simple base class with the common methods? But might as well use template
    *   Only new TF kernels that want conversion require templates
    *   If we go with templates, then would it be easy to just add native kernels?
        *   You will have to add a few lines of code to every kernel
        *   But likely can use ClangMR to update kernels in batch
        *   If it's just a script that can update all kernels, then it seems fine
        *   If it requires a lot more manual work, then maybe we need to prioritize on the set of kernels
    *   What's the current scope of kernels?
        *   There are two for now
        *   Will estimate the ENG cost in Q3. Will implement in Q4.
    *   Should come up with a list
    *   TF has a lot of kernels. Need to depend on Runtime Fallback and/or Kernel Fallback
        *   Kernels outside TF repo do not use special TensorFlow headers but more likely use C++ code that we don't handle
*   What's the next steps?
    *   We are moving forward with kernel fallback, there are some details that need to be figured out
    *   Anna will add more numbers, change inheritance to template, in a week
    *   Need to know how many kernels can be done?
        *   20 not so much
        *   400 is great!
        *   What's the code size overhead? estimated at ~3% based on one kernel

ematejska:
Will mark as accepted once the next steps are completed by Anna above.

annarev:
Just a heads up that I am still working on running more benchmarks to make final template/inheritance decision and update the doc.
I will add a comment here once this doc is ready to merge.

bhack:
@annarev How the kernels are sampled for this template/inheritance benchmarks? By "popularity"? (I suppose that you have a small set)

annarev:
> @annarev How the kernels are sampled for this template/inheritance benchmarks? By "popularity"? (I suppose that you have a small set)

We try to look at kernels that are as small as possible (for e.g. scalar) so that framework overhead is emphasized.

bhack:
Are you also analyzing the compile time? Need we to introduce precompiled headers if we go on with inheritance? Probably you don't care too much about comile time but currently it is already quite heavy to compile Tensorflow sources for contributors without accessing to the "internal" cache (https://github.com/tensorflow/build/issues/5).

annarev:
> Are you also analyzing the compile time? Do we need to introduce precompiled headers if we go on with inheritance? Probably you don't care too much about compile time but currently it is already quite heavy to compile Tensorflow sources for contributors without accessing to the "internal" cache ([tensorflow/build#5](https://github.com/tensorflow/build/issues/5)).

Definitely we don't want to increase compile time significantly given that it already takes a long time. I would actually expect a larger compile time increase due to templates since all templates would need to be expanded. Do you expect inheritance to have a larger build time penalty?
I tried measuring build time, but it is not easy. Basically, building a single op is dominated by all building we already have to do in TensorFlow that I couldn't measure overhead of this specific change.

bhack:
Yes I don't know all the internals but I suppose that It Is a little bit hard to isolate the compile time of a good sample of kernels. In general of course also ISOCPP itself claims an [increased compile time of templates](https://isocpp.org/wiki/faq/templates#template-specialization-speed) but we still need to figure out in our empirical case just to not be aware of this too late (when we have accumulated a significante amount of registered kernels).

rfcs/20200712-tfrt-kernel-fallback.md
annarev: Based on discussion at design review. This is true but not documented.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: I added a clarification in Goals that this document is focusing on running kernels from TFRT while running on mobile. Exact details how TFRT itself is executed on mobile is outside of the scope here.
I also removed reference to TF Lite kernels in this paragraph.

rfcs/20200712-tfrt-kernel-fallback.md
annarev: I removed reference to TF Lite and instead changed it to:
"At the same time one of the goals of Kernel Fallback is to provide a lower-binary-size way to run existing TensorFlow kernels on mobile."
Specifically, this document is only looking at TFRT -> TF kernel interaction on mobile. Whether this path will be triggered purely from TFRT or if a similar approach would be triggered from TF Lite Select is outside of the scope.

annarev:
> Yes I don't know all the internals but I suppose that It Is a little bit hard to isolate the compile time of a good sample of kernels. In general of course also ISOCPP itself claims an [increased compile time of templates](https://isocpp.org/wiki/faq/templates#template-specialization-speed) but we still need to figure out in our empirical case just to not be aware of this too late (when we have accumulated a significante amount of registered kernels).

Quick update. I ran build 3 times at head and 3 times with `OpKernelContext` and `OpKernelConstruction` updated to inherit from interfaces (our preferred approach right now is inheritance, so I didn't test templates).
Time measurements at head (seconds):
`base = np.array([6776.767, 6595.033, 6926.461])`
Time measurements when using inheritance (seconds):
`inheritance = np.array([7109.257, 6722.884, 6660.029])`
p-value (calculated with `stats.ttest_ind(inheritance, base)`) is quite high at `0.72`. Therefore, most likely there is no build time regression.

annarev:
@ematejska This RFC is now ready to accept and merge.

ematejska (approved):

PR #267, opened 2020-07-15
Update Keras categorical input design, mark it as complete.

tanzhenyu:
Only one minor change to split "IndexLookup" to "StringLookup" and "IntegerLookup". The rest is trivial.

But more importantly, add an official guide on how TF1 Estimator users with feature columns can convert their code into TF2 Keras users with KPL.

### DIFF
--- a/rfcs/20191212-keras-categorical-inputs.md
+++ b/rfcs/20191212-keras-categorical-inputs.md
@@ -1,6 +1,6 @@
 # Keras categorical inputs
 
-| Status        | Completed                                            |
+| Status        | Implemented (https://github.com/tensorflow/community/pull/209) |
 :-------------- |:---------------------------------------------------- |
 | **Author(s)** | Zhenyu Tan (tanzheny@google.com), Francois Chollet (fchollet@google.com)|
 | **Sponsor**   | Karmel Allison (karmel@google.com), Martin Wicke (wicke@google.com) |
@@ -23,7 +23,7 @@ Specifically, by introducing the 5 layers, we aim to address these pain points:
 * Users have to define both feature columns and Keras Inputs for the model, resulting in code duplication and deviation from DRY (Do not repeat yourself) principle. See this [Github issue](https://github.com/tensorflow/tensorflow/issues/27416).
 * Users with large dimension categorical inputs will incur large memory footprint and computation cost, if wrapped with indicator column through `tf.keras.layers.DenseFeatures`.
 * Currently there is no way to correctly feed Keras linear model or dense layer with multivalent categorical inputs or weighted categorical inputs, or shared embedding inputs.
-* feature columns offer black-box implementations, mix feature engineering with trainable objects, and lead to
+* Feature columns offer black-box implementations, mix feature engineering with trainable objects, and lead to
   unintended coding pattern.
 
 ## User Benefit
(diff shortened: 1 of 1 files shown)


rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -19,237 +19,407 @@ The proposed layers should support ragged tensors.
 
 ## Motivation
 
-Specifically, by introducing the 4 layers, we aim to address these pain points:
+Specifically, by introducing the 5 layers, we aim to address these pain points:
 * Users have to define both feature columns and Keras Inputs for the model, resulting in code duplication and deviation from DRY (Do not repeat yourself) principle. See this [Github issue](https://github.com/tensorflow/tensorflow/issues/27416).
 * Users with large dimension categorical inputs will incur large memory footprint and computation cost, if wrapped with indicator column through `tf.keras.layers.DenseFeatures`.
-* Currently there is no way to correctly feed Keras linear model or dense layer with multivalent categorical inputs or weighted categorical inputs.
+* Currently there is no way to correctly feed Keras linear model or dense layer with multivalent categorical inputs or weighted categorical inputs, or shared embedding inputs.
+* feature columns offer black-box implementations, mix feature engineering with trainable objects, and lead to
```
goldiegadde, line 26: nit: feature/Feature

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -19,237 +19,407 @@ The proposed layers should support ragged tensors.
...
 ```
 
-### Workflow 2
+Note the hashed output from KPL will be different than the hashed output from feature column, given how seed is choosen. `Hashing` also supports customized `salt`.
 
-The second example gives an instruction on how to transition from categorical feature columns to the proposed layers. Note that one difference for vocab categorical column is that, instead of providing a pair of mutually exclusive `default_value` and `num_oov_buckets` where `default_value` represents the value to map input to given out-of-vocab value, and `num_oov_buckets` represents value range of [len(vocab), len(vocab)+num_oov_buckets) to map input to from a hashing function given out-of-vocab value. In practice, we believe out-of-vocab values should be mapped to the head, i.e., [0, num_oov_tokens), and in-vocab values should be mapped to [num_oov_tokens, num_oov_tokens+len(vocab)).
+2. `tf.feature_column.categorical_column_with_identity`
+This feature column is merely for having identical inputs and outputs except mapping out-of-range value into `default_value`, thus can easily be done at data cleaning stage,
+not be part of feature engineering, and hence dropped in this proposal.
```
goldiegadde, line 56: nit: for better readability, do you want to move this to the end , since this is not replaced and deemed not needed in feature engineering.

rfcs/20191212-keras-categorical-inputs.md
```diff
@@ -1,15 +1,15 @@
 # Keras categorical inputs
 
-| Status        | Accepted                                             |
+| Status        | Completed                                            |
```
ematejska, line 3: Do you mean it's implemented?  I would prefer the terminology Implemented instead of Completed.  Could you also post the PR number where it was implemented?  Something like:

Status: Implemented (Link to PR)

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Done.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Done.

rfcs/20191212-keras-categorical-inputs.md
tanzhenyu: Done.

rfcs/20191212-keras-categorical-inputs.md
goldiegadde: sg!

PR #268, opened 2020-07-16
Add a note about Python versions

martinwicke:
Clarify how Python version compatibility is handled. This is mostly covered by testing, since we do run tests with the minimum version, but not entirely (as seen with 3.8 and async).

### DIFF
--- a/governance/api-reviews.md
+++ b/governance/api-reviews.md
@@ -23,7 +23,7 @@ the comments to get an answer.
 
 TensorFlow supports a range of Python versions, and changes need to be
 compatible with all of them. This means that language features not available in
-TensorFlow's minimum supported version cannot be used. 
+any of TensorFlow's supported version cannot be used.
 
 We regularly reconsider the range of supported versions based on the number of
 affected users (estimated via pip downloads by Python version), and the default
--- a/governance/api-reviews.md
+++ b/governance/api-reviews.md
@@ -19,7 +19,20 @@ the comments to get an answer.
 
 ## High level points
 
+### Python Versions
+
+TensorFlow supports a range of Python versions, and changes need to be
+compatible with all of them. This means that language features not available in
+TensorFlow's minimum supported version cannot be used. 
+
+We regularly reconsider the range of supported versions based on the number of
+affected users (estimated via pip downloads by Python version), and the default
+installed version of Python on important platforms (Linux distributions, OSX,
+...), but these are large, breaking changes to the ecosystem which are not
+triggered by reviews of individual features.
+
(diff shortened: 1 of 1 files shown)


governance/api-reviews.md
```diff
@@ -19,7 +19,20 @@ the comments to get an answer.
 
 ## High level points
 
+### Python Versions
+
+TensorFlow supports a range of Python versions, and changes need to be
```
terrytangyuan, line 24:
```suggestion
TensorFlow supports a range of Python versions and changes need to be
```

governance/api-reviews.md
```diff
@@ -19,7 +19,20 @@ the comments to get an answer.
...
+compatible with all of them. This means that language features not available in
+TensorFlow's minimum supported version cannot be used. 
+
+We regularly reconsider the range of supported versions based on the number of
```
terrytangyuan, line 28: Also consider end of life timelines? https://endoflife.date/python

martinwicke: As seen by Python2, these are mostly aspirational. Realistically, we will make these decisions based on users, not timelines.

martinwicke:
@karmel: The issue is that we don't actually tests all versions. I don't want to separately encode the supported versions, I guess I should point at https://www.tensorflow.org/install.

governance/api-reviews.md
bhack: Are we suggesting to contributors to develop with the minimal supported python version as in `https://www.tensorflow.org/install`?

Cause `docker run --rm tensorflow/tensorflow:devel python3 -V` is `Python 3.6.9`

martinwicke:
Minimum isn't sufficient. For instance 3.6.9 will happily accept `def
test(async=False): ...`, but 3.8 will break.

The point is that your code has to work in the range. If it doesn't, we
might reject it. We should find most problems in testing, so I don't think
this policy would create much confusion.

bhack:
Oh I see the text was changed in another commit, in this case only test infra could verify a full range.

governance/api-reviews.md
byronyi: OSS TF presubmit tests are targeted to Python 3.6. But since we do have tags like `no_oss_py35` etc., it suggests that TF team might run those tests internally for other Python versions as well.

bhack:
Do you think it could be useful to maintain a "machine readable" version of this range? Cause seems that other then in the install page it is controlled only by `TF_PYTHON_VERSION` env

mihaimaruseac (approved):

PR #278, opened 2020-08-11
RFC: tf.__internal__ API namespace.

qlzh727:
Comment period is open until 08/25/2020.

| Status        | Proposed |
| **RFC #**     | [278](https://github.com/tensorflow/community/pull/278)|
| **Author(s)** | Qianli Zhu (scottzhu@google.com) |
| **Sponsor**   | Martin Wicke (wicke@google.com), Alex Passos (apassos@google.com)|
| **Updated**   | 2020-09-03                                          |
| **Intended audience**| tf-api-owners, tf-addons, keras-team, deepmind/sonnet|

## Objective

Adding a new `__internal__` API namespace in TF to host APIs for framework building/testing, etc. The API namespace will serve as a whitelist for client libraries to gradually migrate off the usage tf private API symbol.

## Notice:

This design has been updated comparing to its original proposal. The major changes are:

1. The namespace has been renamed from "tf.internal" to `tf.__internal__`, which gives user more indication that it is a private/protected API.
2. The package serves as a placeholder for private APIs symbols that are currently used by high level APIs and other libraries. Those symbols are not supposed to be public API, and are useful for build/test certain functionality by the client side.
3. This package also serves as a tracking list for the private API usage from high level API like Keras. Any addition to this package will require API owner review with justifications.
4. OSS lib (like tf-addons) shouldn't rely on this as a reliable source of functionality. It doesn't have a formal API contract. Only use it if the TF team is aware of the usage, so that we can properly detect it when we make the change.

### DIFF
--- a/rfcs/20200810-tf-internal-api.md
+++ b/rfcs/20200810-tf-internal-api.md
@@ -1,4 +1,4 @@
-# tf.internal API namespace
+# tf._internal API namespace
 
 | Status        | Proposed |
 :-------------- |:---------------------------------------------------- |
@@ -6,11 +6,14 @@
 | **Author(s)** | Qianli Zhu (scottzhu@google.com) |
 | **Sponsor**   | Martin Wicke (wicke@google.com), Alex Passos (apassos@google.com)|
 | **Updated**   | 2020-08-10                                           |
-| **Intended audience**| tf-api-owners, tf-addons, keras-team, deepmind/sonnet|
+| **Intended audience**| tf-api-owners, keras-team|
 
 ## Objective
 
-Adding a new "internal" API namespace in TF to host APIs for framework building/testing, etc. The API will have looser contracts compared to core TF API.
+Adding a new "_internal" API namespace in TF to host APIs for framework building
+/testing, etc. The API namespace will serve as a whitelist for client libraries 
+to gradually migrate off the usage tf private API symbol.
+
 
 ## Motivation
 
@@ -27,63 +30,62 @@ rely on the public TF API only, to avoid any future breakages due to private
 method/class change.
 
 Historically, Keras was an internal component of TF, and relies heavily on a lot
-of private TF utilities and functionalities that are not part of the public API. 
-Exposing those functions into public APIs will be very confusing to the end 
-user, since those functions are useful to framework-builders, but not typical 
-TensorFlow users. In addition, these utilities would be overly restricted by the 
-standard TensorFlow API contract, as we would like to change/remove them more 
-often than the existing core API.
+of private TF utilities and functionalities that are not part of the public 
+API. Exposing those functions into public APIs will be very confusing to the 
+end user who uses tf to build/train models, since those functions are useful to
+framework-builders, but not typical TensorFlow users.
 
 
 ## Design Proposal
 
-Add a new namespace "tensorflow.internal" to host the framework building/testing 
-related APIs. <b>(See alternative naming in the sections below, the naming of 
-the namespace will be discussed in the design meeting.)</b>
+Add a new namespace "tensorflow._internal" to host the framework building/
+testing related APIs that are currently used by other high level API in TF 
+(Keras, and estimator, etc.). This name space will be treated as a protected 
+API, which shouldn't be used by the end users. It will have no API contract 
+compared to core TF API. TF side must be aware of the usage from client lib, 
+so that it can properly test the change to any of those internal API before TF 
+side making any backward incompatible change.
 
 ### Backward compatibility
-The contract for all the APIs under this namespace is looser compared to core TF
-API. It will remain backward compatible for at least 1 minor TF release. 
-
-This means for any API that is added in 2.x release, it will remain the same in 
-2.x+1. If we choose to remove the same API, we will mark it as deprecated in 
-2.x+1, and delete it at 2.x+2 release. TensorFlow is released every 3 months, 
-and this will give enough time for all the clients to change to the new 
-API/alternative.
-
-Any deprecation and backward incompatible API change will be explicitly called 
-out in TF release notes. This applies to both core API and "internal" API.
+There is no API contract for all the functions under this namespace. TF side 
+can make any backward incompatible change at any time. To prevent any breakage 
+caused on the client side, TF must be aware who is using this API, so that any 
+changes to those API will be verified against the client code.
 
 ### Acceptance Criteria
-The candidate of the "internal" API should:
+The candidate of the "_internal" API should:
 
 1. Does NOT fit for core TF API, otherwise it should be exposed as core TF API.
-1. Are effectively required to build/test/deploy a framework or library on top 
-   of TensorFlow, or to integrate such a framework into TensorFlow's API.
-1. Mature enough and won't change behavior/interface for every release. 
+1. Are currently used by other high level API to build/test/deploy a framework 
+   or library on top of TensorFlow, or to integrate such a framework into 
+   TensorFlow's API.
 1. Non-trivial, otherwise the function should be copied to the client side.
-1. Widely used in the implementation of TF public APIs (i.e. new functionality 
-   isn’t immediately added to the tf.internal namespace)
-1. Has at least two downstream libraries which are known to need it.
+1. Widely used in the implementation of TF public APIs or test TF related 
+   features. (i.e. new functionality isn’t immediately added to the tf._internal
+   namespace)
 
 TF API owners will review the new API proposal, and follow the existing review 
 process for core TF API.
 
 
+### Lifespan of the internal API
+The internal API should serve as a temporary solution for the client side, and 
+they should gradually migrate to alternative public TF core API, or other 
+approaches. Once all the usage for a certain internal API has been removed, it 
+will be deleted from the internal API namespace.
+
+
 ### Documentation and Docstring
 The "internal" API should have the same style and standard as the core 
 TensorFlow API, which is documented [here](https://github.com/tensorflow/community/blob/master/governance/api-reviews.md#docstrings). 
-We should explicitly list out the difference between "internal" API and core
-API, and also choose a different place on tensorflow.org so that existing user 
-are not confused.
+We will hide this from the documentation site, since the end users are the 
+target audience for this API.
 
 ### Naming and sub namespace
 Similar "internal" APIs should be grouped together as sub namespaces, e.g., test 
 related APIs should live under "tf.internal.test". This is aligned with the 
 existing TF naming conversion.
 
-Try not to export experimental APIs since the "internal" API should be mature 
-enough.
 
 ### Current candidate
 The following list is created from the private TF method usage within Keras, 
@@ -110,9 +112,3 @@ and will discuss with the API owner on a case to case basis.
    to add are still high level APIs and utility functions. 
 1. By Martin <b>"tf.\_internal"</b>: the extra "_" emphasis in the pythonic way
    that this is for private usage.
-
-
-## Questions and Discussion Topics
-
-1. Naming of the API namespace.
-
--- a/rfcs/20200810-tf-internal-api.md
+++ b/rfcs/20200810-tf-internal-api.md
@@ -1,6 +1,6 @@
 # tf._internal API namespace
 
-| Status        | Proposed |
+| Status        | Accepted |
 :-------------- |:---------------------------------------------------- |
 | **RFC #**     | [278](https://github.com/tensorflow/community/pull/278)|
 | **Author(s)** | Qianli Zhu (scottzhu@google.com) |
--- a/rfcs/20200810-tf-internal-api.md
+++ b/rfcs/20200810-tf-internal-api.md
@@ -4,7 +4,7 @@
 :-------------- |:---------------------------------------------------- |
 | **RFC #**     | [278](https://github.com/tensorflow/community/pull/278)|
 | **Author(s)** | Qianli Zhu (scottzhu@google.com) |
-| **Sponsor**   | Martin Wicke (wicke@google.com), Alex Apassos (apassos@google.com)|
+| **Sponsor**   | Martin Wicke (wicke@google.com), Alex Passos (apassos@google.com)|
 | **Updated**   | 2020-08-10                                           |
 | **Intended audience**| tf-api-owners, tf-addons, keras-team, deepmind/sonnet|
 
@@ -97,10 +97,9 @@ and will discuss with the API owner on a case to case basis.
 |python.framework.func_graph.FuncGraph |tf.internal.FuncGraph     |
 |python.framework.combinations.*| tf.internal.test.combinations.* |
 |python.distribute.strategy_combinations.* |tf.internal.test.combinations.* |
-|python.training.tracking.base.no_automatic_dependency_tracking|tf.internal.tracking.no_automatic_dependency_tracking |
 |python.util.object_identity.*|tf.internal.object_identity.* |
-|python.util.tf_decorator.*|tf.internal.decorator.* |
-|python.util.tf_inspect.*|tf.internal.inspect.* |
+|python.util.tf_decorator.TFDecorator|tf.internal.decorator.TFDecorator |
+|python.util.tf_decorator.unwrap|tf.internal.inspect.unwrap |
 
 ### Alternative Names
 1. <b>"tf.internal"</b>: It gives the user the impression that this is not a 
@@ -109,7 +108,7 @@ and will discuss with the API owner on a case to case basis.
 1. <b>"tf.infra"</b>: infrastructure is aligned with "building blocks" and low 
    level functionalities, like file system/network etc. So far, the APIs we want
    to add are still high level APIs and utility functions. 
-1. By Martin <b>"tf._internal"</b>: the extra "_" emphasis in the pythonic way
+1. By Martin <b>"tf.\_internal"</b>: the extra "_" emphasis in the pythonic way
    that this is for private usage.
 
 
--- a/rfcs/20200810-tf-internal-api.md
+++ b/rfcs/20200810-tf-internal-api.md
@@ -2,7 +2,7 @@
 
 | Status        | Proposed |
 :-------------- |:---------------------------------------------------- |
-| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)|
+| **RFC #**     | [278](https://github.com/tensorflow/community/pull/278)|
 | **Author(s)** | Qianli Zhu (scottzhu@google.com) |
 | **Sponsor**   | Martin Wicke (wicke@google.com), Alex Apassos (apassos@google.com)|
 | **Updated**   | 2020-08-10                                           |
--- a/rfcs/20200810-tf-internal-api.md
+++ b/rfcs/20200810-tf-internal-api.md
@@ -1,0 +1,119 @@
+# tf.internal API namespace
+
+| Status        | Proposed |
+:-------------- |:---------------------------------------------------- |
+| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)|
+| **Author(s)** | Qianli Zhu (scottzhu@google.com) |
+| **Sponsor**   | Martin Wicke (wicke@google.com), Alex Apassos (apassos@google.com)|
+| **Updated**   | 2020-08-10                                           |
+| **Intended audience**| tf-api-owners, tf-addons, keras-team, deepmind/sonnet|
+
+## Objective
+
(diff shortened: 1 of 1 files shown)


rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,119 @@
...
+send to the API owner for review. We don't expect all of them to be approved, 
+and will discuss with the API owner on a case to case basis.
+
+|Symbol location  |API Name  | 
+:-------------- |:---------------------------------------------------- |
+|python.framework.func_graph.FuncGraph |tf.internal.FuncGraph     |
+|python.framework.combinations.*| tf.internal.test.combinations.* |
+|python.distribute.strategy_combinations.* |tf.internal.test.combinations.* |
+|python.training.tracking.base.no_automatic_dependency_tracking|tf.internal.tracking.no_automatic_dependency_tracking |
+|python.util.object_identity.*|tf.internal.object_identity.* |
```
allenlavoie, line 101: object_identity.py doesn't depend on TF. Should it be a separate (pure Python) library that both TF and Keras use?

There's a more general question about whether that kind of dependency is OK to put in tf.internal due to the friction of splitting it out / packaging it / etc., balanced with the danger of tf.internal becoming a repository for small-ish nifty Python things that aren't otherwise TF related.

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,119 @@
...
+|python.util.tf_inspect.*|tf.internal.inspect.* |
+
+### Alternative Names
+1. <b>"tf.internal"</b>: It gives the user the impression that this is not a 
+   standard user facing API (good), and contains some private/implementation 
+   details that are internal to TF. 
+1. <b>"tf.infra"</b>: infrastructure is aligned with "building blocks" and low 
+   level functionalities, like file system/network etc. So far, the APIs we want
+   to add are still high level APIs and utility functions. 
+1. By Martin <b>"tf._internal"</b>: the extra "_" emphasis in the pythonic way
```
allenlavoie, line 112: nit: markdown is swallowing underscores, might want to escape them

rfcs/20200810-tf-internal-api.md
martinwicke: Yes, it can, and I think it would be minimally troubling since it's a simple, rarely or never changing, pure Python package.

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,119 @@
...
+
+|Symbol location  |API Name  | 
+:-------------- |:---------------------------------------------------- |
+|python.framework.func_graph.FuncGraph |tf.internal.FuncGraph     |
+|python.framework.combinations.*| tf.internal.test.combinations.* |
+|python.distribute.strategy_combinations.* |tf.internal.test.combinations.* |
+|python.training.tracking.base.no_automatic_dependency_tracking|tf.internal.tracking.no_automatic_dependency_tracking |
+|python.util.object_identity.*|tf.internal.object_identity.* |
+|python.util.tf_decorator.*|tf.internal.decorator.* |
+|python.util.tf_inspect.*|tf.internal.inspect.* |
```
martinwicke, line 103: I think there is a smaller set of symbols which would be enough: TFDecorator and unwrap, right?

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,119 @@
...
+no means the full list, but will serve as the first wave of review requests we 
+send to the API owner for review. We don't expect all of them to be approved, 
+and will discuss with the API owner on a case to case basis.
+
+|Symbol location  |API Name  | 
+:-------------- |:---------------------------------------------------- |
+|python.framework.func_graph.FuncGraph |tf.internal.FuncGraph     |
+|python.framework.combinations.*| tf.internal.test.combinations.* |
+|python.distribute.strategy_combinations.* |tf.internal.test.combinations.* |
+|python.training.tracking.base.no_automatic_dependency_tracking|tf.internal.tracking.no_automatic_dependency_tracking |
```
martinwicke, line 100: This sounds like a symbol that should be public, not internal.

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,119 @@
...
+### Current candidate
+The following list is created from the private TF method usage within Keras, 
+when we were trying to convert Keras to use the public TF API only. This is by 
+no means the full list, but will serve as the first wave of review requests we 
+send to the API owner for review. We don't expect all of them to be approved, 
+and will discuss with the API owner on a case to case basis.
+
+|Symbol location  |API Name  | 
+:-------------- |:---------------------------------------------------- |
+|python.framework.func_graph.FuncGraph |tf.internal.FuncGraph     |
```
martinwicke, line 97: @saxenasaurabh mentioned usages could be replaced with `tf.function(f).get_concrete_function(inputs).graph`? That would be preferable if feasible.

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,119 @@
...
+1. <b>"tf.infra"</b>: infrastructure is aligned with "building blocks" and low 
+   level functionalities, like file system/network etc. So far, the APIs we want
+   to add are still high level APIs and utility functions. 
+1. By Martin <b>"tf._internal"</b>: the extra "_" emphasis in the pythonic way
+   that this is for private usage.
+
+
+## Questions and Discussion Topics
+
+1. Naming of the API namespace.
```
martinwicke, line 118: I want to suggest we do not expose these symbols as a public namespace, and instead use "_internal". This would avoid creating another carveout from semver, and the effect for downstream projects is the same (they have to know what they're doing). 

Alternatively, we could expose these as a module separate from tensorflow (similar to tfdbg or other projects are in separate modules).

rfcs/20200810-tf-internal-api.md
tomerk: Would the `tf.function` traces get saved in some sort of never-cleared cache? Or is it possible to explicitly gc them? part of our usage of `FuncGraph` is to encapsulate traces that can then be immediately cleaned up.

tomerk: And also related to this: would it trace twice rather than once? That would also be a performance hit to our existing usages.

martinwicke: I would much prefer exposing this type of behavior as regular arguments to get_concrete_function (or similar), since these do appear to be valid optimizations that other framework authors may want to use.

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,119 @@
+# tf.internal API namespace
+
+| Status        | Proposed |
+:-------------- |:---------------------------------------------------- |
+| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)|
+| **Author(s)** | Qianli Zhu (scottzhu@google.com) |
+| **Sponsor**   | Martin Wicke (wicke@google.com), Alex Apassos (apassos@google.com)|
```
lithuak, line 7: Apassos -> Passos

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,119 @@
...
+and this will give enough time for all the clients to change to the new 
+API/alternative.
+
+Any deprecation and backward incompatible API change will be explicitly called 
+out in TF release notes. This applies to both core API and "internal" API.
+
+### Acceptance Criteria
+The candidate of the "internal" API should:
+
+1. Does NOT fit for core TF API, otherwise it should be exposed as core TF API.
```
bhack, line 60: Can you expand on point 1? It Is very ambiguous.

rfcs/20200810-tf-internal-api.md
bhack: > There's a more general question about whether that kind of dependency is OK to put in tf.internal due to the friction of splitting it out / packaging it / etc., balanced with the danger of tf.internal becoming a repository for small-ish nifty Python things that aren't otherwise TF related.

I have the same concern for the general topic.

qlzh727: Yes, ideally this doesn't have TF specific logic and can be put into a standalone python package. Do we have any existing package that can host it (eg google absl?) or we need to create a new one?

rfcs/20200810-tf-internal-api.md
qlzh727: The general guild line for TF API is https://github.com/tensorflow/community/blob/master/governance/api-reviews.md. API owner group will discuss individual API proposal in the weekly meeting.

rfcs/20200810-tf-internal-api.md
qlzh727: Thanks for the notice. Updated.

rfcs/20200810-tf-internal-api.md
qlzh727: Yes, the inspect behavior only need to unwrap. Updated.

rfcs/20200810-tf-internal-api.md
qlzh727: Ack. Remove it from the list, and will send a cl for making it a public API.

rfcs/20200810-tf-internal-api.md
qlzh727: I think the naming of the namespace will be the biggest topic for the design meeting, and the "tf._intenral" has been listed as one of the option here.

Since those module would have dependency from and to TF code, its code won't be able to split out from TF repo. I am not sure if they could live under a different named PIP package (if this is what you mean for a "separate module").

rfcs/20200810-tf-internal-api.md
qlzh727: The general guild line for TF API is https://github.com/tensorflow/community/blob/master/governance/api-reviews.md. API owner group will discuss individual API proposal in the weekly meeting.

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,118 @@
...
+
+TF API owners will review the new API proposal, and follow the existing review 
+process for core TF API.
+
+
+### Documentation and Docstring
+The "internal" API should have the same style and standard as the core 
+TensorFlow API, which is documented [here](https://github.com/tensorflow/community/blob/master/governance/api-reviews.md#docstrings). 
+We should explicitly list out the difference between "internal" API and core
+API, and also choose a different place on tensorflow.org so that existing user
```
qlzh727, line 77: @markdaoust and @yashk2810, do u have any suggestion here for a place on tensorflow.org to host this internal API?

yashk2810: If its in the TF package then it will be on the API docs page here: https://www.tensorflow.org/api_docs/python/tf?version=nightly with a `tf.internal` namespace as all the others in the left TOC.

Does that work?

martinwicke: How many symbols are going to be left in this? I'd still prefer we don't have this publicly.

yashk2810: Yes, we can block this from being displayed on tensorflow.org if that's the way we want to go.

qlzh727: I think we can block it in the beginning, since it will cause confusion for end user. We might need to choose a different place for it, targeting framework builder only.

rfcs/20200810-tf-internal-api.md
kkimdev: It would be helpful to see the actual usages to determine this, but to answer the questions:

- Caching: Unless the graph is held elsewhere, deleting the instance of tf.function(f) should delete everything.
- Number of traces: Once if first trace doesn't create any variable, Twice if first trace creates variables

cc @saxenasaurabh to keep me honest.

rfcs/20200810-tf-internal-api.md
jaingaurav: I don't think we should expose object_identity at all. Rather, we should push framework developers to use `ref` or just solve the problem independently. In particular, don't see why should make this API available to framework builders and not for end-users.

rfcs/20200810-tf-internal-api.md
jaingaurav: Don't we need to trim the FuncGraph API before considering exposing this?

rfcs/20200810-tf-internal-api.md
jaingaurav: Also I think object_identity is fairly trivial and thus I believe fails one of the acceptance criteria.

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,118 @@
...
+This means for any API that is added in 2.x release, it will remain the same in 
+2.x+1. If we choose to remove the same API, we will mark it as deprecated in 
+2.x+1, and delete it at 2.x+2 release. TensorFlow is released every 3 months, 
+and this will give enough time for all the clients to change to the new 
+API/alternative.
+
+Any deprecation and backward incompatible API change will be explicitly called 
+out in TF release notes. This applies to both core API and "internal" API.
+
+### Acceptance Criteria
```
jaingaurav, line 57: nit: please fix numbering.

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Naming and sub namespace
+Similar "internal" APIs should be grouped together as sub namespaces, e.g., test 
+related APIs should live under "tf.internal.test". This is aligned with the 
+existing TF naming conversion.
+
+Try not to export experimental APIs since the "internal" API should be mature 
+enough.
+
+### Current candidate
```
jaingaurav, line 88: Could you please document the 2+ downstream libraries that require each of these APIs?

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,118 @@
...
+and will discuss with the API owner on a case to case basis.
+
+|Symbol location  |API Name  | 
+:-------------- |:---------------------------------------------------- |
+|python.framework.func_graph.FuncGraph |tf.internal.FuncGraph     |
+|python.framework.combinations.*| tf.internal.test.combinations.* |
+|python.distribute.strategy_combinations.* |tf.internal.test.combinations.* |
+|python.util.object_identity.*|tf.internal.object_identity.* |
+|python.util.tf_decorator.TFDecorator|tf.internal.decorator.TFDecorator |
+|python.util.tf_decorator.unwrap|tf.internal.inspect.unwrap |
```
jaingaurav, line 102: What is TF-specific about the decorator libraries? Are these utilities candidates to simply be moved to a separate repo?

rfcs/20200810-tf-internal-api.md
saxenasaurabh: +1 to @kkimdev. Caching should not be a concern here because the lifetime of cached entries is tied to the instance returned by `tf.function(f)`.
Regarding multiple traces, it should not be a concern if variables are not being created. And if there are, then using `tf.function` is actually safer I think.

martinwicke:
@qlzh727 can we evaluate alternatives? It seems to me that we're talking about a very small set of symbols, some of which are independent of TF and could live in their own packages (object_identity, maybe the decorator). Some of which have "better" replacements (FuncGraph?). This leaves us with combinations (independent of TF) and strategy_combinations.

It's the latter that gives me a headache. But I'm not sure we need a special namespace for it, rather, we need to think through what we actually want to do with it (since now we're talking about a single thing).

rfcs/20200810-tf-internal-api.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Backward compatibility
+The contract for all the APIs under this namespace is looser compared to core TF
+API. It will remain backward compatible for at least 1 minor TF release. 
+
+This means for any API that is added in 2.x release, it will remain the same in 
+2.x+1. If we choose to remove the same API, we will mark it as deprecated in 
+2.x+1, and delete it at 2.x+2 release. TensorFlow is released every 3 months, 
+and this will give enough time for all the clients to change to the new 
+API/alternative.
```
kkimdev, line 52: Q: Is this something that Keras will take advantage of?

rfcs/20200810-tf-internal-api.md
qlzh727: > @qlzh727 can we evaluate alternatives? It seems to me that we're talking about a very small set of symbols, some of which are independent of TF and could live in their own packages (object_identity, maybe the decorator). Some of which have "better" replacements (FuncGraph?). This leaves us with combinations (independent of TF) and strategy_combinations.
> 
> It's the latter that gives me a headache. But I'm not sure we need a special namespace for it, rather, we need to think through what we actually want to do with it (since now we're talking about a single thing).

Sure. Let me update the doc and see if there is any possible alternatives.

mdanatg:
> It's the latter that gives me a headache.

If it's just a testing utility, has forking been considered?

qlzh727:
> > It's the latter that gives me a headache.
> 
> If it's just a testing utility, has forking been considered?

Yes, we usually fork the util if it is small enough and doesn't have any dependency with other private API. The combination does have some private api deps like context.eager_context(), tf2.eabled() etc. The distribute combination has more deps with strategy and other DS related classes, which are not public API yet

qlzh727:
Selected notes from the design meeting:

=====================
Q: the _underscore in "tf._internal"
A: I really want to make it very hard for people to use this.
Q: What if some of these things are useful?
A: Using internals of TF is bad, unquestionably, because it makes it impossible for us to refactor. If we collect them in a place it’s less bad so we can triage and monitor and make them go away. Some functionality is missing from our API but should be available. Err on the side of “if we can’t promise anything then let’s not promise anything”. The status of the API doesn’t depend on the guarantees, it depends on the usage (hyrum’s law and all that).
Conclusion: will keep the underscore.

=====================
Q: Release timeline?
A: Want it available on 2.4 so keras can use it then.
Conclusion: Will make it available on 2.4.

=====================
Q: Is the plan to dump everything to _internal?
A: No, we’ll be careful to only put things that is useful in _internal.

=====================
Q: why do we need to make it available in a specific version of TF?
A: keras needs to import TF and then use TF._internal.

=====================
Q: what is the deprecation process for the API
A: First the usage on Keras side need to be removed, then we can remove the internal api. Keras nightly builds against TF nightly so this is ok.

bhack:
What is the stauts of this?

qlzh727:
This is implemented, and we somehow haven't close this RFC yet. Let me check with the PM for this.

bhack:
Ok, so Is there a place where we can track what candidates we have in queue for this namespace?
E.g. I often see in this RFC that we talk about testing but just checking one of the main testing utility (for layers) it is still not migrated:
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/testing_utils.py#L87

qlzh727:
So far we don't have a public user visible place to track all the items (since they are suppose to hide from normal users). You can probably get all the \_\_internal__ just by code search them https://cs.opensource.google/tensorflow/tensorflow. Keras code are not exposed under tf.\_\_internal__ namespace, and if you would like to add more (eg for layer_test), feel free to send PRs to keras-team/keras

bhack:
> Keras code are not exposed under tf.__internal__ namespace, and if you would like to add more (eg for layer_test), feel free to send PRs to keras-team/keras

In general the problem Is that if I don't see what you have in the internal queue/roadmap It hard to understand what is WIP and instead what we need to contribute.

E.g. for the specific mentioned example we have this since 2019:

https://github.com/tensorflow/tensorflow/issues/28601


If you don't have  any other in progress migration related to this RFC we could discuss the specific case in that ticket.

qlzh727:
> > Keras code are not exposed under tf.**internal** namespace, and if you would like to add more (eg for layer_test), feel free to send PRs to keras-team/keras
> 
> In general the problem Is that if I don't see what you have in the internal queue/roadmap It hard to understand what is WIP and instead what we need to contribute.
> 
> E.g. for the specific mentioned example we have this since 2019:
> 
> [tensorflow/tensorflow#28601](https://github.com/tensorflow/tensorflow/issues/28601)
> 
> If you don't have any other in progress migration related to this RFC we could discuss the specific case in that ticket.

Sure. We don't have any active migration right now. Let's update that ticket instead.

ematejska:
@qlzh727 What's the status of this RFC?

qlzh727:
Oh, this one is done long times ago, and we can mark it as closed/implementated.

ematejska (approved):

qlzh727:
Forgot to update some of the naming in this PR. See most recent fix in https://github.com/tensorflow/community/pull/410

PR #287, opened 2020-09-03
RFC: Enabling TensorFloat-32 by default

reedwm:
This RFC will be open for comment until October 14, 2020. The comment period is longer than usual in order to give a lot of time to investigate and test TF32.

# Enabling TensorFloat-32 by default

| **RFC #**     | [287](https://github.com/tensorflow/community/pull/287)|
| **Author(s)** | Reed Wanderman-Milne (reedwm@google.com)             |
| **Sponsor**   | Sanjoy Das (sanjoy@google.com)                       |
| **Updated**   | 2020-09-03                                           |

## Objective

Enable [TensorFloat-32](https://blogs.nvidia.com/blog/2020/05/14/tensorfloat-32-precision-format/) by default. 

This will improve performance but lower numeric precision for certain ops like matmul and convolutions by default.


rfcs/20200903-tensor-float-32-default.md
```diff
@@ -0,0 +1,46 @@
...
+
+## User Benefit
+
+The benefit of turning on TF32 by default is that all Ampere users will get the performance benefit, even if they don’t know about TF32 and therefore do not call the API function to enable it.
+
+Turning on TF32 by default also makes the GPU behavior more consistent with TPUs. TPUs also use a lower-precision internal dtype (bfloat16) for matrix multiplications and convolutions by default when float32 is used.
+
+## Design Proposal
+
+The design is to flip the default value of [`tf.config.experimental.enable_tensor_float_32_execution`](https://www.tensorflow.org/api_docs/python/tf/config/experimental/enable_tensor_float_32_execution) from False to True.
```
mdanatg, line 30: Would it be practical to provide a user control for setting the default type with a global toggle, for all numerical ops? That might be useful in case there isn't a one-size-fits-all.

reedwm: This is somewhat orthogonal to TensorFloat-32, since TF32 is not a TensorFlow dtype. Even with TF32 enabled, `tf.linalg.matmul` outputs float32 tensors when passed float32 inputs. The only effect of TF32 is increased performance and inputs to ops like matmul are rounded from 23 bits of precision to 10 bits of precision.

There should be an API to set the default dtype for ops like `tf.ones` that don't take an input tensor. But I think ops like matmul should continue to unconditionally output tensors of the same dtype as the inputs.

ematejska:
Design Reive Notes were added to the RFC document.  This is accepted.

ematejska (approved):

PR #291, opened 2020-09-28
RFC: Sharded Variable for Parameter Server Training

ckkuang:
*This RFC will be open for comment until Monday, October 12th, 2020.*

# Sharded Variable For Parameter Server Training

| Status        | Proposed                                           |
| **Author(s)** | Chenkai Kuang (chenkai@google.com)                   |
| **Sponsor**   | Priya Gupta (priyag@google.com)                      |
| **Updated**   | 2020-09-28                                           |

# Objective

Support variable partitioning in single-client parameter server distribution strategy ([RFC](https://github.com/tensorflow/community/blob/master/rfcs/20200306-single-client-parameter-server.md)), to scale up cases like large embedding lookups and large variable fetches and updates.


ckkuang:
Design review notes:

- Initializer change
    - Francois: the `partition` arg: class vs named tuple vs callable vs two separate argos?
        - Francois: can it be hidden under kwargs?
            - Chenkai: No. has to be an public api in a public method
            - Yuefeng: users may want to override it
        - Francois: okay with an additional arg

- Integration with gradient tape API
a. What is proposed: gradients = tape.gradient(y, sharded_var.variables)
b. we could also do: gradients = tape.gradient(y, sharded_var)
    - Chenkai: not a concern for Keras layers users
        - To support b, we need another object to hold the gradients of a sharded_variable, and that object needs to behave like a Tensor.
        - It will error out if a sharded var is passed to the gradient tape.
    - Francois: probably not a big deal
    - Priya: a) is good, nice to support b) in the future transparently

- Expose shardedvar? and saved_model support
    - Priya: we can still make it a public symbol, but not document its constructor
    - Priya: we don’t want users to use it directly, but what about other libraries?
        - Keras, saved_model?
    - Cesar: yes for saved_model. But if we reconstruct shardvar in C++, so the Python symbol is unrelated.
    - Chenkai: serialization/deserialization logic can be reused, etc. when loading we rewrite the function and add parallel lookup ops
    - Allen: complicated, prefer to make a graph agnostic to # shards, avoid changing saved\_model
    - Cesar: Agree. for MirroredVar, some cases not working. Single graph is nice.
    - Chenkai: ShardedVar does some model parallelism, some methods take it as a var, some take it as a list of var
    - Allen: some lowering is needed, some utils like op handler can be used
    - Priya: Out of the scope of this doc, dtensor can probably make it work in the future
    - Allen: yes, dtensor probably will reuse the checkpointing, saving etc as well
    - Chenkai: we won't have a conclusion in this meeting. Will investigate later.

- ShardedVar for other strategies
    - Chenkai: matmul will be different for different strategies. For ps strategy, it can still be different
    - Priya: can we use dispatch mechanism for gpu matmul
        - Chenkai: independent of strategies
    - Andy: will it be an error if you don't define it for a particular op?
        - Chenkai: no for psstrategy because of auto-concat
        -  Andy: autoconcat may cause memory error which is confusing but fallback is important for e.g. matmul

- Gus: will large vocab work?
    - Chenkai: should be able to handle, this design has no implications on scalability other than the scale that TF can handle

- Cesar: for saved_model there are some ideas, no design yet
    - Chenkai: we can have a followup session but this design is independent currently
    - Priya: adding that is an incremental work?
    - Cesar: important, will undo some work
    - Chenkai: training and saving can be made independent
    - Allen: N python objects, concat, N corresponding ops, different for those who use a packed representation
    - Casar: different graphs for training and serving, will undo something for serving, ideally we should have one graph
    - Allen: move to a packed representation to support this, good to think through that
    - Chenkai: if we switch, API won't change, gradient api may change, we can make it backward-compatible
    - Casar: test driven design, write down small piece code for use cases and what user code will look like.
    - Allen: let’s make it experimental and make longer plans for saved\_model, not too many users to low level API
    - Casar: serving restraining needs it

- Chenkai: Follow up for packed representation and serving. Training part shouldn't block 2.4.

ematejska:
@ckkuang What's the status of this RFC? Did you finish the followup items?

ckkuang:
The follow up items are quite involved so it'll require another RFC. The design in this doc reflects what is being implemented in [ParameterServerStrategy](https://www.tensorflow.org/api_docs/python/tf/distribute/experimental/ParameterServerStrategy). In that sense, shall we merge this RFC? I'll let @yuefengz commend on the future work on the followup items.

ematejska (approved):

ematejska:
Followed up with @yuefengz.  This will be merged and followups will be tracked separately.

PR #318, opened 2020-10-27
RFC: Modular TensorFlow Graph C API

ShengYang1:
This RFC will be open for comment until Wednesday, Nov 18th, 2020.

# Modular TensorFlow Graph C API

| **RFC #**     | [318](https://github.com/tensorflow/community/pull/318)|
| **Author(s)** | Yang Sheng (yang.sheng@intel.com), Zhoulong Jiang (zhoulong.jiang@intel.com), Yiqiang Li (yiqiang.li@intel.com),  Eric Lin (eric.lin@intel.com), Jianhui Li (jian.hui.li@intel.com) |
| **Sponsor**   | Eugene Zhulenev (ezhulenev@google.com)                  |
| **Updated**   | 2020-10-27                                           |

## **Objective**

TensorFlow currently provides a C++ API for registering a custom graph optimizer in [Grappler](https://www.tensorflow.org/guide/graph_optimization). This project aims to create a modular/plugin-based TensorFlow implementation with C APIs. Plugins will be able to register custom graph optimizers. Users only need to install the plugin in a specified directory, and the mechanism is able to discover and plug in the capabilities offered by the plugin.

This RFC is based on the Modular TensorFlow [RFC](https://github.com/tensorflow/community/blob/master/rfcs/20190305-modular-tensorflow.md), which aims at extending the TensorFlow design to plug in capabilities like adding a new graph optimizer.


eric-haibin-lin (commented):
Thanks for the proposal. Would this RFC also enable custom XLA pass plugins, or is mainly targeting TF graph?

ShengYang1:
> Thanks for the proposal. Would this RFC also enable custom XLA pass plugins, or is mainly targeting TF graph?

Hi, @eric-haibin-lin , XLA pass is not included in this RFC, it is mainly targeting TF graph.

eric-haibin-lin:
thanks for the reply!

penpornk:
There will be a public design review meeting for this RFC.
Time: Thursday, November 19th, 2020 from 3:00-4:00pm PT

ShengYang1:
@googlebot I fixed it.

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -0,0 +1,527 @@
...
+
+<!-- Here is what plugin authors need to do:
+1. Implementing `P_Optimize` function.
+2. Copying proto files from proper to plugin, implementing `P_MessageToBuffer`/`P_BufferToMessage` functions.
+3. (Optional)Implementing `plugin::util` functions for graph transformation.
+4. Registering the optimizer -->
+
+### Struct/Function/Object Overview
+- Struct
+  - Struct that should be filled by the plugin: `P_OptimizerConfigFns`, `P_RegistrationParams`, `TF_OptimizerBuilder`
```
annarev, line 51: Does `P_` prefix mean that the struct should be filled by the plugin? I am thinking if `TP_` prefix would be better here (that might match what we do for `StreamExecutor` C API better where we have `SP` and `SE` prefixes).

Also, should `TF_OptimizerBuilder` be `(T)P_OptimizerBuilder`?

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -0,0 +1,527 @@
...
+- Object:
+  - Object defined in plugin: [Object](#proto) generated by protobuf
+
+### Usage Overview
+
+* **Graph Optimization function**
+
+  Graph [Optimize](https://github.com/tensorflow/tensorflow/blob/r2.3/tensorflow/core/grappler/optimizers/graph_optimizer.h#L58) function is the main part that plugin authors need to implement. The C API looks like below. Both input and output graphs are represented by serialized `TF_Buffer` objects:
+  ```cpp
+  void P_Optimize(void* optimizer, TF_Buffer* graph_buf, TF_Buffer* optimized_graph_buf, TF_Status* s);
```
annarev, line 68: Would it make sense to have a struct for `optimizer` instead of `void*`? I am wondering if there would be some properties we would want to pass in the future.

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -0,0 +1,527 @@
...
+
+  Proper:
+  ```cpp
+  Status MessageToBuffer(const tensorflow::protobuf::MessageLite& in, TF_Buffer* out);
+  Status BufferToMessage(const TF_Buffer* in, tensorflow::protobuf::MessageLite& out);
+  ```
+
+  Plugin:
+  ```cpp
+  Status P_MessageToBuffer(const plugin::protobuf::MessageLite& in, TF_Buffer* out);
```
annarev, line 101: Just to make sure, `P_MessageToBuffer` and `P_BufferToMessage` are not a part of the API surface, right?

ShengYang1: Yes. I think the name needs to be changed, they are confusing, maybe "plugin::MessageToBuffer" is more clear. I will clarify it in RFC.

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -0,0 +1,527 @@
...
+
+  If pluggable graph optimizer is registered to a device type, e.g., GPU, its optional for plugin authors to provide a recommended configuration indicate whether some of existing optimizers in proper can be turned on/off, by populating flags in `P_RegistrationParams`.
+
+  ```cpp
+  TF_Bool get_remapping() { return false; }
+  TF_Bool get_auto_mixed_precision() { return true; }
+
+  void TF_InitGraphPlugin(P_RegistrationParams* params, TF_Status* status) {
+    // Plugin authors can turn on/off some optimizers.
+    params.config_fns.get_remapping = get_remapping;
```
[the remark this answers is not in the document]
ShengYang1: Each field has three states: on, off and default. If function is not provided, it means plugin authors want to use the default value defined in [rewrite_config.proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/rewriter_config.proto).It's hard to distinguish between off and default using TF_Bool only.

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -0,0 +1,527 @@
...
+  By default, the registered optimizer are turned on, users can turn off them. The registered optimizers would be runnning if they are turned on and the graph device type is matched with registered device type.
+  ```python
+  >> from tensorflow.core.protobuf import rewriter_config_pb2
+  >> config = tf.compat.v1.ConfigProto()
+  >> config.graph_options.rewrite_options.use_plugin_optimizers = rewriter_config_pb2.RewriterConfig.ON
+  ```
+
+* **Configuring existing optimizers**
+
+  If pluggable graph optimizer is registered to a device type, e.g., GPU, its optional for plugin authors to provide a recommended configuration indicate whether some of existing optimizers in proper can be turned on/off, by populating flags in `P_RegistrationParams`.
```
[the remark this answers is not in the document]
ShengYang1: Yes, plugin authors should register the optimizer along with device_type and configuration, and only one optimizer and configuration is allowed for each device type.

rfcs/20201027-modular-tensorflow-graph-c-api.md
ShengYang1: This mechanism is following `Compute` function in kernel C API. Here `void* optimizer` is a struct defined in plugin and created in `P_Create` function, it only contains some members that are used in plugin.

> I am wondering if there would be some properties we would want to pass in the future.
I guess what you referred is `TF_OptimizerBuilder`. Currenty create/delete/optimize is included since these are most useful. We can add more if needed.

rfcs/20201027-modular-tensorflow-graph-c-api.md
ShengYang1: Yes. I have already changed those names.

rfcs/20201027-modular-tensorflow-graph-c-api.md
ShengYang1: I have already changed and updated the RFC.

penpornk:
A friendly reminder that there will be a public design review for this RFC later this after noon (Thursday 11/19/20, 3-4pm PT).
[Meeting link](https://meet.google.com/nky-rahk-xxv)

Out of curiosity, @wchao1115 and @kulinseth, do you plan to attend? Is this of interest to you?

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -0,0 +1,546 @@
...
+  // and potentially a set of nodes to feed.
+  typedef struct TF_GrapplerItem TF_GrapplerItem;
+
+  // Get TF_GrapplerItem from TF_Buffer.
+  TF_GrapplerItem* TF_GetGrapplerItem(TF_Buffer* buffer);
+
+  // Get a set of node names that must be preserved. This includes feed and
+  // fetch nodes, keep_ops, init_ops. Fills in `max_values` and `storage_size`,
+  // they will be used in `TF_GetNodesToPreserveList`
+  void TF_GetNodesToPreserveSize(TF_GrapplerItem* item, int* max_values,
```
annarev, line 278: Should `max_values` be `num_values`? My guess is this function returns an actual value count.

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -0,0 +1,546 @@
...
+  #endif
+
+  // TF_GrapplerItem represents a combination of a graph, one of more fetch nodes,
+  // and potentially a set of nodes to feed.
+  typedef struct TF_GrapplerItem TF_GrapplerItem;
+
+  // Get TF_GrapplerItem from TF_Buffer.
+  TF_GrapplerItem* TF_GetGrapplerItem(TF_Buffer* buffer);
+
+  // Get a set of node names that must be preserved. This includes feed and
```
annarev, line 275: What does `preserved` mean in this context? Does it mean kept in the graph and not removed?
Can you update the comment to add more details?

rfcs/20201027-modular-tensorflow-graph-c-api.md
ShengYang1: Yes, it is an actual value count. I have changed.

rfcs/20201027-modular-tensorflow-graph-c-api.md
ShengYang1: Yes. It means thoes nodes cannot be transformed/removed. I'll add more details.

ezhulenev:
```only one optimizer is allowed to be registered per device type```

What does it mean optimizer per device in practice? Grappler meta-optimizer runs on graphs that could be placed on multiple devices at the same time.

**Comment by ezhulenev** [comment]
/cc @kulinseth and Alex, regarding TF dialect question raised during the meeting. I guess we can do something similar to https://github.com/tensorflow/mlir-hlo to pull TF dialect out of the TF repo, but this will be trickier because it unfortunately has dependencies (not much) on tensorflow core.

You probably can get a more informed answer in this group: https://groups.google.com/a/tensorflow.org/g/mlir

**Comment by penpornk** [comment]
### Design review meeting notes

@annarev: Should structs populated by plug-in be prefixed with `TP_` instead of just `P`, to be consistent with StreamExecutor C API? ([Original comment](https://github.com/tensorflow/community/pull/318#discussion_r524912573))
@ShengYang1: Yes. Changed the names.

@annarev: Should the `void* optimizer` parameter to `P_Optimize` be a struct? ([Original comment](https://github.com/tensorflow/community/pull/318#discussion_r524916772))
@ShengYang1: Yes, it is a struct, following the `Compute` function in the [kernel C API RFC](https://github.com/tensorflow/community/blob/master/rfcs/20190814-kernel-and-op-registration.md).
```c++
// Struct for Optimizer. Plugin authors must provide an optimize function.
// Creation and deletion functions are optional.
typedef struct TP_Optimizer {
  size_t struct_size;
  void* ext;  // reserved for future use
  void* (*create_func)();
  void (*optimize_func)(void*, TF_Buffer*, TF_Buffer*);
  void (*destory_func)(void*);
} TP_Optimizer;
```

@kulinseth: Is the format for ['tf' dialect](https://www.tensorflow.org/mlir/tf_ops) in the buffer TBD?
We have a graph optimizer pass that is already using MLIR. Do we have to go through GraphDef for this?
@aselle: MLIR is out of the scope of this RFC.
@rmlarsen: This is for GraphDef only. MLIR needs a separate effort.
@ematejska: How much migration effort would it take to move to MLIR-based graph optimizer?
@Jianhui-Li: Our optimizer mostly identifies a group of ops to fuse into one large op. We expect to make appropriate changes to match the new interface, but the optimizations should stay the same.
@ematejska: So we don’t know yet.
@aselle: Protobufs are ABI-compatible. In theory, it’s possible to create some abstractions on top of both GraphDef and `tf` dialect, but the benefits from doing so are dubious.
@rmlarsen: Could redefine `TF_Buffer`.
@sanjoy: If you only care about fusing small ops to a bigger op, we could create some other simpler representation as well.
@aselle: We could do this completely descriptively. But it often ended up messy in practice, with op attributes and other things.
@Jianhui-Li: We need a more flexible way. For example, depending on the accelerator type, we might want to offload a bigger chunk of ops.
@sanjoy: Rules should be able to handle attributes too, right?
@aselle: In principle, yes. This is similar to TFLite delegates where we write functions that take a graph, find a particular sequence of ops, and represent them in some other ways. The problem was when delegate writers assumed they could handle something, but did not cover it completely. 
@ematejska: There could be a problem with new attributes too.
@aselle: Yes. Backward compatibility doesn’t prevent this from failing when there is a new attribute.
@penpornk: Optimizers could also want shape inferences too. 

@aselle: Are there any version compatibility requirements for Grappler?
@rmlarsen: The C++ API surface has been stable for years. The graph optimization passes order changes over time. But these have never caused any churn so far.
@penpornk: The area that could have versioning problems is the ops themselves. 
@rmlarsen: Yes, for example, op `foo` has a `v2`, causing Grappler and XLA to no longer work. 
It’s up to the plug-in authors to track and handle this.

@annarev: `P_MessageToBuffer` and `P_BufferToMessage` are not a part of the API surface, correct? ([Original comment](https://github.com/tensorflow/community/pull/318#discussion_r524918095))
@ShengYang1: Yes, they are not part of the API surface.

@annarev: Can the configurations in `TP_OptimizerConfigFns` be `TF_Bool`s instead of functions that return `TF_Bool`?
@ShengYang1: There are three possible values: true, false, and default. `TF_Bool` cannot represent default value.
@sanjoy: Can we just use a tristate?
@rmlarsen: Agreed. Explicit tristates are nicer than functions.
@yiqianglee: We can try that.

@annarev: Would existing optimizer configuration also be per-device-type? ([Original comment](https://github.com/tensorflow/community/pull/318#discussion_r524921379))
@ShengYang1: Yes.

@ezhulenev: What does it mean to have an optimizer per a device in practice? ([Original comment](https://github.com/tensorflow/community/pull/318#issuecomment-730696813))
@ShengYang1: Each PluggableDevice can register its own graph optimizer. We can also register a graph optimizer without registering a PluggableDevice with it. But a PluggableDevice cannot register multiple graph optimizers.
@ezhulenev: So if there are custom passes for “CPU”, “GPU”, and “XPU”, all three passes will be run after the main Grappler passes in a random order?
@yiqianglee: Yes.

@sanjoy: Why do you have to tie a graph optimization pass to a PluggableDevice?
@yiqianglee: Pairing each PluggableDevice with its own optimizer makes it easier to limit the scope of graph transformation to the device.
@sanjoy: Even with them tied together, we’d still need to trust the plug-in to some extent.
@ezhulenev: One benefit with this approach is that you can just skip the pass if you don’t have the corresponding PluggableDevice.
@sanjoy: You could also skip the serialization/deserialization if there are no PluggableDevices loaded. 
@rmlarsen: In multiple `session.run()` calls where some don’t contain plug-in code, we can avoid the serialization.

@ezhulenev: What about optimizer plug-ins with conflicting configurations?
@yiqianglee: This is what we would like to get feedback from here.
@ezhulenev: Remapper only fuses ops placed on CPU and GPU. So it shouldn’t touch anything on those PluggableDevices in practice (unless they are registered as “GPU”).
@rmlarsen: We could do logical AND, i.e., don’t run a pass if at least one plug-in set `enable=false` for that pass.  
@Jianhui-Li: We could also make plug-ins only able to turn a pass off (and not on). To turn on a pass that has previously been disabled, the user needs to make changes on the TensorFlow side.
@rmlarsen: Turning Grappler passes off could have negative performance implications.
@penpornk: How about plug-ins print warnings whenever a pass is disabled?
@Jianhui-Li: Printing could help.
@rmlarsen: Each plug-in should also run a set of performance benchmarks to verify that the pass(es) it is disabling wouldn’t have too much effect.
@penpornk: What if different plug-ins disable different passes? Each of them turns off only one pass and passes the performance benchmarks, but together, they disable too many passes and then everything is slow.
@ezhulenev: We could still run with remapper turned on for non-CPU/GPU device types.
@rmlarsen: Will need to add logic in the [MetaOptimizer](https://github.com/tensorflow/tensorflow/blob/c93279fe1093c78c92e17b49e29af2a6de3493a3/tensorflow/core/grappler/optimizers/meta_optimizer.h#L35).

@kulinseth: Is the order of these plug-in passes towards the end of all Grappler passes?
@ShengYang1: Yes.

@annarev: Should `max_values` in `TF_GetNodesToPreserveSize` be `num_values`? ([Original comment](https://github.com/tensorflow/community/pull/318#discussion_r527114459))
@ShengYang1: Yes.

@annarev: What does `preserved nodes` mean in the context of `TF_GetNodesToPreserveSize`? ([Original comment](https://github.com/tensorflow/community/pull/318#discussion_r527115796)) 
@ShengYang1: It means nodes that cannot be transformed/removed. Will add more details.

@kulinseth: Quick question. For custom ops that are exposed through an op kernel library and are visible in the TensorFlow namespace, would they work with the plug-in passes?
@annarev: Yes. The plug-in should be able to see the op names in the GraphDef.

Alexandre Rames: Maybe this is not the right venue for the question. I’d like to learn more about the visibility of the `tf` dialect. Can plug-ins see it? Or is it only visible in TensorFlow?
For example, if plug-ins want to work on the MLIR `tf` dialect directly. We have partial lowerings to some other dialects.
@kulinseth: This is consistent with what we want.
@ezhulenev: We want to make `tf` dialect visible to plug-ins too. It’s just no one has looked at it yet. And, as of now, I don’t know how it will be done. Could you write an email explaining the requirements? We can start from there.

@ematejska: The RFC should be updated to reflect the conclusions on conflicting plug-in optimizer configurations. After that, we will see if we can conclude the RFC on the PR or if we will need another design review meeting.

kulinseth:
> /cc @kulinseth and Alex, regarding TF dialect question raised during the meeting. I guess we can do something similar to https://github.com/tensorflow/mlir-hlo to pull TF dialect out of the TF repo, but this will be trickier because it unfortunately has dependencies (not much) on tensorflow core.
@ezhulenev Thanks! for the pointer, much appreciated. Looks like this would be a good template to start with. Like you said as we dig into details we will know if the dependencies can be resolved.

> You probably can get a more informed answer in this group: https://groups.google.com/a/tensorflow.org/g/mlir

sanjoy:
CC @joker-eph @jpienaar

ShengYang1:
cc: @penpornk @ezhulenev @kulinseth @annarev @aselle @rmlarsen @ematejska @Jianhui-Li @ematejska @sanjoy
Hi all, 
Here are some updates after the review meeting:
- Update C API comments with more details. ([Original comment](https://github.com/tensorflow/community/pull/318#discussion_r527115796))
- Define a new enum `TF_TriStates` to represent true/false/default value.
- Add some clarifications on conflicting optimizer configurations in a table.

Please review.

mihaimaruseac:
There will be a design meeting for this on 15th of December at 10 am Pacific.

If you want to join, [here](https://meet.google.com/nky-rahk-xxv) is the [Meet link](https://meet.google.com/nky-rahk-xxv)

mihaimaruseac:
> There will be a design meeting for this on 15th of December at 10 am Pacific.
> 
> If you want to join, [here is the Meet link](https://meet.google.com/nky-rahk-xxv)

The above should have been posted to a different RFC. Apologies for the confusion

penpornk (commented):
@ShengYang1 Thank you for the changes and sorry for the delay! I took a quick look and have minor comments. The rest seems fine to me.

penpornk:
Hi all,

To push this RFC forward, I'd like to set a deadline for responses. It has been nine days since the authors made the changes so I think another week is sufficient. 

Please raise your concerns here by next Tuesday, December 8, 2020. If there are no outstanding concerns by then, I think we should wrap this up.

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -169,7 +186,89 @@ Plugin graph optimization is targeting backend device specific optimization, and
...
+    TF_OFFSET_OF_END(TP_OptimizerConfigs, scoped_allocator_optimization)
+
+  // Struct for Optimizer. Plugin authors must provide an optimize function.
+  // Creation and deletion functions are optional.
+  typedef struct TP_Optimizer {
+    size_t struct_size;
+    void* ext;  // reserved for future use
+    void* (*create_func)();
+    void (*optimize_func)(void*, TF_Buffer*, TF_Buffer*);
+    void (*destory_func)(void*);
```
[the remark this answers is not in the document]
ShengYang1: Thanks for your review. All comments have been addressed.

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -102,60 +102,77 @@ When initializing, TensorFlow loads the plugin and registers a new graph optimiz
...
 
-  If user turns on the optimizer, the recommended configuration is automatically applied and printed through the same API. If optimizers are turned off, default configuration is used.
-
-  ```python
-  >> config.graph_options.rewrite_options.use_plugin_optimizers = rewriter_config_pb2.RewriterConfig.ON
-  {'remapping': False, 'auto_mixed_precision': False, ...}
-  ```
-
-  When multiple plugins are registered and running for a graph, an error would be raised if their recommended configurations are different. Users should unload one of the plugins or disable plugin optimizers to resolve the conflict.
+  When multiple plugins are successfully registered and running for a graph, their recommended configurations may differ with each other. If any of the config has turned off the optimizer, the optimizer will be disabled. The following table lists all possible scenarios. In the table, plugin's config is represented by tristates, `ON` means turn on optimizer, `OFF` means turn off optimzier, and `DEFAULT` means uses proper's config.
```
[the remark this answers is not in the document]
ShengYang1: Done

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -102,60 +102,77 @@ When initializing, TensorFlow loads the plugin and registers a new graph optimiz
...
-* **Execution order**
-
-  Custom optimizers will be registered at the end of Grappler’s meta-optimizer. Plugin authors should be aware of the restrictions of their pass running at this specific point in the execution pipeline.
+  | Proper's config | Plugin1's config | Plugin2's config  | Final config | Notes|
+  |:-------------- |:-------------- |:-------------- |:-------------- |:-------------- |
+  | ON    | ON/DEFAULT | ON/DEFAULT | **ON**| The optimizer is enabled. |
+  | ON    | OFF | OFF | **OFF**| The optimizer is disabled, unless users manually unload the plugin. Grappler prints warnings to remind users that config has been changed based on plugin's config.|
+  | ON    | ON/DEFAULT | OFF | **OFF**| The optimizer is disabled if at least one plugin turns off it. Grappler prints warnings to remind users that config has been changed based on plugin's config, and potention performance regression may happen due to the conflict configs.|
+  | ON    | OFF | ON/DEFAULT | **OFF**| Same as previous scenario.|
+  | OFF   | ON/DEFAULT/OFF | ON/DEFAULT/OFF | **OFF**| The optimizer is always disabled when user turns off it. |
```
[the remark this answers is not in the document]
ShengYang1: Done

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -102,60 +102,77 @@ When initializing, TensorFlow loads the plugin and registers a new graph optimiz
...
 
-* **Execution order**
-
-  Custom optimizers will be registered at the end of Grappler’s meta-optimizer. Plugin authors should be aware of the restrictions of their pass running at this specific point in the execution pipeline.
+  | Proper's config | Plugin1's config | Plugin2's config  | Final config | Notes|
+  |:-------------- |:-------------- |:-------------- |:-------------- |:-------------- |
+  | ON    | ON/DEFAULT | ON/DEFAULT | **ON**| The optimizer is enabled. |
+  | ON    | OFF | OFF | **OFF**| The optimizer is disabled, unless users manually unload the plugin. Grappler prints warnings to remind users that config has been changed based on plugin's config.|
+  | ON    | ON/DEFAULT | OFF | **OFF**| The optimizer is disabled if at least one plugin turns off it. Grappler prints warnings to remind users that config has been changed based on plugin's config, and potention performance regression may happen due to the conflict configs.|
```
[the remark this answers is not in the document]
ShengYang1: Done

rfcs/20201027-modular-tensorflow-graph-c-api.md
```diff
@@ -0,0 +1,554 @@
...
+
+### Versioning Strategy and Stability
+
+* **Graph C API**
+
+  Version strategy of Graph C API follows Semantic Versioning 2.0.0 ([semver](http://semver.org/)). Each release version has a format `MAJOR.MINOR.PATCH`, as outlined in [TensorFlow version compatibility](https://www.tensorflow.org/guide/versions#semantic_versioning_20). Struct size is used to track compatibility. More details can be found in [StreamExecutor C API Versioning Strategy RFC](https://github.com/tensorflow/community/blob/master/rfcs/20200612-stream-executor-c-api/C_API_versioning_strategy.md)
+
+* **GraphDef**
+
+  The compatibility of `GraphDef` between plugin and proper is guaranteed by protobuf library.
```
yisitu, line 187:
```suggestion
  The compatibility of `GraphDef` between plugin and proper follows the same compatibility [rules](https://developers.google.com/protocol-buffers/docs/cpptutorial?hl=en#extending-a-protocol-buffer) and [guarantees](https://developers.google.com/protocol-buffers/docs/proto3?hl=en#updating) as protobuf library.
```

ShengYang1: Thanks for your suggestion. Updated.

penpornk:
@ematejska The deadline for additional feedback has passed. Since there is no outstanding concern, should we merge this?

ematejska (approved):

Javier761227:
#

> 

>

wchao1115:
@PatriceVignola

Mookel:
Hi, @ShengYang1 I am not sure whether this is the right place to ask question, but I will appreciate that if you can give me some advice:-).

In your RFC, you mentioned that: 

> **Internal util functions**
> TensorFlow proper provides a series of functions to help modify graphs more conveniently in core/grappler/utils folder. Since creating C APIs for these functions is very messy, they would not be included in C APIs. Plugin authors can manually copy this part into plugin side, or they can write their own util functions.

I found that the utility classes in the core/grapper/utils are quite useful, especially the functions provided in graph_view.h/.cc , but the implementation of these classes depends on components from other part of TensorFlow core libraries, which must use C++ APIs. eg:

- https://github.com/tensorflow/tensorflow/blob/957590ea15cc03ee2e00fc61934647d54836676f/tensorflow/core/grappler/utils/graph_view.h#L29
- https://github.com/tensorflow/tensorflow/blob/957590ea15cc03ee2e00fc61934647d54836676f/tensorflow/core/grappler/utils/graph_view.h#L28


So I think just copying these files into plugin side mabye not enough and my question is: If I want to modify the graph itself and do some pattern matches on the graph to support large set of operators fusion  in the plugin side, are there any other better solutions?

Thanks very much!

ShengYang1:
> Hi, @ShengYang1 I am not sure whether this is the right place to ask question, but I will appreciate that if you can give me some advice:-).
> 
> In your RFC, you mentioned that:
> 
> > **Internal util functions**
> > TensorFlow proper provides a series of functions to help modify graphs more conveniently in core/grappler/utils folder. Since creating C APIs for these functions is very messy, they would not be included in C APIs. Plugin authors can manually copy this part into plugin side, or they can write their own util functions.
> 
> I found that the utility classes in the `core/grappler/utils` are indeed quite useful, especially the functions provided in `graph_view.h/.cc` and `pattern_util.h/.cc`, but the implementation of these classes depends on components from other TensorFlow core libraries, which must also use C++ APIs. For instance:
> 
> * https://github.com/tensorflow/tensorflow/blob/957590ea15cc03ee2e00fc61934647d54836676f/tensorflow/core/grappler/utils/graph_view.h#L29
> * https://github.com/tensorflow/tensorflow/blob/957590ea15cc03ee2e00fc61934647d54836676f/tensorflow/core/grappler/utils/graph_view.h#L28
> 
> So I think just copying these files into plugin side mabye not enough and my question is: If I want to modify the graph itself and do some pattern matches on the graph to support large set of operators fusion in the plugin side, are there any other better solutions?
> 
> Thanks very much!

Thanks for your question. Actually I prefer directly copying all those APIs(`tensor_id.h`, `node_def_util.h`,...) into plugin side, and naming them as `plugin::TensorId`, `plugin::AddNodeAttr`, since they only depends on protobuf objects and basic c++ objects. In such case, your code in plugin will be almost the same as those in proper, except for the namespace :)

Mookel:
> 

Thanks @ShengYang1 , I will try:-).

Mookel:
> Thanks for your question. Actually I prefer directly copying all those APIs(`tensor_id.h`, `node_def_util.h`,...) into plugin side, and naming them as `plugin::TensorId`, `plugin::AddNodeAttr`, since they only depends on protobuf objects and basic c++ objects. In such case, your code in plugin will be almost the same as those in proper, except for the namespace :)

@ShengYang1 Thanks very much for your valuable suggestions , I will try:-).

PR #334, opened 2020-11-23
RFC: Distributed Keras training APIs with parameter servers

rchao:
**This RFC will be open for comment until Friday, December 7th, 2020.**
 
# Distributed Keras training APIs with parameter servers
 
| Status        | Under review                                             |
| **RFC #** | 334                                                                 |
| **Author(s)** | Rick Chao, Tom O'Malley, Zhenyu Tan, Yuefeng Zhou (Google)                                  |
| **Sponsor**   | Francois Chollet, Priya Gupta (Google)                                |
| **Updated**   | 2020-11-21                                        |
 
 
## Goals
 
*   Parameter server training support for Keras compile/fit style training API
*   Minimal code changes across usage with other strategies
*   Minimal performance implications

### DIFF
--- a/rfcs/20201121-keras-model-fit-ps.md
+++ b/rfcs/20201121-keras-model-fit-ps.md
@@ -10,9 +10,11 @@
 
 ## Background
 
-With the recent release of TF2 parameter server training support ([ddoc](https://github.com/tensorflow/community/blob/master/rfcs/20200306-single-client-parameter-server.md)) ([API](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/distribute/parameter_server_strategy_v2.py)) ([tutorial](https://www.tensorflow.org/tutorials/distribute/parameter_server_training)), custom training loop (CTL) users have started using the `ParameterServerStrategy` and `ClusterCoordinator` APIs for parameter server style distributed training. `ParameterServerStrategy` provides implementation of variable placement, and APIs for defining computation, and `ClusterCoordinator` provides APIs for dataset creation, asynchronous function scheduling and remote execution. The asynchronicity brought by `ClusterCoordinator` provides scalability and training fault tolerance, and at the same time implications such as the need for remote resource creation.
+With the recent release of TF2 parameter server training support ([API](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/distribute/parameter_server_strategy_v2.py)) ([tutorial](https://www.tensorflow.org/tutorials/distribute/parameter_server_training)), custom training loop (CTL) users have started using the `tf.distribute.experimental.ParameterServerStrategy` and `tf.distribute.experimental.coordinator.ClusterCoordinator` APIs for parameter server style distributed training. `ParameterServerStrategy` provides implementation of variable placement, and APIs for defining computation, and `ClusterCoordinator` provides APIs for dataset creation, asynchronous function scheduling and remote execution. The asynchronicity brought by `ClusterCoordinator` provides scalability and training fault tolerance, and at the same time implications such as the need for remote resource creation.
 
-While CTL user flow has since been supported, Keras `model.fit` training API is not yet. It has been a common ask (as shown in a survey conducted earlier this year) for availability, given its simplicity and support for a variety of machine learning models, metrics, optimizers, etc. 
+TF2 parameter server training is based on one coordinator task, multiple workers, and multiple (usually fewer than workers) parameter servers (referred to as "ps"). Workers and parameter servers run TensorFlow servers, while the coordinator creates resources on workers and parameter servers, dispatches functions, coordinates the training amd writes checkpoints etc.
+
+While CTL user flow has been supported since the release of TF 2.4, Keras `model.fit` training API is not yet. It has been a common ask (as shown in a survey conducted earlier this year) for availability, given its simplicity and support for a variety of machine learning models, metrics, optimizers, etc. 
 
 In this design, we will discuss the changes in `model.fit` API that we expect to make to accommodate asynchronous, coordinator-based parameter server training flow, and challenges the integration may have given the historical focus of synchronous distributed training with `model.fit`.
 
@@ -24,6 +26,15 @@ In this design, we will discuss the changes in `model.fit` API that we expect to
 *   Parameter server training support for Keras compile/fit style training API
 *   Minimal code changes across usage with other strategies
 *   Minimal performance implications
+
+
+## Glossary
+
+* Parameter server training: a distributed training method utilizing multiple machines to scale up model training, utilizing a set of training workers, and a set of parameter servers that store the training variables.
+
+* Coordinator: A task (referring to a program run on a dedicated machine) where the python program creates variables on parameter servers, defines functions to be executed on remote workers, and controls the training via `ParameterServerStrategy` and `ClusterCoordinator` APIs.
+
+* `model.fit`: A Keras API for running epoch and step based training loops, with user-provided optimizers, metrics, loss, and callbacks etc.
 
 
 ## Proposed options and solutions
@@ -44,13 +55,13 @@ with strategy.scope():
   model = ... # Building a Keras model
 model.compile(optimizer=..., loss=...)  # `ClusterCoordinator` is created
 def dataset_fn(): 
-  ... # Make use of `preproc_stage` for transformation
+  return tf.data.Dataset.X... # Make use of `preproc_stage` for transformation
 history = model.fit(dataset_fn, epochs=..., steps_per_epoch=...,  callbacks=[...])
 logging.info("result: %r", history)
 ```
 
 
-with a dataset:
+with a dataset instance:
 
 
 ```
@@ -60,8 +71,9 @@ with strategy.scope():
   preproc_stage = ... # Some Keras preproc layers
   model = ... # Building a Keras model
 model.compile(optimizer=..., loss=...)  # `ClusterCoordinator` is created
-dataset = tf.data.Dataset.... # Make use of `preproc_stage` for transformation
-history = model.fit(dataset, epochs=..., steps_per_epoch=...,  callbacks=[...])
+dataset = tf.data.Dataset.X... # Make use of `preproc_stage` for transformation
+# model.fit serializes and deserializes dataset onto workers
+history = model.fit(dataset, epochs=..., steps_per_epoch=...,  callbacks=[...]) 
 logging.info("result: %r", history)
 ```
 
@@ -81,7 +93,7 @@ Previous discussion indicates that although an API modification is needed, the s
 
 ##### `dataset_fn` path
 
-In TF2 parameter server training, `ClusterCoordinator` naturally supports a dataset function to be passed in to `create_per_worker_dataset` API, which creates datasets on remote workers. By leveraging such data factory support, `model.fit` with `dataset_fn` can be implemented by subclassing the existing Keras `DataHandler` to provide a worker-distributed dataset for Keras to use (i.e. call `iter` on). Please see `DataHandler` section below for proposed changes.
+In TF2 parameter server training, `ClusterCoordinator` naturally supports a dataset function to be passed in to `create_per_worker_dataset` API, which creates datasets on remote workers. By leveraging such data factory support, `model.fit` with `dataset_fn` can be implemented by subclassing the existing Keras `DataHandler` (a Keras internal private API) to provide a worker-distributed dataset for Keras to use (i.e. call `iter` on). Please see `DataHandler` section below for proposed changes.
 
 In terms of how users pass a dataset factory into `model.fit`, there are a couple of options:
 
@@ -97,6 +109,11 @@ def dataset_fn():
 history = model.fit(dataset_fn, epochs=..., steps_per_epoch=...,  callbacks=[...])
 ```
 
+Pros: 
+* `callable` does not require users to use additional APIs and may be less overhead.
+
+Cons:
+* Less future proof as there could be different intepretation of callable passed as `dataset` to `model.fit` in the future.
 
 
 ###### Option 2: dataset factory
@@ -130,6 +147,12 @@ class DataFactory(object):
     # take any arg.
     return self.x(*args, **kwargs)
 ```
+
+Pros:
+* If `dataset` has a different intepretation, for example it takes an argument instead of none, we get a adapting layer with a `DatasetFactory`.
+
+Cons:
+* This requires users to use an additional symbol.
 
 
 The following discussion is tentatively based on option 1, where a simple callable is taken.
--- a/rfcs/20201121-keras-model-fit-ps.md
+++ b/rfcs/20201121-keras-model-fit-ps.md
@@ -157,8 +157,6 @@ To support the existing `model.fit` API contract where a `dataset` is taken, it 
 
 In parameter server training with a `dataset_fn`, recent work recommends that resources and variables used in the input pipeline are created outside the `dataset_fn`, but within `strategy.scope`, to have variable placement on PS, and resources on the coordinator. This recommendation applies to Keras preprocessing layers (KPL), where read-only resources and variables are created at layer creation. For the `dataset` path, the variables and resources are eagerly created by the user, and they end up having the same placement as `dataset_fn` user path. However, `tf.data` `replicate` API does not support the resources referenced in the dataset graph to be accessed once serialized and deserialized, as opposed to `dataset_fn` case where resources can still be accessed, in the remotely executed `dataset_fn`. This imposes limitations on `dataset` path to use cases where resources are not used.
 
-TODO(yuefengz/tanzheny): Can you add a reference to the latest recommendation once available?
-
 
 ###### Implication on resource creation
 
@@ -172,7 +170,7 @@ In both paths, resources created outside of dataset scope will be placed on the 
 
 ###### Proposed design for `dataset` path
 
-This is to be completed. TODO(tanzheny/rchao/yuefengz)
+This is to be completed.
 
 
 #### The setup of `ClusterCoordinator`
--- a/rfcs/20201121-keras-model-fit-ps.md
+++ b/rfcs/20201121-keras-model-fit-ps.md
@@ -109,7 +109,9 @@ logging.info("result: %r", history)
 There are a couple of points worth noting in the above user code:
 * The `dataset` argument of `model.fit` can no longer be a dataset instance. In fact, in the short term, it most likely will be some form of dataset factory, due to the challenges discussed below.
 * `steps_per_epoch` argument will be required for PS training, at least in the short term. This is because `OutOfRangeError` is raised from `ClusterCoordinator` APIs as soon as one worker exhausts its worker dataset, at which point other workers may have datasets remaining to be processed, and this `OutOfRangeError` indicates neither every dataset is visited roughly once, nor every dataset is visited roughly number of workers times. We thus require explicit steps per epoch, and recommend users to always repeat and shuffle the input dataset.
-* Batch level callback will be disabled when `ParameterServerStrategy` is used; that is, if users override `on_batch_begin` and `on_batch_end`,  an error will be raised.
+* Batch level callback will be disabled when `ParameterServerStrategy` is used; that is, if users override `on_batch_begin` and `on_batch_end`,  an error will be raised. This is necessary for reasonable performance as described below. 
+* The cluster is synced at the end of every epoch. This is an implementation detail users do not necessarily need to be aware of, however is important for the correctness of epoch-level callbacks.
+* `run_eagerly=True` case is not supported. This is because `ClusterCoordinator.schedule` requires a `tf.function` to be `schedule`d, and regular python function cannot.
 
 
 
@@ -270,7 +272,7 @@ This option is with the assumption that there is always only one `ParameterServe
 
 #### Keras `Model` changes
 
-The train function in `Model.make_train_function` can be swapped with a wrapper that takes an iterator (which, when executed on remote workers, would be the worker-specific iterator inside the function being executed), and returns the resulting `RemoteValue`.
+The train function in `Model.make_train_function` can be swapped with a wrapper that takes a `distribute_iterator` (when the scheduled function is executed on remote workers, the function will receive the actual worker-specific iterator inside the function being executed), and returns the resulting `RemoteValue`.
 
 
 ```
@@ -287,8 +289,9 @@ class Model(...):
     self.train_function = ...
 
     if self._cluster_coordinator:
-      self.train_function = lambda iterator: self._cluster_coordinator.schedule(  # pylint: disable=g-long-lambda
-          train_function, args=(iterator,))
+      # Note that `train_function` has to be a `tf.function`.
+      self.train_function = lambda distribute_iterator: self._cluster_coordinator.schedule(  # pylint: disable=g-long-lambda
+          train_function, args=(distribute_iterator,))
 
     return self.train_function
 ```
@@ -348,6 +351,20 @@ class DataHandler(object):
     return logs
 ```
 
+The `DataHandler` `model.fit` uses depends on whether or not it is using a `ClusterCoordinator`:
+
+```
+def get_data_handler(*args, **kwargs):
+  if model._cluster_coordinator:
+    return ClusterCoordinatorDataHandler(*args, **kwargs)
+  return DataHandler(*args, **kwargs)
+```
+
+and `get_data_handler` will replace where we have `DataHandler` instantiation currently:
+
+```
+data_handler = data_adapter.DataHandler(...)
+``` 
 
 A new `DataAdapter` is also needed to make sure the training API knows a `callable`, or a dataset factory is a supported path:
 
@@ -366,7 +383,7 @@ class DataFactoryAdapter(DataAdapter):
 
 #### Multiple steps within a train function
 
-Keras training API has a mechanism to run multiple steps within one `tf.function`’ed train function, which is aimed at less time spent on RPCs between the coordinator and the workers, and thus better training performance. This is specified as the `steps_per_execution` argument in the `model.compile` call. Parameter server training can naturally benefit from this mechanism, without the need of code changes, but it is worth noting that all steps run within a `tf.function` will be executed on the same worker. The major implication of this is possible limitations on callbacks, as explained in the “Callbacks” section below.
+Keras training API has a mechanism to run multiple steps within one `tf.function`’ed train function, which is intended for less time spent on RPCs between the coordinator and the workers, and thus better training performance. This is specified as the `steps_per_execution` argument in the `model.compile` call. Parameter server training can naturally benefit from this mechanism, without the need of code changes, but it is worth noting that all steps run within a `tf.function` will be executed on the same worker. The major implication of this is possible limitations on callbacks, as explained in the “Callbacks” section below.
 
 
 #### Callbacks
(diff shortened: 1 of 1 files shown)


rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+#### `dataset` vs `dataset_fn` argument in `model.fit` API
+
+Current `model.fit` API takes a dataset from which an iterator is created, and the train function is built with this iterator. However, `ClusterCoordinator` only supports taking a no-argument function that returns a `Dataset`. We propose to have both options supported, with an initial focus on `dataset_fn` path due to its lighter implications on `model.fit` implementation.
+
+Previous discussion indicates that although an API modification is needed, the simplicity and less overhead may well justify such callable support, with a goal to support both.
+
+
+##### `dataset_fn` path
+
+In TF2 parameter server training, `ClusterCoordinator` naturally supports a dataset function to be passed in to `create_per_worker_dataset` API, which creates datasets on remote workers. By leveraging such data factory support, `model.fit` with `dataset_fn` can be implemented by subclassing the existing Keras `DataHandler` to provide a worker-distributed dataset for Keras to use (i.e. call `iter` on). Please see `DataHandler` section below for proposed changes.
```
fchollet, line 84: Worth noting for readers that DataHandler is an internal, private API.

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+    # For example, if we allow different argument signature of user-provided
+    # `dataset_fn`, this works as an abstraction layer. 
+    # If we now only allow zero-arg, but later extend it to one-arg (e.g. input
+    # context), we omit the the arg when we observe that the function doesn't
+    # take any arg.
+    return self.x(*args, **kwargs)
+```
+
+
+The following discussion is tentatively based on option 1, where a simple callable is taken.
```
fchollet, line 135: A list of pros/cons for each option would be useful.

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+  model = ... # Building a Keras model
+model.compile(optimizer=..., loss=...)  # `ClusterCoordinator` is created
+def dataset_fn(): 
+  ... # Make use of `preproc_stage` for transformation
+history = model.fit(dataset_fn, epochs=..., steps_per_epoch=...,  callbacks=[...])
+logging.info("result: %r", history)
+```
+
+
+with a dataset:
```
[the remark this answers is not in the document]
fchollet: In this case, the dataset instance is serialized and deserialized on each worker?

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+
+
+###### Implication on no strategy/other strategies
+
+If `model.fit` is allowed to take a `dataset_fn`, use cases for synchronous strategies, and no strategy, can be readily applied. That is, when a dataset is needed, the `callable` is invoked eagerly to obtain a `dataset`, and the remaining workflow will apply.
+
+
+###### Signature of `dataset_fn`
+
+The signature (input argument and return value) of `dataset_fn` taken by `model.fit` should basically follow the signature `ClusterCoordinator.create_per_worker_dataset` takes. There has been discussion around whether that should take an `InputContext` for effective sharding. Current decision is that it does not, since we do not expect users to shard the dataset considering workers can get preempted.
```
fchollet, line 145: Good call. This should not be user-facing

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+    self._cluster_coordinator = None
+
+class ClusterCoodinator(object):
+  def shut_down(self):
+    # Join the threads and terminate resources. We don't have this implemented yet.
+```
+
+
+
+##### Option 2: Have an attribute in `ParameterServerStrategy` that holds the `ClusterCoordinator`
```
fchollet, line 204: +1

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+
+
+```
+def to_numpy(logs):
+  cluster_coordinator.join()  # Sync the workers.
+  return {k: v.fetch() for k, v in logs.items()}  # Return the NumPy results
+```
+
+
+This logic can be handled by the `CallbackList` object, which already handles converting `tf.Tensor` logs to `NumPy`. However, obtaining these logs requires us to sync the workers, which will result in a slowdown if done every batch. Currently we only plan to sync the workers once every epoch at the end of the epoch.
```
fchollet, line 335: We should not have callbacks include PS-specific logic. The task of synchronizing (and deciding when it's appropriate) should be done in the Model.

Note that it is necessary for us to provide a way to synchronize and run callbacks at higher frequency than once per epoch. This is critical. We could use `steps_per_execution` to configure this (with a high default value for the PS use case).

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+
+
+##### Batch-level callbacks
+
+Some users may want to use batch-level `Callback`s. When users use `steps_per_execution=N`, the `Callback`s will only execute every `N` steps, and so batch-level callbacks might not be prohibitively slow for large `N`. However, in most cases, batch-level callbacks will cause a significant slowdown and are likely to be added only in error. We have a few options for handling batch-level Callbacks.
+
+
+###### Option 1: Detect and support batch-level callbacks
+
+We have a mechanism in `CallbackList` to detect when batch-level callbacks are being used. This mechanism will only block asynchronous TPU execution when batch-level callbacks require it. We can use this mechanism to also only block asynchronous PSStrategy execution when batch-level callbacks require it. This would allow us to support batch-level callbacks, without paying a performance penalty in the case where they are not used.
```
fchollet, line 350: This should be done in the Model. The model should initiate the sync every `steps_per_execution` steps, and call the callbacks at that time with a dict of plain values.

Callbacks must be agnostic to PS.

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+
+
+##### Epoch-level callbacks
+
+Since the workers will sync every epoch anyway, fetching the remote values incurs minimal additional overhead and so epoch-level `Callback`s can be easily supported.
+
+
+##### Batch-level callbacks
+
+Some users may want to use batch-level `Callback`s. When users use `steps_per_execution=N`, the `Callback`s will only execute every `N` steps, and so batch-level callbacks might not be prohibitively slow for large `N`. However, in most cases, batch-level callbacks will cause a significant slowdown and are likely to be added only in error. We have a few options for handling batch-level Callbacks.
```
fchollet, line 345: I disagree, batch-level callbacks are useful and important. We just need to make sure that they're called at an appropriate frequency (especially by default)

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+
+
+###### Option 1: Detect and support batch-level callbacks
+
+We have a mechanism in `CallbackList` to detect when batch-level callbacks are being used. This mechanism will only block asynchronous TPU execution when batch-level callbacks require it. We can use this mechanism to also only block asynchronous PSStrategy execution when batch-level callbacks require it. This would allow us to support batch-level callbacks, without paying a performance penalty in the case where they are not used.
+
+
+###### Option 2: Detect, warn, and support batch-level callbacks
```
fchollet, line 353: No to this and the remaining 2 options below.

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+
+
+#### Optimizer variables
+
+Similarly, the hyper and slot variables an `optimizer` object uses, would be created at gradient application, at which point `model.fit` has entered `strategy.scope` for correct placement. For the variables that need to be colocated with other variables, they should work because `ParameterServerStrategy` has entered colocation device scope when these variables are being created.
+
+
+#### model.evaluate and model.predict
+
+Initially, we aim to have `evaluate` and `predict` to only be carried out on the coordinator. That is, it does not involve distribution via a `ClusterCoordinator`.
```
fchollet, line 382: Sounds good.

fchollet (commented):
Thank you for the detailed and well-written RFC!

rfcs/20201121-keras-model-fit-ps.md
yuefengz: It is desirable that some callbacks like checkpointing callbacks will be invoked in the middle of an epoch. But increasing the `steps_per_execution` to a large value has disadvantages: these steps inside a single function call cannot be distributed across workers for load-balancing purposes; worker failure (leading to the interruption of a function execution) will retry all these `steps_per_execution` steps.

Also checkpointing doesn't necessary need synchronization in asynchronous training. IMO it would be useful for Keras to support time-based callback for checkpointing that doesn't need synchronization. The checkpointing, especially in the middle of an epoch, is mainly for failure recovery purposes and therefore time-based callbacks would easily allow users to know that at most X minutes of training would be lost if there is PS failure.

rfcs/20201121-keras-model-fit-ps.md
yuefengz: +1

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+```
+cluster_resolver = ...
+strategy = tf.distribute.experimental.ParameterServerStrategy(cluster_resolver)
+with strategy.scope():
+  preproc_stage = ... # Some Keras preproc layers
+  model = ... # Building a Keras model
+model.compile(optimizer=..., loss=...)  # `ClusterCoordinator` is created
+def dataset_fn(): 
+  ... # Make use of `preproc_stage` for transformation
+history = model.fit(dataset_fn, epochs=..., steps_per_epoch=...,  callbacks=[...])
```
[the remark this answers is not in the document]
yuefengz: Since the `ClusterCoordinator.schedule` can schedule a function to any worker. So it would be less error-prone to just ask users to create the same repeated, but shuffled differently, datasets on different workers. In this case, `steps_per_epoch` is needed.

It is right now difficult to achieve visitation guarantee with `ClusterCoordinator.schedule` but may be easier with `tf.data` service in the future.

yuefengz (commented):
Overall I think we need to distinguish clearly between what is already existent and proposed and what is proposed here.

rfcs/20201121-keras-model-fit-ps.md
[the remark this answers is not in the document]
omalleyt12: Does this mea we should *require* the user to return an infinite dataset in the PS case?

rfcs/20201121-keras-model-fit-ps.md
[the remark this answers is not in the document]
omalleyt12: Yeah, from what I understand though, there are currently issues with trying to serialize and deserialize the Dataset onto each worker, which is why IMO we'll probably have to start with only supporting a "Dataset factory" input type for PS

rfcs/20201121-keras-model-fit-ps.md
omalleyt12: +1 as well, the lifetimes of the PSStrategy and Coordinator seem to be the same, whereas the lifetimes of a Model and a Strategy can be different

rfcs/20201121-keras-model-fit-ps.md
omalleyt12: > We should not have callbacks include PS-specific logic

Agree, really the only change needed is in `tf_utils.to_numpy_or_python_type`, which is used by `CallbackList` when it detects that it needs to convert `Tensor` logs into `NumPy` for a batch. A user would write Callbacks without having to known anything about this, they'll still see NumPy data.

The change would roughly be:

```
def to_numpy_or_python_type(data):
  # Sync workers and get NumPy values
  if isinstance(data, RemoteValue):
    get_strategy().join()
    return data.fetch()
  else:
    # Current logic stays the same for other strategies.
    ...
```

rfcs/20201121-keras-model-fit-ps.md
omalleyt12: In addition to the `steps_per_execution` logic in `Model`, the `CallbackList` object has this mechanism so that it only blocks when absolutely necessary. This saves a lot of time for TPUs. This is complimentary to `steps_per_execution`

But we can re-use this logic while still fully supporting batch-level Callbacks. This logic is just so that, for the users who *don't* pass batch-level Callbacks, we don't end up unnecessarily slowing down training

For users who pass batch-level Callbacks, the `CallbackList` object can detect this and pass the `logs` every batch, with the modifications to `tf_utils.to_numpy_or_python_type` noted above.

rfcs/20201121-keras-model-fit-ps.md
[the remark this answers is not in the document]
yuefengz: Yes, we have explicitly documented the requirement: https://www.tensorflow.org/tutorials/distribute/parameter_server_training#more_about_dataset_creation

There is not error checking in `ParameterServerStrategy` now, but would be nice to add it.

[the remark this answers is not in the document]
rchao: With the current design of parameter server training APIs, error handling allows OutOfRangeError to be reported to user, but the caveat is that it can mean only one worker has exhausted the dataset, while others still have some remaining to be processed. The behavior may likely be different from what the users would expect, especially when dataset is sharded, so the recommendation is to provide an infinite dataset that is shuffled with a different seed.

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+| **Sponsor**   | Francois Chollet, Priya Gupta (Google)                                |
+| **Updated**   | 2020-11-21                                        |
+
+
+## Background
+
+With the recent release of TF2 parameter server training support ([ddoc](https://github.com/tensorflow/community/blob/master/rfcs/20200306-single-client-parameter-server.md)) ([API](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/distribute/parameter_server_strategy_v2.py)) ([tutorial](https://www.tensorflow.org/tutorials/distribute/parameter_server_training)), custom training loop (CTL) users have started using the `ParameterServerStrategy` and `ClusterCoordinator` APIs for parameter server style distributed training. `ParameterServerStrategy` provides implementation of variable placement, and APIs for defining computation, and `ClusterCoordinator` provides APIs for dataset creation, asynchronous function scheduling and remote execution. The asynchronicity brought by `ClusterCoordinator` provides scalability and training fault tolerance, and at the same time implications such as the need for remote resource creation.
```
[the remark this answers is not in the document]
rchao: Sounds good. Added a "Glossary" section.

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+
+```
+cluster_resolver = ...
+strategy = tf.distribute.experimental.ParameterServerStrategy(cluster_resolver)
+with strategy.scope():
+  preproc_stage = ... # Some Keras preproc layers
+  model = ... # Building a Keras model
+model.compile(optimizer=..., loss=...)  # `ClusterCoordinator` is created
+def dataset_fn(): 
+  ... # Make use of `preproc_stage` for transformation
```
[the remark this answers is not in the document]
rchao: Added.

rfcs/20201121-keras-model-fit-ps.md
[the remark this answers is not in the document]
rchao: Done, and I'm thinking the same @omalleyt12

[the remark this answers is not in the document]
rchao: The dataset instance path would require the library to serialize and deserialize the dataset on workers.

rfcs/20201121-keras-model-fit-ps.md
rchao: Added a note here.

rfcs/20201121-keras-model-fit-ps.md
rchao: Added lists for two options.

rfcs/20201121-keras-model-fit-ps.md
rchao: Correct

rfcs/20201121-keras-model-fit-ps.md
rchao: Sounds good. Thanks for the input

rfcs/20201121-keras-model-fit-ps.md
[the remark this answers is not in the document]
omalleyt12: SGTM, IMO then in the DataAdapter class for the DatasetFactory, we should add an error if the dataset isn't infinite that explains this

rfcs/20201121-keras-model-fit-ps.md
rchao: Re: The task of synchronizing (and deciding when it's appropriate) should be done in the Model, one option is having `logs` resolved in `model.fit`, after the steps are done, and before epoch_end callbacks are called. Concretely, in the current code, we do:

```
      for epoch, iterator in data_handler.enumerate_epochs():
        ...
        with data_handler.catch_stop_iteration():  # This will sync
          for step in data_handler.steps():
              tmp_logs = self.train_function(iterator)
              logs = tmp_logs
              ...
        logs = data_handler.resolve_logs(logs)  # convert remote to concrete
        if logs is None:
          raise ValueError('Expect x to be a non-empty array or dataset.')
        epoch_logs = copy.copy(logs)

        # Run validation.
        ...

        callbacks.on_epoch_end(epoch, epoch_logs)  # epoch_logs are concrete values
```

With this, we would just need to provide an implementation for `resolve_logs`. In the case with `ClusterCoordinatorDataHandler`, it would be `logs.fetch()`, whereas the default is an no-op. I believe this could avoid a change in `tf_utils.to_numpy_or_python_type`.

omalleyt12: I think it's OK to have the logic in `tf_utils.to_numpy_or_python_type` (this utility is basically already just "the thing that takes values returned by Model methods and converts them to what the user should see"

IIUC, Francois's concern was that (please correct me if I'm wrong Francois) users who write `Callback`s shouldn't have to be aware of any of this, shouldn't have to know what a `RemoteValue` is, etc. 

As long as we achieve that, either `data_handler.resolve_logs` or `tf_utils.to_numpy_or_python_type` seems fine to me, with a slight preference for `tf_utils` because it consolidates all of our `logs` munging into one place

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,508 @@
...
+
+
+*   Task preemption or unavailability is handled by the program or cluster, without being reported as failure
+*   The training progress can withstand task preemption and subsequent recovery 
+
+The first goal is fulfilled by the failure handling mechanism provided by the `ClusterCoordinator` library, where worker unavailability is handled within the program, and PS/coordinator unavailability is handled by the cluster’s auto-restart mechanism (by exiting with an exit code that the cluster management recognizes). From the user's perspective, they can leverage proposed `handle_restarable_error` context manager:
+
+
+```
+with handle_restartable_error():
```
yuefengz, line 473: Is this API part of the proposal?

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,501 @@
...
+cluster_resolver = ...
+strategy = tf.distribute.experimental.ParameterServerStrategy(cluster_resolver)
+with strategy.scope():
+  preproc_stage = ... # Some Keras preproc layers
+  model = ... # Building a Keras model
+model.compile(optimizer=..., loss=...)  # `ClusterCoordinator` is created
+def dataset_fn(): 
+  ... # Make use of `preproc_stage` for transformation
+history = model.fit(dataset_fn, epochs=..., steps_per_epoch=...,  callbacks=[...])
+logging.info("result: %r", history)
```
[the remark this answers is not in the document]
tanzhenyu: @fchollet It's important that we discuss what is preferred solution here, i.e., dataset, or dataset_fn

rfcs/20201121-keras-model-fit-ps.md
tanzhenyu: +1

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,508 @@
...
+
+In addition to the built-in evaluation `model.fit` provides, sidecar evaluation is also supported with the `SidecarEvaluator` [API](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/keras/distribute/sidecar_evaluator.py). User is expected to kick start an additional task `evaluator`, in which the python program runs a `SidecarEvaluator` as follows:
+
+
+```
+model = tf.keras.models.Sequential(...)
+model.compile(metrics=tf.keras.metrics.SparseCategoricalAccuracy(...)
+data = tf.data.Dataset.from_tensor_slices(...)
+
+SidecarEvaluator(
```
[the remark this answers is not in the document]
tanzhenyu: I think it's here: "Initially, we aim to have `evaluate` and `predict` to only be carried out on the coordinator"

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,558 @@
...
+
+If we go with Option (1), we should disallow batch-level `Callback`s, since in this case `ParameterServerStrategy` with batch-level `Callback`s will always be slower than training on a single machine.
+
+If we go with Option (2) we should support batch-level `Callback`s, but we will use existing logic in `CallbackList` to detect when batch-level `Callback`s are passed, and only incur the performance penalty of syncing workers each batch when the user has passed batch-level `Callback`s (for context, none of Keras's built-in `Callbacks` other than the `ProgressBar` will incur this penalty). This logic was originally put in place to ensure that TPU async mode was only blocked when needed, and applies equally well to `ParameterServerStrategy` without significant modifications. We will also re-use existing logic to log a warning to the user when their batch-level `Callback`s are causing a significant slowdown in training time. This logic also resides in the `CallbackList` object.
```
rchao, line 408: Question @omalleyt12 - does TPU have ProgressBar by default? It looks like it blocks async when verbose=1, which seems to be the default.

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,558 @@
...
+- If a user passes a `Dataset` with `D` batches to each worker, they might expect that setting `steps_per_epoch=D` will visit every batch of the `Dataset`. However, only the first `D/W` batches will be visited with this mental model.
+
+**Option 2: A `step` is one batch on every worker**
+
+With this mental model, every time `Model.train_function` is called, it schedules one batch to execute on each worker. This means that if there are `W` workers, passing `steps_per_epoch=100` will actually run `100 * W` batches of training, with each worker seeing `100` batches.
+
```
yuefengz, line 397: This option would make the total steps dependent on the number of workers, which is not desirable from users' point of view. It is at odds with the programming model provided by the CTL + PSStrategy. The proposed mental model would also break down if we support elasticity in the future. 

In this option, the name of the `steps_per_epoch` argument should be changed to `steps_per_epoch_per_worker`. 

Alternatively, you can sync every `W` steps only run `steps_per_epoch/W` steps, effectively turning it into synchronous parameter server training.

rfcs/20201121-keras-model-fit-ps.md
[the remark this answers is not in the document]
yuefengz: @tanzhenyu you mean this is not part of the proposal? Then we should make it clear here.

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,558 @@
...
+If we wanted to support batch-level `Callback`s with the `step` mental model Option 1 above, we could alternatively introduce a new concept to `Model.fit` of "asynchronous callbacks". An asynchronous `Callback` wouldn't block until its results were ready, instead we would schedule all batches upfront, and then send the results of the batches to the asynchronous `Callback` when the `Coordinator` determines that the batch has finished executing.
+
+**Pros:**
+
+- This allows for a form of batch-level `Callback`s that doesn't need to block. It would work with either mental model of `steps` outlined above. It would work well for certain tasks such as `Checkpointing`.
+
+**Cons:**
+
+- Asynchronous `Callback`s would be limited in what they could do. Any changes that an asynchronous object makes to a `tf.Variable` (such as the `learning_rate`) would not take effect until the next epoch, since all of the batches were already scheduled before the `Callback` executes.
```
rchao, line 420: I think some of the scheduled functions in the same epoch can still pick up the updated variables because the variable read happens at function execution, but not at scheduling time (unless scheduled functions execute super quickly and callback's variable updates are super slow).

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -0,0 +1,558 @@
...
+- This would require a separate code path in `Model.fit`, since the order in which functions are scheduled and `Callback`s are executed would be different in this approach.
+- It's not clear how we should handle it when a user passes a mix of synchronous and asynchronous `Callback`s (for instance, if the user passes in one of our existing built-in `Callback`s in addition to an asynchronous `Callback`).
+
+Asynchronous `Callback`s might be worth exploring in a future extension to the functionality of `Model.fit` + `ParameterServerStrategy` integration, but should likely be out-of-scope for the initial design.
```
rchao, line 424: Agreeing with this.

rfcs/20201121-keras-model-fit-ps.md
```diff
@@ -118,16 +118,25 @@ There are a couple of points worth noting in the above user code:
...
+
+*The idea behind a no-argument function is that the workers are deemed the same, and thus the datasets should be the same on every worker. At this time, we do not recommend sharding.
 
-In TF2 parameter server training, `ClusterCoordinator` naturally supports a dataset function to be passed in to `create_per_worker_dataset` API, which creates datasets on remote workers. By leveraging such data factory support, `model.fit` with `dataset_fn` can be implemented by subclassing the existing Keras `DataHandler` (a Keras internal private API) to provide a worker-distributed dataset for Keras to use (i.e. call `iter` on). Please see `DataHandler` section below for proposed changes.
+**The rationale behind using a `dataset_fn` as opposed to `dataset` was a historical choice as we could not get sharding to work well with fault tolerance.
```
byronyi, line 137: Is it still a hard requirement if the data comes from tf.data service processed in distributed_epoch mode?

rfcs/20201121-keras-model-fit-ps.md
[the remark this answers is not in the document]
rchao: Yes - added this in DataAdapter class part.

rfcs/20201121-keras-model-fit-ps.md
rchao: Sounds good Tom - and thanks for updating the callback section which reflects the current consensus.

rfcs/20201121-keras-model-fit-ps.md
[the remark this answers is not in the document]
rchao: SidecarEvaluator API will be covered in a separate RFC (and is just briefly mentioned in this RFC).

rfcs/20201121-keras-model-fit-ps.md
rchao: Updated this part. We expect users to use try-except.

rfcs/20201121-keras-model-fit-ps.md
rchao: I think having tf.data service support would be ideal; workers then would just fetch the next example without having to worry that OutOfRangeError occurs. There would not be sharding needed either.

ematejska:
@rchao What's the status of this RFC?

ematejska:
This has been accepted.

ematejska (approved):

PR #335, opened 2020-12-01
Cpp gradients

saxenasaurabh:
This RFC will be open for comment until Wednesday, December 16th, 2020.

cc: @saxenasaurabh @rohan100jain

# C++ Gradients 

| **RFC #**     | |
| **Author(s)** | Saurabh Saxena (Google) |
| **Sponsor**   | Rohan Jain (Google)                 |
| **Updated**   | 2020-12-01                                           |

## Objective

We propose performing gradient computation entirely in C++. This aligns with TensorFlow Core team’s vision of providing first-class C++ APIs for building ML models. We mainly focus on reverse-mode autodiff here and leave forward-mode AD as future work. A C API for this implementation is also left as future work but we imagine it to be a straightforward wrapper over the provided C++ APIs.

### DIFF
--- a/rfcs/20201201-cpp-gradients.md
+++ b/rfcs/20201201-cpp-gradients.md
@@ -1,0 +1,8 @@
+# C++ Gradients 
+
+| Status        | Proposed       |
+:-------------- |:---------------------------------------------------- |
+| **RFC #**     | |
+| **Author(s)** | Saurabh Saxena (Google) |
+| **Sponsor**   | Rohan Jain (Google)                 |
+| **Updated**   | 2020-12-01                                           |
--- a/rfcs/20201201-cpp-gradients.md
+++ b/rfcs/20201201-cpp-gradients.md
@@ -6,3 +6,410 @@
 | **Author(s)** | Saurabh Saxena (Google) |
 | **Sponsor**   | Rohan Jain (Google)                 |
 | **Updated**   | 2020-12-01                                           |
+
+
+## Objective
+
+We propose performing gradient computation entirely in C++. This aligns with TensorFlow Core team’s vision of providing first-class C++ APIs for building ML models. We mainly focus on reverse-mode autodiff here and leave forward-mode AD as future work. A C API for this implementation is also left as future work but we imagine it to be a straightforward wrapper over the provided C++ APIs.
+
+
+## Motivation
+
+**Performance**: Running op gradient functions in Python means we incur python overheads for a transformation pass that could entirely be done in C/C++. This most severely affects eager mode training of compute-light pipelines as we show in the performance impact section.
+
+**Support C++ training**: Lack of autodiff support in C/C++ could be detrimental to high performance training pipelines at Google scale. This has often come up in discussions internally.
+
+**Autodiff in non-python TF: **Relying on python gradient functions means we cannot support eager-mode autodiff from other language front-ends such as Java.
+
+In addition, we try to address some shortcomings of the current GradientTape design:
+
+
+
+1. The tape is currently in-charge of filling zeros for missing incoming gradients so that the gradient functions do not need any None-grad checks. This is sub-optimal since the tape now needs to keep around additional metadata(or actual tensors) from the forward pass around in order to be able to create these zeros. To prevent excessive memory usage we have a [hand-curated](https://github.com/tensorflow/tensorflow/blob/dfe6a8ea3725f57be1eaffc2d38c55c18cbd287a/tensorflow/c/eager/tape.h#L630) list of ops with indices of incoming gradients that are allowed to be None. This list is highly likely incomplete and is hard to keep up to date. We address this by moving this logic from the tape to individual gradient functions. Since most ops have a single output this is usually a no-op. 
+2. We have a similar problem with inputs/outputs of the forward op. Since the tape does not know which of the op’s inputs/outputs the gradient function might require, it will need to [keep all tensors](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/python/eager/pywrap_tfe_src.cc;l=3196;drc=68d2777a170c0448ba9d853ef0f7327bdba1df2f) around which can be prohibitive. To limit the memory usage here we [auto-generate](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/python/eager/pywrap_gradient_exclusions.cc) a list of ops with indices of inputs/outputs that can be dropped by parsing ASTs of gradient functions. This again indicates that this should be handled on a per-op level. We address this by allowing individual gradient function instances to incref needed tensors from the forward pass.
+
+
+## Design Overview
+
+The gradients infrastructure will be built on top of the abstract interfaces for op execution which provide a backend agnostic way of tracing and executing ops. We provide APIs for authoring GradientFunctions and registering them into a GradientRegistry for name based lookup. We provide a gradient Tape API that is close to python’s tf.GradientTape and shares most of the implementation with the existing tape.
+
+
+## Detailed Design
+
+
+### APIs
+
+
+#### GradientFunction {#gradientfunction}
+
+An op’s gradient is defined by subclassing GradientFunction
+
+
+```
+class GradientFunction {
+ public:
+  virtual Status Compute(
+      AbstractContext* ctx,
+      absl::Span<AbstractTensorHandle* const> grad_outputs,
+      absl::Span<AbstractTensorHandle*> grad_inputs) = 0;
+  virtual ~GradientFunction() {}
+};
+```
+
+
+GradientFunction::Compute receives gradients wrt op’s outputs in grad\_outputs and needs to populate gradients wrt op’s inputs in grad\_inputs. This is the same signature we use for authoring python gradients with the addition of an AbstractContext, which provides an API creating operations (eagerly or traced). In python this context is stored in a global variable and is implicitly captured. For the C++ API we chose to pass this context explicitly.
+
+The reason GradientFunction is a class and not a callable is so that each op’s gradient function can keep the necessary state needed from forward pass for the gradient computation (see ExpGradientFunction below for an example).
+
+Examples:
+
+
+```
+class AddGradientFunction : public GradientFunction {
+ public:
+  Status Compute(AbstractContext* ctx,
+                 absl::Span<AbstractTensorHandle* const> grad_inputs,
+                 absl::Span<AbstractTensorHandle*> grad_outputs) override {
+    // Tape never calls a gradient function if there are no incoming grads.
+    DCHECK(grad_inputs[0]);
+    grad_outputs[0] = grad_inputs[0];
+    grad_outputs[1] = grad_inputs[0];
+
+    grad_outputs[0]->Ref();
+    grad_outputs[1]->Ref();
+    return Status::OK();
+  }
+  ~AddGradientFunction() override {}
+};
+
+class ExpGradientFunction : public GradientFunction {
+ public:
+  explicit ExpGradientFunction(AbstractTensorHandle* exp) : exp_(exp) {
+    exp->Ref();
+  }
+  Status ExpGradientFunction::Compute(
+      AbstractContext* ctx, absl::Span<AbstractTensorHandle* const> grad_inputs,
+      absl::Span<AbstractTensorHandle*> grad_outputs) {
+    vector<AbstractTensorHandle*> conj_outputs(1);
+    TF_RETURN_IF_ERROR(
+        Conj(ctx, {exp_.get()}, absl::MakeSpan(conj_outputs), "Conj_Exp_Grad"));
+    AbstractTensorHandlePtr conj_output_releaser(conj_outputs[0]);
+
+    TF_RETURN_IF_ERROR(
+        Mul(ctx, {conj_outputs[0], grad_inputs[0]}, grad_outputs, "Mul_Exp_Grad"));
+    return Status::OK();
+  }
+
+  ~ExpGradientFunction() override {}
+
+ private:
+  AbstractTensorHandlePtr exp_;
+};
+```
+
+
+**C++ gen\_ops**
+
+Authoring gradient functions requires calling elementary ops in C++. Using low level op building APIs everywhere would be very verbose so we plan to auto-generate C++ APIs for op-building using registered OpDefs, similar to how we do for python. This is part of a larger TensorFlow core team’s effort to provide efficient python bindings which are light wrappers for C++ API calls. For now we have been handwriting [C++ op building functions](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/c/experimental/ops/) using the unified execution APIs. Once ready we will replace these with auto-generated C++ op functions.
+
+
+#### GradientRegistry
+
+We provide a registry to store the mapping from op type to factory functions that return the GradientFunction for an op’s instance. The factory function takes as input the ForwardOperation, which contains metadata from the forward operation, and returns a GradientFunction. This allows gradient function authors to control which inputs/outputs of the forward op to keep around by increasing the ref-count on AbstractTensorHandle.
+
+Additionally, we provide a utility function RegisterNotDifferentiable to mark an op as non-differentiable. This can be used to implement tf.no\_gradient. We also provide a NotDifferentiableGradientFunction which returns nullptr output gradients. This can be used to implement tf.stop\_gradient.
+
+
+```
+struct ForwardOperation {
+ public:
+  string op_name;
+  std::vector<AbstractTensorHandle*> inputs;
+  std::vector<AbstractTensorHandle*> outputs;
+  std::vector<int64> skip_input_indices;
+  AttrBuilder attrs;
+};
+
+using GradientFunctionFactory = std::function<GradientFunction*(const ForwardOperation& op)>;
+
+// Map from op name to a GradientFunctionFactory.
+class GradientRegistry {
+ public:
+  Status Register(const string& op,
+                  GradientFunctionFactory gradient_function_factory);
+  Status Lookup(const ForwardOperation& op,
+                std::unique_ptr<GradientFunction>* gradient_function) const;
+
+ private:
+  absl::flat_hash_map<string, GradientFunctionFactory> registry_;
+};
+
+class NotDifferentiableGradientFunction : public GradientFunction {
+  Status Compute(
+      AbstractContext* ctx,
+      absl::Span<AbstractTensorHandle* const> grad_outputs,
+      absl::Span<AbstractTensorHandle*> grad_inputs) override;
+};
+Status RegisterNotDifferentiable(GradientRegistry* registry, const string& op);
+```
+
+
+Examples:
+
+
+```
+GradientFunction* AddRegisterer(const ForwardOperation& op) {
+  return new AddGradientFunction;
+}
+
+GradientFunction* ExpRegisterer(const ForwardOperation& op) {
+  return new ExpGradientFunction(op.outputs[0]);
+}
+```
+
+
+
+#### Tape
+
+The API for C++ Tape is very similar to python’s tf.GradientTape. The implementation for this interface is almost entirely shared with the C++ tape in c/eager/tape.h. 
+
+
+```
+class Tape {
+ public:
+  explicit Tape::Tape(bool is_persistent);
+
+  // Adds this tensor to the list of watched tensors.
+  //
+  // This is a no-op if the tensor is already being watched either from an
+  // earlier call to GradientTape::Watch or being an output of an op with
+  // watched inputs.
+  void Watch(const AbstractTensorHandle* tensor);
+
+  // Records an operation with given inputs and outputs
+  // on the tape and marks all its outputs as watched if at
+  // least one input of the op is watched and has a trainable dtype.
+  // op_name is optional and is used for debugging only.
+  void RecordOperation(absl::Span<AbstractTensorHandle* const> inputs,
+                       absl::Span<AbstractTensorHandle* const> outputs,
+                       GradientFunction* gradient_function,
+                       const string& op_name = "");
+
+  // Returns whether any tensor in a list of tensors is being watched and has
+  // a trainable dtype.
+  bool ShouldRecord(absl::Span<const AbstractTensorHandle* const> tensors) const;
+
+  // Unwatches this tensor on the tape. Mainly used for cleanup when deleting
+  // eager tensors.
+  void DeleteTrace(const AbstractTensorHandle*);
+
+  // Computes the gradient and stores the result in result.
+  // Raises an error if result is not the same length as sources.
+  Status ComputeGradient(
+      AbstractContext* ctx, absl::Span<AbstractTensorHandle* const> targets,
+      absl::Span<AbstractTensorHandle* const> sources,
+      absl::Span<AbstractTensorHandle* const> output_gradients,
+      absl::Span<AbstractTensorHandle*> result);
+
+  // Returns whether the tape is persistent, i.e., whether the tape will hold
+  // onto its internal state after a call to ComputeGradient.
+  bool IsPersistent() const;
+};
+```
+
+
+Example:
+
+
+```
+// Computes
+// y = tf.exp(inputs[0])
+// outputs = grad(y, inputs)
+Status ExpGradModel(AbstractContext* ctx,
+                    absl::Span<AbstractTensorHandle* const> inputs,
+                    absl::Span<AbstractTensorHandle*> outputs,
+                    const GradientRegistry& registry) {
+  Tape tape(/*persistent=*/false);
+  tape.Watch(inputs[0]); 
+  std::vector<AbstractTensorHandle*> exp_outputs(1);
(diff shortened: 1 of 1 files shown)


bhack:
Just a side note out of curiosity. 
Is there anyone involved in this RFC who also collaborates  on https://wg21.link/P2072?

saxenasaurabh:
> Just a side note out of curiosity.
> Is there anyone involved in this RFC who also collaborates on https://wg21.link/P2072?

I don't think so. But looks like a great paper! From quickly skimming the paper they seem to be proposing language level autodiff support in C++, similar to Swift, as opposed to framework level. They have some claims about control flow gradients being tricky for framework which is not the case with TF. Nonetheless, it may be worth reaching out to them. Let me know if you have suggestions on who to contact.

cc: @mdanatg @liufengdb

bhack:
> > Just a side note out of curiosity.
> > Is there anyone involved in this RFC who also collaborates on https://wg21.link/P2072?
> 
> I don't think so. But looks like a great paper! From quickly skimming the paper they seem to be proposing language level autodiff support in C++, similar to Swift, as opposed to framework level. They have some claims about control flow gradients being tricky for frameworks which is not the case with TF. Nonetheless, it may be worth reaching out to them. Let me know if you have suggestions on who to contact.
> 
> cc: @mdanatg @liufengdb

I don't know if @mfoco is still active on this in ISOCPP SG19 cc: @fraggamuffin

fraggamuffin:
Yes, we are still working on this, though progress has slowed due to covid since the Prague meeting where this paper was presented. Currently SG19 ML C++ group is working on 3 main proposals with the following schedule:
    Dec 10, 2020: 1800 UTC: stats paper
    Jan 14, 2021: 1800 UTC: differential calculus + reinforcement learning
    Feb 11, 2021: 1800 UTC: Graph paper

You can subscribe to this to get notification of future meeting schedules and call info.. This group, unlike ISO C++ is an outreach group and is open without needing ISO Working Group expert status.
https://lists.isocpp.org/mailman/listinfo.cgi/sg19

saxenasaurabh:
> Yes, we are still working on this, though progress has slowed due to covid since the Prague meeting where this paper was presented. Currently SG19 ML C++ group is working on 3 main proposals with the following schedule:
> Dec 10, 2020: 1800 UTC: stats paper
> Jan 14, 2021: 1800 UTC: differential calculus + reinforcement learning
> Feb 11, 2021: 1800 UTC: Graph paper
> 
> You can subscribe to this to get notification of future meeting schedules and call info.. This group, unlike ISO C++ is an outreach group and is open without needing ISO Working Group expert status.
> https://lists.isocpp.org/mailman/listinfo.cgi/sg19

Thanks @fraggamuffin, I would be excited to hear updates on this. Subscribed.

mfoco:
> > > Just a side note out of curiosity.
> > > Is there anyone involved in this RFC who also collaborates on https://wg21.link/P2072?
> > 
> > 
> > I don't think so. But looks like a great paper! From quickly skimming the paper they seem to be proposing language level autodiff support in C++, similar to Swift, as opposed to framework level. They have some claims about control flow gradients being tricky for frameworks which is not the case with TF. Nonetheless, it may be worth reaching out to them. Let me know if you have suggestions on who to contact.
> > cc: @mdanatg @liufengdb
> 
> I don't know if @mfoco is still active on this in ISOCPP SG19 cc: @fraggamuffin

Thank you for reaching out!
Sure I'm interested. Currently we're having regular meetings with @vgvassilev collecting ideas for a R1 of P2072. Mainly we're collecting use cases (and seems like you have a good one, even if it's lacking one of the big pushes to promote a language change, as @saxenasaurabh noted already).

rfcs/20201201-cpp-gradients.md
```diff
@@ -0,0 +1,415 @@
...
+
+
+#### tf.function and functional control flow gradients [out of scope for now]
+
+Eventually we plan to implement tf.function and functional control flow gradients in C++ but for now we will leave them in python and directly call the python gradient function.
+
+
+#### IndexedSlices
+
+Gradient function of a [gather](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/python/ops/array_grad.py;l=582;drc=d724cdbce69862cbb80617dd6573baa83bd3e819) returns `IndexedSlices` for efficiency. We need to support `IndexedSlices` as part of the input and output gradients of a gradient function. Currently there is no good C++ representation for these. One possible representation would be to wrap the component tensors in an `IndexedSlicesTensorHandle` that subclasses `AbstractTensorHandle`. This way IndexedSlices would be transparent to the tape. The C++ gen ops can choose to handle `IndexedSlices` appropriately or simply densify them by calling a C++ equivalent of `convert_to_tensor`.
```
tomerk, line 343: Is this related to the FullType stuff I saw floating around elsewhere? (If we had a C++ representation for CompositeTensors/Extension types we could possibly just return those)

rfcs/20201201-cpp-gradients.md
```diff
@@ -0,0 +1,415 @@
...
+  // This is a no-op if the tensor is already being watched either from an
+  // earlier call to GradientTape::Watch or being an output of an op with
+  // watched inputs.
+  void Watch(const AbstractTensorHandle* tensor);
+
+  // Records an operation with given inputs and outputs
+  // on the tape and marks all its outputs as watched if at
+  // least one input of the op is watched and has a trainable dtype.
+  // op_name is optional and is used for debugging only.
+  void RecordOperation(absl::Span<AbstractTensorHandle* const> inputs,
```
tomerk, line 193: In python our execution path automatically calls `recordoperation` as needed. When users are looking to have C++ only training pipelines, are they looking to manually define their model in terms of raw C++ ops? And if so, would they have to manually call `RecordOperation` for each op?

tomerk: Ah, I see you mention this in the `#### Automatic variable tracking` section below.

saxenasaurabh: That's a great question! I am glad you brought this up. Another part of the C++ API that I am working on, and that is briefly mentioned in this doc, are the multi-backend [op execution interfaces](https://cs.opensource.google/search?q=f:abstract_(context%7Coperation%7Ctensor_handle).h&ss=tensorflow%2Ftensorflow:tensorflow%2F) for implementing this earlier [RFC](https://github.com/tensorflow/community/blob/master/rfcs/20191203-single-eager-graph-path.md). I imagine there being a tape execution context which just traces ops on the tape and delegates to some backing context for say tracing or eager execution or another `TapeContext`. This is implemented [here](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/c/experimental/gradients/tape/). 

I wrote this up but didn't add it to this doc because the execution interfaces are still not finalized yet and may change with @allenlavoie's work on op handlers. I am happy to add more details on [`TapeContext`](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/c/experimental/gradients/tape/tape_context.h), [`TapeOperation`](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/c/experimental/gradients/tape/tape_operation.h) here if that would be useful.

Automatic variable tracking is separate. It will remain a higher level feature for now.

rfcs/20201201-cpp-gradients.md
saxenasaurabh: Yes, I ran this idea by @mdanatg. We will possibly want to implement something like `CompositeTensorHandle` and subclass `IndexedSlicesTensorHandle` from that. The main takeaway here is that the tape API can support non-dense tensors by representing those as subclasses of `AbstractTensorHandle`.

rfcs/20201201-cpp-gradients.md
saxenasaurabh: I think it makes sense to clarify this a bit in the doc. Let me add something.

saxenasaurabh: Added.

vgvassilev:
> > > > Just a side note out of curiosity.
> > > > Is there anyone involved in this RFC who also collaborates on https://wg21.link/P2072?
> > > 
> > > 
> > > I don't think so. But looks like a great paper! From quickly skimming the paper they seem to be proposing language level autodiff support in C++, similar to Swift, as opposed to framework level. They have some claims about control flow gradients being tricky for frameworks which is not the case with TF. Nonetheless, it may be worth reaching out to them. Let me know if you have suggestions on who to contact.
> > > cc: @mdanatg @liufengdb
> > 
> > 
> > I don't know if @mfoco is still active on this in ISOCPP SG19 cc: @fraggamuffin
> 
> Thank you for reaching out!
> Sure I'm interested. Currently we're having regular meetings with @vgvassilev collecting ideas for a R1 of P2072. Mainly we're collecting use cases (and seems like you have a good one, even if it's lacking one of the big pushes to promote a language change, as @saxenasaurabh noted already).

Thanks for adding me in the loop. I will add a bunch of other people who might be interested: @efremale, @grimmmyshini, @davidlange6, @wsmoses.

I will read the proposal once again in depth. I am no TF expert (not even a user). If the gradients are only accessible via wrappers that may not be as performant as just synthesizing the gradient function. In [clad](https://github.com/vgassilev/clad) we use both approaches and the function-like wrapper object, unless carefully used, seems to be underperforming. I am sure it can be made efficient but may require paying extra attention to the initial design.

Making a step back, do you guys have a list of what's the minimal language support to implement that particular proposal? I assume you prefer language support and not a operator overloading implementation which requires data type?

bhack:
There was a typo in your username on the CLAD link :wink:  is https://github.com/vgvassilev/clad

wsmoses:
Also working with @vgvassilev @mfoco on some C++ language-level things.

It might also be worthwhile to take a quick read of the Enzyme paper (https://proceedings.neurips.cc/paper/2020/file/9332c513ef44b682e9347822c2e457ac-Paper.pdf, https://github.com/wsmoses/Enzyme).

Of specific note, we made a TF plugin for Enzyme that allows importing C/C++ programs into TF without additional rewriting which seems relevant/helpful.

saxenasaurabh:
> > > > > Just a side note out of curiosity.
> > > > > Is there anyone involved in this RFC who also collaborates on https://wg21.link/P2072?
> > > > 
> > > > 
> > > > I don't think so. But looks like a great paper! From quickly skimming the paper they seem to be proposing language level autodiff support in C++, similar to Swift, as opposed to framework level. They have some claims about control flow gradients being tricky for frameworks which is not the case with TF. Nonetheless, it may be worth reaching out to them. Let me know if you have suggestions on who to contact.
> > > > cc: @mdanatg @liufengdb
> > > 
> > > 
> > > I don't know if @mfoco is still active on this in ISOCPP SG19 cc: @fraggamuffin
> > 
> > 
> > Thank you for reaching out!
> > Sure I'm interested. Currently we're having regular meetings with @vgvassilev collecting ideas for a R1 of P2072. Mainly we're collecting use cases (and seems like you have a good one, even if it's lacking one of the big pushes to promote a language change, as @saxenasaurabh noted already).
> 
> Thanks for adding me in the loop. I will add a bunch of other people who might be interested: @efremale, @grimmmyshini, @davidlange6, @wsmoses.
> 
> I will read the proposal once again in depth. I am no TF expert (not even a user). If the gradients are only accessible via wrappers that may not be as performant as just synthesizing the gradient function. In [clad](https://github.com/vgvassilev/clad) we use both approaches and the function-like wrapper object, unless carefully used, seems to be underperforming. I am sure it can be made efficient but may require paying extra attention to the initial design.
> 
> Making a step back, do you guys have a list of what's the minimal language support to implement that particular proposal? I assume you prefer language support and not an operator overloading implementation which requires a special data type?

I was not aware of Clad. Looks super nice! I am particularly excited about the custom derivatives mechanism. We are working on adding model building APIs in TF C++. The `GradientFunction`s implemented for this proposal could be called from custom derivatives functions for clad. That would be interesting to try. 

Gradient computation is often memory intensive. The custom gradient signature in clad seems similar to `tf.recompute_grad` (i.e. you are passed in the original inputs and you can recompute whatever outputs/intermediates u need for gradients). It would be nice to be able to re-use computation here if needed by sharing intermediates from the forward pass. The `GradientFunction` design here allows this. (Apologies, if this is supported already and I missed it)

saxenasaurabh:
> Also working with @vgvassilev @mfoco on some C++ language-level things (like Clad/Enzyme integration).
> 
> It might also be worthwhile to take a quick read of the Enzyme paper, which performs AD at LLVM (and thus C/C++ and also Rust, Julia, Swift, etc) to great benefit (https://proceedings.neurips.cc/paper/2020/file/9332c513ef44b682e9347822c2e457ac-Paper.pdf, https://github.com/wsmoses/Enzyme).
> 
> Of specific note, we made a TF plugin for Enzyme that allows importing C/C++ programs into TF without additional rewriting which seems relevant/helpful.
> 
> Also since I think TF usually eventually goes to LLVM (via XLA), I imagine integration might allow for even better optimization between interfaces/ops.
> 
> cc @vchuravy @ftynse

Thanks, will do!

kkimdev:
Overall, looks good, thanks!

A couple of high level questions & feedback:
- Is there any different between Python gradient registration and Python tape implementation?  I think it's good to have a section discussing it or explicitly mention that it's 1:1 translation from the current Python implementation to C++.
- Can this be a basis for language-agnostic gradient registration and tape, unifying Python's gradient registration & tape as well?  For example, at high level, it seems it should be possible to implement Python's gradient registry thorough the new C++ gradient registry API, by having Python implementation of C++ `GradientFunction` class.

zaleslaw:
Great, it's very important for such projects as tensorflow-java or Tensorflow.NET. But do you have any plans to port gradients for Concatenate (looks like them are implemented in Python only in _ConcatGradHelper function in array_grad.py) and it's a really huge blocker for many modern CNN arhitectures.

Craigacp:
The TF Java SIG are very interested in this, as @zaleslaw says we've got a bunch of missing gradients (as they are purely python) which makes it near impossible to write certain classes of models in Java.

If there's something we can do to advance this let us know as it's very important. We're currently using the C API to access the TF runtime, so we could work towards defining that C API on top of the C++ API here. My understanding is that the C API is supposed to be stable and the C++ is less stable, but is this RFC going to cause the C++ API to become more stable? If so then we may need to switch which would be a lot of effort, and so we'd need to better understand the roadmap for the native APIs in TF.

cc @karllessard

kkimdev:
As you mentioned, a lot of Python gradient implementation should be re-written in C++ for other languages to take advantage of this, so once this is integrated, any help on those will be greatly helpful.

C API is not a part of this proposal but can be a follow-up.  One quick question: for Java, how important is to have ABI stable endpoint(C) compared to just having stable C++ API?

karllessard:
Architecturally-speaking, the C API is suppose to serve officially as the only "stable" intertace to the TensorFlow runtime so that is why we, SIG JVM, rely exclusively on it but a lot of TF features are still missing. 

I did not read the current proposal yet in details but are we talking here about adding missing gradients to the C++ TF client or to the C++ TF core? I guess the client could guarantee more backward-compatible stability than the core but basing the Java client on top of another client would be a bit contradictory with the original design.

kkimdev:
IIUC, the current proposal is for the C++ core, but we're having active discussions about having C++ client in general.

A bit more context: current TF C API is lacking and there have been discussions about having first-class public C++ API, but there are some details to be worked out.  For example, one way is [internal C++ API] -> [public C API] -> [header only public C++ API] and that has been the approach for the most recent saved model C++ API effort https://github.com/tensorflow/tensorflow/tree/master/tensorflow/cc/saved_model/experimental/public but it's on hold now.  One downside is that it has too much boilerplate code to support ABI-stable C++ API.   Another approach can be just [public C++ API] -> [public C API], dropping ABI stability from [public C++ API].  Then a remaining question is, the value of [public C API] and prioritization.  For example, one reasonable option could be prioritizing C API for kernel&device that will greatly benefit from stable ABI, and let other language clients to use C++ API, in case ABI stability is not a strong requirement.

saxenasaurabh:
There are ongoing discussions/prototypes of a more complete C/C++ API for TF2 (op building, tf.function, distribution etc). We are yet to finalize what those APIs would be and I expect more RFCs to be going out for that in Q1'ish. This proposal is about defining an (internal for now) C++ API that we can use to author gradient functions in C++ instead of python. We will be updating the python implementation of `tf.GradientTape` to use C++ gradient functions. Later on we can expose C/C++ APIs for a tape that uses these gradient functions.

@zaleslaw @Craigacp integrating C++ gradients into tensorflow-java and other projects would be exciting! I think we have ~300 gradient functions and we have only migrated about a [dozen](https://cs.opensource.google/search?q=public%5C%20GradientFunction%20-f:test&sq=&ss=tensorflow%2Ftensorflow:tensorflow%2Fc%2Fexperimental%2Fgradients%2F). I will write a guide on how to contribute and maybe setup a tracker. Any help with adding more gradient functions would be hugely appreciated!

zaleslaw:
@saxenasaurabh Either way, you can count me in. 
Great idea about "how to contribute", we could mapreduce this task together.
I have a few questions related to the topic:

1. What C++ version are you going to use (I finished on C++ 11 in my developer path)?
2. Of course, there a lot of gradients in Python, TFJS, .NET TF. Do you plan to use any of them as a source of inspiration?
3. Will be great to prepare a doc with full list of operands before setup tracker (we could prioriteze gradients there), what do you think? Where is the full-list of ops to have grads?

kkimdev:
[Discussed offline] Also, there should be a way for users to register C++ gradient for custom ops, which is done by `@tf.RegisterGradient` in Python now.  Though we can wait for a larger C++ API design roadmap for now.

yongtang:
@kkimdev For customer ops it would be great if a C API is in place to register gradient. The biggest drawback is that with C++ API there is no stable backward-compability and custom ops has to be rebuilt constantly with tensorflow code for every release. It is not an issue for tensorflow core repo itself (as it is more or less a monorepo) It is really an issue for outside community to maintain their custom ops (where they have to rebuilt constantly to keep binary compatibility).

karllessard:
I agree with @yongtang , and to add to this, it is not only a problem for external projects but also for other projects that are part of the TensorFlow ecosystem, like the Java client.

I think we all agree that there is a lot of Python logic that needs to be migrated to C++ in order to make it accessible for all other languages and I'm happy to see these kind of initiatives. I'm just concerned that if wrapping this work behind a stable ABI is not a priority, this will remain an experiment and users might feel reluctant to incorporate it safely inside their projects.

Now should this stable ABI remain in pure C or should it migrate to C++, totally or partially? C makes it simple for sure, the actual JNI layer used in the Java client is really thin for that reason. If there is a great motivation at Google to move to C++ though, we can also adapt our code to bind to it. But the sooner we know in which direction that goes, the better. Should this be adressed as a separate topic?

miaout17:
Naive questions -- If I understand correctly, the gradient computation was already doable in C++ with APIs like `tensorflow::AddSymbolicGradients`. I also have a vague impression that symbolic gradient is going to be replaced with Gradient Tape, but I don't remember much details. 

Could you elaborate what's the new functionality brought by the this RFC, in terms of constructing the gradient graph?

saxenasaurabh:
> Overall, looks good, thanks!
> 
> A couple of high level questions & feedback:

Sorry, I missed your comment!

> 
> * Is there any difference between this C++ proposal and the Python gradient registration&tape?  I think it's good to have a section discussing it or explicitly mention that it's 1:1 translation from the current Python implementation to C++.

The tape API described here is a simplified wrapper over the existing C++ `Tape` that the python `GradientTape` uses so feature wise this should have parity with the python tape. This API actually matches more closely to the python's `GradientTape`. I think there are some jacobian APIs in the python tape that might still remain in python for now.

> * Can this be a basis for language-agnostic gradient registration and tape, unifying Python's gradient registration & tape as well?  For example, at high level, it seems it should be possible to implement Python's gradient registry thorough the new C++ gradient registry API, by having Python implementation of C++ `GradientFunction` class.

Yes. That's one of the primary motivations. I have mentioned this briefly in the `Python rollout` section. I intend to use this Tape and registry in python and build a `GradientFunction` that calls into python for gradients that are yet to be migrated.

miaout17:
It seems to me that the RFC has 2 major innovations:

1. Being able to register and compute the gradient in C++. 
2. Put the gradient computation of an op into one C++ function, which is `GradientFunciton::Compute`

I think the 1st is great and I'd like to talk more about the 2nd. 

It's understandable that we want to minimize the overhead of gradient computation. In Eager mode it might be very expensive since it involves Python overhead. In graph mode it might be negligible since it's just kernel dispatching in C++. 

However I have a few questions:

- Do we have a way to put the GradientFunction in a TF Graph? (SavedModel or GraphDef etc) I believe we will have use cases where we want to pre-construct the graph and just run it as is. This includes the upcoming TFLite training. 
- Have you considered to define a gradient as a decomposition of other ops? You can still avoid Python overhead this way. I'm brining up this point because:
  - TensorFlow has an exploding number of ops. I think we discussed and concluded that we should not define more fused gradient ops like `ReluGrad`, as long as we can compose the logic from other ops. It seems to me that the new `GradientFunction::Compute` has some similar problem. 
  - The system will be simpler if we can have one less concept. We can just register a subgraph decomposed ops in C++. We don't need to support a new concept `GradientFunction` in downstream systems, like TFRT / TFLite etc. 

This is based on my limited understanding by reading the doc. Let me know if I misunderstood something. Thanks!

saxenasaurabh:
> Naive questions -- If I understand correctly, the gradient computation was already doable in C++ with APIs like `tensorflow::AddSymbolicGradients`.

Great question. Indeed we have `tensorflow::AddSymbolicGradients` for TF1 but it is graph-mode only and it is not complete. The gradients API proposed here uses eager/graph-agnostic C++ APIs (`AbstractTensorHandle` mainly). This allows us to eventually land a C/C++ GradientTape API that works both in graph and eager modes. One thing we are doing differently from the earlier C++ graph gradients effort is that we plan to integrate these C++ gradients into python. IIUC the C++ graph gradients are only used in `TF_AddGradients` and for the `SymbolicGradient` op right now.

> I also have a vague impression that symbolic gradient is going to be replaced with Gradient Tape, but I don't remember 
much details.

Could you explain this a bit? 

> Could you elaborate what's the new functionality brought by the this RFC, in terms of constructing the gradient graph?

For graph mode, this mainly moves the gradient graph building to C++ which should hopefully speed things up substantially for large models.

saxenasaurabh:
Notes from design review:

*   Changes since the RFC was sent out
    *   Clarify the status of the API -- Currently an experimental API in the internal framework
    *   Add details about `TapeContext`
*   Comments from C++ machine learning group
    *   Just a FYI -- They're trying to get autodiff in C++
*   Comments about non-dense and composite tensors
    *   The design should be possible to be extended to support `IndexedSlices`, `SparseTensor` and other `CompositeTensor`s
    *   For gradients we possibly only care about `IndexedSlices`.
    *   Composite tensor is not a concept exposed in C++ at all currently. 
*   What are the new functionalities in this RFC?
    *   The existing C++ Tape API can compute gradients, but is very low level and not really usable
    *   Functionalities like registering gradient in C++ is new
*   Do we have a way to register C++ gradient for custom ops?
    *   Yes
    *   Gradient registry is simply a dictionary from op’s name to gradient function factory
*   Can you use composition of existing ops to implement gradients?
    *   Yes
    *   Gradients are not kernels. It's actually composition of existing ops
*   Does the `Compute` function have the capability to build the graph? or does it always compute the gradient eagerly?
    *   Relies on the new abstraction that shares the implementation between Eager and Graph
    *   Using `AbstractContext` and `AbstractTensorHandle`
    *   It seems we have multiple ways to do "decomposition" now:
        *   The TF composite op proposal (in Python)
        *   Using the C++ `AbstractContext` and `AbstractTensorHandle` in graph mode
*   This RFC is mostly an internal change
    *   No Python frontend APIs change
    *   No change to the GraphDef/SavedModel representation. E.g. The new `GradientFunction` is not a concept in the graph. 
*   Are there any cases where the semantic doesn't work out with the dispatching mechanism?
    *   Theoretically there's no problem
    *   In certain cases information may only be available in python e.g. handle\_data of DT\_RESOURCE and DT\_VARIANT tensors. Dan’s full-type proposal will help expose that in C++. We will port concepts as necessary from python to C++.
    *   Writing control flow gradient in C++ will be interesting project, but we don't think there's a limitation in the framework
*   What are the existing C++ gradient functions used for?
    *   There's a `TF_AddGradients` C API
    *   These APIs are very closely tightly to graph mode. It's hard to repurpose these to work with Eager mode. 
    *   `tensorflow::AddSymbolicGradients` only have very few users, like lingvo (they recently migrated to tf.function as well)
*   Question around still using string to identify ops
    *   There are 2 layers of the API
    *   The lower level doesn't use the op name. It only takes inputs and produces outputs. 
    *   The op name mainly comes from the registry. 
    *   The op name thing isn't an issue if we "intern" the strings properly. 
        *   E.g. If you use both Python and C++, the single source of truth of intern-ing should be Python. 
        *   It might require a significant change to gen\_ops but it is probably independent from this project. 
*   Status of RFC?
    *   There is nothing controversial 
    *   This is an internal implementation. There are no public APIs.

ematejska (approved):

rnett:
What's the planned strategy for defining/registering gradients?  Currently it seems like each `GradientRegistry` needs to be initialized manually, are there plans for a global base registry or something similar?

Also, is there a better place for questions about this?

PR #343, opened 2021-01-07
Proposed updates to the RFC template

ematejska:


rfcs/yyyymmdd-rfc-template.md
```diff
@@ -2,10 +2,8 @@
 
 | Status        | (Proposed / Accepted / Implemented / Obsolete)       |
 :-------------- |:---------------------------------------------------- |
-| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)|
```
ewilderj, line 5: The number header was added relatively recently to help people feel comfortable with referring to RFCs by number if they wished. It would be good to heart from the developer community if they still care about being able to do this.

ematejska: Thanks, I'll check.

ematejska: In practice, nobody has been using this.

ematejska:
@ewilderj I'm still making some changes and going to seek feedback so not quite ready for review.

ewilderj:
@ematejska got it!

rfcs/yyyymmdd-rfc-template.md
theadactyl: I do feel like it could be pretty straightforward to refer to an RFC number as you would an issue or PR number without it being explicitly numbered here. I find myself updating this field for all submitted RFCs, so in the name of reducing busywork I'm pro dropping it. 

@bhack any thoughts here?

bhack: I think that it is quite useful to have a reference to the PR to retrieve the context, the review comments and original related discussions to the RFC. These are available only in the PR after the RFC is merged.
But if it cannot be managed/updated and generally the RFC is not updated with subsequent PRs it could be quite easy to find the original PR on github.

lc0: git/github history would also show at what PR it was merged, although it would require a few more clicks

PR #349, opened 2021-01-25
RFC: TFX Periodic Training

1025KB:
# TFX Periodic Training

Status        | Proposed
:------------ | :--------------------------------------------------------------
**Author(s)** | Jiayi Zhao (jyzhao@google.com), Ruoyu Liu (ruoyu@google.com)
**Sponsor**   | Zhitao Li (zhitaoli@google.com), Mitch Trott (trott@google.com)
**Updated**   | 2021-01-15

## Objective

This doc proposes a periodic batch pipeline for training on regularly arriving
blocks of data. For each new arriving data block, a batch pipeline run is
scheduled for training. In each run, ExampleGen processes and outputs one Span,
which contains the new increment of data. The Resolver is responsible for
retrieving Spans within the rolling window for training. If an older Span needs
to be updated, a new version of the Span is generated by a backfill pipeline run
with specified Span id.

### Goal

*   Propose TFX periodic batch pipeline, which provides basic automation for
    continuously arriving blocks of data, as an extension to current one-shot
    batch pipeline support.
*   Clearly delineate the boundary between automation and manual handling of
    exception cases.
*   Provide explicit versioning of data.


rfcs/20210115-tfx-periodic-training.md
```diff
@@ -0,0 +1,445 @@
...
+
+As mentioned above, each ExampleGen execution generates only one Span at a time.
+ExampleGen can be configured to process a specific Span (always taking its
+latest Version); if a Span is not specified, ExampleGen selects the input data
+corresponding to the latest Version of the latest Span by default. The latest
+here means the one that has the highest Span id.
+
+Relying on “latest Span” semantics in periodic automation introduces a new
+failure mode: if the data isn’t available as expected the pipeline will
+reprocess old data. While it’s possible to guard against this with upstream
```
casassg, line 343: reprocessing can be avoided with caching. Wondering what's the tradeoff between letting TFX just skip re-processing with caching vs using runtime parameters.

rfcs/20210115-tfx-periodic-training.md
```diff
@@ -0,0 +1,445 @@
...
+monitoring and alerting, we recommend specifying the span as runtime parameter
+when triggering the run. For example, for a daily cron-triggered pipeline, pass
+the scheduled date to the job for calculating target span of ExampleGen. The
```
casassg, line 346: Is that even possible? Currently RuntimeParameters are only supported in Kubeflow and I believe Kubeflow does not allow for dynamic parameters when using [run triggers](https://www.kubeflow.org/docs/pipelines/overview/concepts/run-trigger/). How do you set the scheduled date in Kubeflow? Is there plans to add support for it on Airflow?

rfcs/20210115-tfx-periodic-training.md
```diff
@@ -0,0 +1,445 @@
...
+pipeline will fail when the specified span’s input data is not available. This
+provides deterministic behavior for each periodic run and makes it easier for
+the pipeline owner to identify which span needs to be backfilled after failure.
```
casassg, line 349: Also, related to this. Are there plans to add a span based rolling resolver? Given currently there's only [this](https://github.com/tensorflow/tfx/blob/master/tfx/dsl/experimental/latest_artifacts_resolver.py) and it's based on latest available instead.

rfcs/20210115-tfx-periodic-training.md
```diff
@@ -0,0 +1,445 @@
...
+    once for updating the model after all Spans are updated.
+
+We plan to provide the ability to specify a partial pipeline to execute at
+runtime (see Future Work below) to enable this workflow.
+
+An alternative approach is to define the two pipeline portions as TFX pipelines
+in their own right, to make backfill easier. This is not recommended because it
+makes the normal case of periodic training more brittle: the two pipelines must
+now be triggered in sequence by some non-TFX mechanism, thereby multiplying the
+possible fault conditions.
```
casassg, line 321: How about using a Cross pipeline ResolverNode. Aka resolve from another pipeline. This allows you to do some kind of asyncronous support:
- ExampleGen pipeline runs every hour
- Training pipeline runs every day and takes latest 48 spans from the other pipeline.

rfcs/20210115-tfx-periodic-training.md
```diff
@@ -0,0 +1,445 @@
...
+data, and the model will be trained based on the ExampleGen’s output Span.
+
+<div style="text-align:center"><img src='20210115-tfx-periodic-training/single_span.png', width='600'></div>
+
+*For example, assume a Span corresponds to a week of data, and every Monday, the
+previous week’s data is finalized. Users can set up a routine TFX pipeline to
+run every Wednesday, so that ExampleGen will pick up the previous week’s data
+and generate a corresponding output Span for training.*
+
+#### Rolling Window
```
casassg, line 214: Not sure, but is the plan to support different windows per split in the dataset.

If I'm doing serialized data, I may want to train on N-5 to N-1 days and evaluate on N day.

rfcs/20210115-tfx-periodic-training.md
```diff
@@ -0,0 +1,445 @@
...
+*   Integrate with Google Cloud AI Platform, e.g., utilize
+    [Pub/Sub](https://cloud.google.com/pubsub) and
+    [Beam Streaming](https://beam.apache.org/documentation/sdks/python-streaming/)
+    for input data generation, and use
+    [Cloud Functions](https://cloud.google.com/functions) for auto triggering of
+    TFX pipelines.
+*   Support partial pipeline execution.
+*   Trainer warm-start support for Keras.
+*   Asynchronous execution for components.
+*   Custom ExampleGen guidance.
```
casassg, line 445: Special interest on QueryBased

junbopark:
We use TFX pipeline to train on daily available data. As has been mentioned in other comments would it be possible to
- implement SPAN in query-based ExampleGen? Our data gets inserted into date-labeled BigQuery table (e.g., data_20210125).
- ability to choose train/eval split based on span-ID (like the comment above)? For example, we'd like the train/eval splits to be last five days / today (so last five spans / current span).

rfcs/20210115-tfx-periodic-training.md
1025KB: "reprocessing can be avoid with caching." yes for the component without input change.
If a component's input depends on mlmd (e.g., resolver), the outcome might change for the reprocessing.
E.g., the latest blessed model will change when you run the pipeline the second time.

besides that, though it's cached without reprocessing, users might not notice there is a data missing happened upstream. Unless there are good alert/monitoring system available at user side, we suggest to specify the span to expose this issue when upstream is missing.

rfcs/20210115-tfx-periodic-training.md
1025KB: For beam based DagRunner, it's main job is "compile and run" so it's by default runtime param.
For Kubeflow, the runtime params can be specified in UI or using Client API. I will confirm if run triggers supports it.
For Airflow, it has its own [runtime parameters](https://stackoverflow.com/questions/53663534/for-apache-airflow-how-can-i-pass-the-parameters-when-manually-trigger-dag-via) supports,

rfcs/20210115-tfx-periodic-training.md
1025KB: Yep, WIP, will add a resolver that takes range_config

rfcs/20210115-tfx-periodic-training.md
1025KB: For async, it's kind of risky to break sync pipeline into multiple sub pipeline and schedule separately. Thus we only recommend break pipeline when there is [backfill needs](https://github.com/tensorflow/community/blob/875c04645f9029cb3c5d75bfdb8bf63e5560e9d9/rfcs/20210115-tfx-periodic-training.md#partial-pipeline-execution) , It has a lot of potential issues and not very robust. 
Internally we have a centralized orchestrator for async, if async is indeed needed in your case, stay tuned and we are working on it!

rfcs/20210115-tfx-periodic-training.md
1025KB: it depends on how you set up your pipeline, 
If Evaluator takes Resolver's output (span n to n-windowsize), it will use eval based on n spans,
if Evaluator takes ExampleGen's output (only have current span n), it will use span n to eval.

rfcs/20210115-tfx-periodic-training.md
1025KB: WIP! This doc uses file based as examples, but the concept is not tied to file based

1025KB:
> We use TFX pipeline to train on daily available data. As has been mentioned in other comments would it be possible to
> 
> * implement SPAN in query-based ExampleGen? Our data gets inserted into date-labeled BigQuery table (e.g., data_20210125).
> * ability to choose train/eval split based on span-ID (like the comment above)? For example, we'd like the train/eval splits to be last five days / today (so last five spans / current span).

This doc uses file based as examples, but the concept is not tied to file based.

E.g., in file based, the input pattern is path/{SPAN}/{Version}/data
and in query based, the query pattern can be select * from table where span={SPAN}

we plan to provide a simple query based example later

rfcs/20210115-tfx-periodic-training.md
casassg: > For Airflow, it has its own runtime parameters supports
Indeed, wondering if we can add support for those from TFX (use RuntimeParameter similarly). This could be accomplished by using template_fields in the operator and mapping RuntimParamete to be [DagParam]( https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/dagparam/index.html#airflow.models.dagparam.DagParam) object (only Airflow 2.0>)
> Kubeflow
run triggers does support parametrization, but it does not support having a parameter be the current date/execution date as far as I know.

rfcs/20210115-tfx-periodic-training.md
casassg: Definetly interested on Async! Looking forward to that!

rfcs/20210115-tfx-periodic-training.md
casassg: When I meant evaluate I mean eval split. Aka splits in the dataset have different spans.

1025KB: each run of ExampleGen generates a span with multiple splits, e.g., train and eval.

Thus depends on how many span you get, if Evaluator takes span 1,2,3, it will get a train split (that contains span1train, span2train, span3train) and a eval split (contains span1eval, span2eval, span3eval)

rfcs/20210115-tfx-periodic-training.md
1025KB: > > For Airflow, it has its own runtime parameters supports
> 
> Indeed, wondering if we can add support for those from TFX (use RuntimeParameter similarly). This could be accomplished by using template_fields in the operator and mapping RuntimParamete to be [DagParam](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/dagparam/index.html#airflow.models.dagparam.DagParam) object (only Airflow 2.0>)

Not Airflow expert, but I wonder if we can directly use the params in airflow pipeline dsl?

e.g. ExampleGen(span={some airflow runtime param})

> 
> > Kubeflow
> 
> run triggers does support parametrization, but it does not support having a parameter be the current date/execution date as far as I know.

casassg: > Not Airflow expert, but I wonder if we can directly use the params in airflow pipeline dsl?

They won't. Airflow be default resolves `template_fields` which are class variables `self.X`. Current implementation just wraps everything into a [function](https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/airflow/airflow_component.py#L118) which means that Airflow has no visibility into it. 

One way to do it would be to add all exec_properties as a class variable, mark it as template_field and set all exec_properties there in a list. It may work, would need to test this.

1025KB: > > For Airflow, it has its own runtime parameters supports
> 
> Indeed, wondering if we can add support for those from TFX (use RuntimeParameter similarly). This could be accomplished by using template_fields in the operator and mapping RuntimParamete to be [DagParam](https://airflow.apache.org/docs/apache-airflow/stable/_api/airflow/models/dagparam/index.html#airflow.models.dagparam.DagParam) object (only Airflow 2.0>)
> 
> > Kubeflow
> 
> run triggers does support parametrization, but it does not support having a parameter be the current date/execution date as far as I know.

I see, Note that in Managed Pipeline we will for sure have a placeholder for the date/time.
For kubeflow run trigger, if it's not supported, a workaround can be we use client API to pass in date/span in cli, and you can have a script/cron that triggers CLI periodically.

casassg: > I see, Note that in Managed Pipeline we will for sure have a placeholder for the date/time.
For kubeflow run trigger, if it's not supported, a workaround can be we use client API to pass in date/span in cli, and you can have a script/cron that triggers CLI periodically.

That's true. Still not ideal though. Wondering if anyone from Kubeflow can weigh in here.

1025KB: > > I see, Note that in Managed Pipeline we will for sure have a placeholder for the date/time.
> > For kubeflow run trigger, if it's not supported, a workaround can be we use client API to pass in date/span in cli, and you can have a script/cron that triggers CLI periodically.
> 
> That's true. Still not ideal though. Wondering if anyone from Kubeflow can weigh in here.

Hi, I asked Kubeflow team, they do have that,  here is an [example](https://github.com/kubeflow/pipelines/blob/34d23aa924720ead13fe67ebca5c1ab9926585ee/backend/src/crd/samples/scheduledworkflow/parameterized.yaml#L29-L34)

casassg: Thats awesome! Thanks! Would be great if we could include this reference on the RFC or public docs as an example on how to do scheduled training.

1025KB: > default='[[ScheduledTime]]'

Screen Shot 2021-01-29 at 3 20 03 PM
you can also change "hello world" to placeholder like "[[ScheduledTime]]" in UI, if it's not specified in UI, it will use default value like the code you wrote above

rfcs/20210115-tfx-periodic-training.md
misterpeddy: @casassg I'm wondering what is the end goal behind the setup you suggest (compared to having the ExampleGen run every day with the training pipeline, and using a resolver to yield the latest 48 spans)?

Are you trying to reduce the end-to-end latency of your training pipeline by pre-materializing 23 of the 24 new spans? Or is there some other utility in doing the materialization incrementally (ex. load on whatever data source you're ingesting data from?)

casassg: There's 2 reasons. 
1) data latency. By decoupling data processing from training, we can get faster latency for data needed (aka we just need to wait to generate 1 hour instead of having to generate a full day of data before starting our pipeline). This is the same principle and benefit explained in ASYNC pipelines, but using several pipelines and cross-pipeline dependencies to achieve it (this already is well supported within Airflow btw)

2) Better redundancy. Sometimes data can be delayed due to problems in data processing upstream. In this case, we may want to run the pipeline w the latest 24 spans (given its hour based, you are only 1 hour delayed vs unknown waiting time when issue there are several issues). This could be also achieved if we can trigger manually parts of the pipeline.

misterpeddy: Thanks - 1 makes sense and agreed that async pipelines is the more robust way of accomplishing this.

on the failure mode you describe in 2, is the "problems in data processing upstream" transient errors that upon retry will succeed? Because if not, and the data is simply not "ready"/accessible in the upstream source, the only benefit of more frequent ingestion is knowing about the upstream failure earlier and fixing it before the training run happens; so I don't quite follow the 'run the pipeline w the latest 24 spans' mitigation. 

(to provide broader context, I'm trying to validate the hypothesis that async will solve _all_ the use cases that you are tempted to solve with this cross-pipeline setup :)) 

Thanks for the discussion!

casassg: For 2 what I meant is using Latest 24 available spans. This could be useful when refreshing your model is more time sensitive than it is having the latest data.

Example:
I want to deploy a model every day at 9PM  w 24 hours data (bc online metrics go down after 1 day due to changing data)
- 7PM - Data pipeline starts failing  - Alerts team
- 9PM - Model trains w data from DAY-1.7PM to DAY.7PM (latest 24 spans) - This avoids keeping an "expired" model around for too long and only uses 2 hour old data which is not terrible.
- 11PM - Fix data pipeline upstream and backfill data (ExampleGen) from 7PM to 11PM spans for next model.

I do think this can be achieved w Async pipelines (at least from what I understand) but just making sure.

1025KB: This can be done with sync pipeline too (but yep, with higher latency)

e.g., config ExampleGen pattern to grab a certain set of data, or allow Resolver to tolerate some missing data.

But agree, async can make it more automatic due to the centralized orchestration

rfcs/20210115-tfx-periodic-training.md
casassg: My use case was more of:

I have 5 spans {1,2,3,4,5}
- I want to use spans 1-4 for train split
- I want to use span 5 for eval split. 

Currently this is impossible due to range_config being a component level config instead of Split config.

1025KB: currently both sync/async follows the same [span/version/split concept](https://github.com/tensorflow/tfx/blob/master/docs/guide/examplegen.md#span-version-and-split)

so if you do want to have different split config, you need to have 2 examplegens, this is same for sync and async.

Note that for model training, if it take span 1-4, the training and testing dataset must from those span 1-4
but for Evaluator, it can take a different dataset compared to training, and it can configure which split to use in that dataset.

So if you can have the same splits for all spans, you can setup the sync pipeline like this,
span 1-5 has train/eval/test, span 1-5's train/eval is provided to training to train a model, span 5's test is provided to evaluator for evaluation

zhitaoli:
Design Review Notes:

Jumbo: question on using different span, 4 for training and one for eval. From user’s point of view, Data is precious and we want to use all of them

Jiayi: this should be doable, depends on you write pipeline. If resolver -> evaluator returns 4 spans, evaluator should be all spans for evaluation.

Robert: if you use all data for training, eval is not useful.

Jiayi: we separate out the data between training and evaluation.

Jumbo: will play with it once we have it.

Gerard: proposed to use example gen. How will cache be handled? 

Jiayi: resolver could return previous run’s output. 

Gerard: do we need N resolvers?

Jiayi: currently working on a resolver returning multiple spans.

Gerard: worried about window of resolving. 

Jiayi: output from current pipeline does not need a resolver.

Gus: Yes once seeing the code it’s obvious. Think about TFX pipeline’s base dir as all data. Resolver allows rolling range, latest version of latest span, etc. Different resolvers for different trainers, etc. 

Jiayi: Will provide an end to end example for a simple resolver.

Gus: Are we externalizing range config fully?

Jiayi: Yes. starting from simple static range and rolling range. Example PR

Robert: any impact to batch inference?
Jiayi: it should be the same to config spans to pass to bulk inferrer.

Gerard: It will be really useful to have a real example for passing the span from/to orchestrator. 

Jiayi: For kubeflow, the difference between compile and runtime parameters are more visible. And we will document how this works.

ematejska (approved):

PR #364, opened 2021-03-10
RFC: (TFX) Support structured data in TFX through `struct2tensor` and `DataView`

brills:
_This RFC will be open for comment until March 19, 2021._

Status        | Proposed
:------------ | :--------------------------------------------------------------
**RFC # | 364
**Author(s)** | Zhuo Peng (zhuo@google.com)
**Sponsor**   | Zhitao Li (zhitaoli@google.com)
**Updated**   | 2021-03-05

## Objective

This RFC proposes several additions to TFX in order to support building ML
pipelines that process __structurally richer__ data that TFX does not have
apriori knowledge about how to parse. Such knowledge is provided by the
user, through __`struct2tensor`__ (showcased in this RFC) or other TensorFlow
graphs and made available to all TFX components through __Standardized TFX
inputs__ and __`DataView`s__.


rfcs/20210305-tfx-struct2tensor.md
```diff
@@ -0,0 +1,221 @@
...
+
+*   (4) `TFXIO` will also create `TensorRepresentations` according to the output
+    signature of the saved TF graph, so that the following is identity: PICTURE
+    3
+
+*   (5) For TF trainers, `TFXIO` creates a `tf.data.Dataset` that:
+
+    -   reads the serialized records of protobufs as a string tensor
+    -   `.map()` the string tensor to decoded it into tensors using the saved
+        `struct2tensor` query.
```
casassg, line 140: One question we have here is: How are you supposed to use structured datasets with tf.transform. Afaik TF.Transform outputs tf.example which is fine since on trainer then it will load fine, but you will need to access the dataview on trainer to create your serving signature.

Not sure if I'm missing something here.

CC @newtonle

brills: There are two ways:

1. You can disable TF Transform to output transformed examples [1]. Then, TF Transform will only output a TF Graph that contains the preprocessing_fn. That preprocessing_fn can already take tf.RaggedTensors and output tf.RaggedTensors, which struct2tensor can produce.

1. Potentially TF Transform can allow encoding tf.RaggedTensors and arbitrary tf.SparseTensors into tf.Examples. There are already some related work going on there.

[1] https://github.com/tensorflow/tfx/blob/c35e625ae058c35f80f9f71918071b840a8e787c/tfx/examples/ranking/ranking_pipeline.py#L105

brills: > but you will need to access the dataview on trainer to create your serving signature.

Trainer will offer an API to access the DataView (in the form of a tf.function). Currently though, since that tf.function essentially lives in the module file where the trainer code lives, you can get it without getting the DataView (https://github.com/tensorflow/tfx/blob/c35e625ae058c35f80f9f71918071b840a8e787c/tfx/examples/ranking/ranking_utils.py#L96)

casassg: Would be interesting to learn on 2.

Regarding 1, would be nice to see an example where TFT tries to manipulate the ragged tensors. The Ranking example seems to not do anything (just calculate a vocabulary).

I think having it would be nice. Since relying on being in the same module may have unintended side effects (like if I'm using a newer trainer run_fn for an old dataview).

brills: > Regarding 1, would be nice to see an example where TFT tries to manipulate the ragged tensors. The Ranking example seems to not do anything (just calculate a vocabulary).
you can do all things supported by TF / TFT regarding those ragged tensors provided as the input to the preprocessing_fn. Note that TFT may not have a lot of analyzers that support RaggedTensors, but on the other hand, those analyzers may not need to understand the RaggedTensor's structure, which means you can simply apply the analyers against the RaggedTensor's `.flat_value`, then construct another RaggedTensor which has exactly the same structure, but with flat values replaced by the transformed values.


> I think having it be provided in the API (guessing TrainerArgs) would be nice. Since relying on being in the same module may have unintended side effects (like if I'm using a newer trainer run_fn for an old dataview).

Yes. That API will come.

brills: >> I think having it be provided in the API (guessing TrainerArgs) would be nice. Since relying on being in the same module may have unintended side effects (like if I'm using a newer trainer run_fn for an old dataview).

> Yes. That API will come.

https://github.com/tensorflow/tfx/commit/d2052ecddb47ac8f1a6162254b91427cbcc9ba8b

zhitaoli:
LGTM

theadactyl (approved):
LGTM

PR #372, opened 2021-03-24
RFC: API & Artifact Compatibility for TFX OSS & Libraries

zhitaoli:
# API & Artifact Compatibility for TFX OSS & Libraries

This RFC will be open for comments until April 7, 2021

| Status        | Proposed                                                                                      |
| **RFC #**     | [API & Artifact Compatibility for TFX::OSS & Libraries.](https://github.com/tensorflow/community/pull/372)|
| **Author(s)** | Zhitao Li (zhitaoli@google.com), Jiayi Zhao (jyzhao@google.com), Zhuo Peng (zhuo@google.com)         |
| **Sponsor**   | Konstantinos Katsiapis (katsiapis@google.com), Zhitao Li (zhitaoli@google.com)                 |
| **Updated**   | 2021-03-21                                           |

## Objective

Defining requirements and processes for TFX::OSS and related libraries to achieve
necessary API & Artifact compatibility.

### Goals

*   Clearly define a stable API surface of TFX::OSS, and lay out a structure to
    determine what falls into the stable public APIs.
*   API stability: TFX::OSS (as pipeline) and all APIs from Key Dependency Libs
     exposed through TFX::OSS should achieve desired backward compatibility.

### DIFF
--- a/rfcs/20210307-tfx-api-artifact-compat.md
+++ b/rfcs/20210307-tfx-api-artifact-compat.md
@@ -203,6 +203,12 @@ Therefore we propose the following for all TFX 1.x versions:
         to test tfx with `apache-beam` RC so such breakages should be
         exceedingly rare in the future.
 
+For users who wants to use older versions of `apache-beam` and `tfx` together,
+while TFX team cannot directly support them in our released packages, we plan to
+refactor our build scripts so that beam dependency can be overridden by
+themselves, so they can build Python packages and container images in a
+self-service fashion.
+
 #### PyArrow {#pyarrow}
 
 TFX’s dependency on PyArrow comes from
@@ -360,11 +366,8 @@ compatibility:
 The following Dag runners will remain experimental:
 
 *   AirflowDagRunner
-    *   Honestly speaking, support of this one is on the fence. Due to lack of
-        proper configuration for a distributed backend support, it’s not much
-        more powerful than the LocalDagRunner when used out-of-box. However,
-        some partner teams managed to connect this to a distributed backend.
-        Until that is supported in TFX, we don’t think this one is stable.
+    *   Due to low utilization of the AirflowDagRunner, we want to avoid calling
+        it stable. Support will continue to graduate it into production as utilization evolves.
 
 The following dag runners are slated to be deprecated and therefore won’t
 comply:
--- a/rfcs/20210307-tfx-api-artifact-compat.md
+++ b/rfcs/20210307-tfx-api-artifact-compat.md
@@ -1,0 +1,592 @@
+# API & Artifact Compatibility for TFX::OSS & Libraries
+
+| Status        | Proposed                                                 |
+| :------------ | :------------------------------------------------------- |
+| **Author(s)** | Zhitao Li (zhitaoli@google.com), Jiayi Zhao              |
+:               : (jyzhao@google.com), Zhuo Peng (zhuo@google.com)         :
+| **Sponsor**   | Konstantinos Katsiapis (katsiapis@google.com), Zhitao Li |
+:               : (zhitaoli@google.com)                                    :
+| **Updated**   | 2021-03-21                                               |
+
+[TOC]
+
+## Objective {#objective}
+
+Defining requirements and processes for TFX::OSS and related libraries (see list
+below) to achieve necessary API & Artifact compatibility, in order to satisfy
+the following efforts (which drives the requirements and scope of this doc).
+
+### 1.0 Release of OSS {#1-0-release-of-oss}
+
+It has been long sought for TFX and the related projects to reach a stable 1.0
+release, possibly subject to [Semantic Versioning](https://semver.org/) (aka
+SemVer). Most important requirement is that **all minor versions within the same
+major version must be backward compatible**, which applies to both public API
+and Artifacts.
+
+### Goals {#goals}
+
+*   Clearly define a stable API surface of TFX::OSS, or at least lay out a
+    structure to determine what falls into the stable public APIs.
+*   API stability: TFX::OSS (as pipeline) and all APIs from the
+    [Key Dependency Libs](#key-dependency-libraries-in-scope) exposed through
+    TFX::OSS should achieve desired backward compatibility (see
+    [Key Decisions](#key-decisions)).
+
+#### Key Dependency Libraries In Scope {#key-dependency-libraries-in-scope}
+
+The following libraries are included as key dependencies of TFX::OSS, and
+subject to the compatibility scope laid out in this document:
+
+*   [tfx-bsl](https://pypi.org/project/tfx-bsl/)
+*   [ml-metadata](https://pypi.org/project/ml-metadata)
+*   [tensorflow-metadata](https://pypi.org/project/tensorflow-metadata/)
+*   [tensorflow-data-validation](https://pypi.org/project/tensorflow-data-validation/)
+*   [tensorflow-model-analysis](https://pypi.org/project/tensorflow-model-analysis/)
+*   [tensorflow-transform](https://pypi.org/project/tensorflow-transform/)
+*   [ml-pipelines-sdk](https://pypi.org/project/ml-pipelines-sdk)
+
+### Non-Goals {#non-goals}
+
+*   LTS Version: No LTS version is targeted in the current scope of this doc.
+
+### User Benefit {#user-benefit}
+
+Enterprise users of TFX::OSS will benefit from a stable distribution of TFX
+after these goals, and being assured that their TFX pipelines will keep working
+after an upgrade, within the defined compatibility scope.
+
+## Motivation {#motivation}
+
+Specific Enterprise users who have requested a stable distribution:
+
+*   SAP Concur: quote from a partner : _“I think the engineering management is
+    simply waiting for a 1.x production release of TFX for the final adoption.”_
+
+## Design Proposal {#design-proposal}
+
+There are generally two types of compatibilities:
+
+*   _Backward Compatibility_ allows a newer version of software to interoperate
+    with older versions of the same or different software.
+*   _Forward Compatibility_ allows a version of software to accept input from a
+    future version of the software itself, or a different software producer.
+
+Due to the nature of TFX libraries (python libraries which usually produces some
+Artifacts), there are also several aspects for each libraries involved:
+
+1.  Artifact Type (aka Metadata): This is the metadata which describes the
+    produced Artifact(s) from a TFX lib, e.g. `Statistics`/`Schema` for
+    TFDV/TFMA, `TransformGraph` for Transform;
+2.  Artifact Payload: This is the actual payload of the Artifact, which is often
+    files on external storage;
+3.  API: This generally refers to all publicly visible protobuf definitions, and
+    public Python symbols, including but not limited to functions, classes,
+    variables and decorators (see our definition of public Python APIs below).
+    All C++ APIs are out-of-scope for this document.
+
+### Key Decisions {#key-decisions}
+
+The following decisions are provided for easy digestion:
+
+1.  Python APIs shipped by `tfx` and `ml-pipelines-sdk` packages shall be ready
+    for 1.0 release and approximately SemVer compliant. Specifically:
+    1.  Public Python APIs, Protobufs and RPC interfaces will be backward
+        compatible within all 1.x releases, with the exception of:
+        1.  Experimental APIs: either with `experimental` in its symbol/path, or
+            explicitly called out in PyDoc.
+            1.  We propose only use `__init__.py` for public APIs. PyDoc can be
+                generated that way.
+        2.  Internal APIs: Any other API which is not covered by our public API
+            rules listed above are considered internal to the library, even if
+            they do not have "internal" or underscore prefix. These APIs should
+            not be directly used by users.
+        3.  C++ APIs: none of the C++ APIs from any related libraries or
+            repository will be considered public or stable. We will call this
+            out in the release log and README of each project.
+    2.  For Artifact Types (as well as evolution): requirements are being
+        collected and will be discussed separately.
+    3.  Artifact Payload: Generally required backward compatibility, no
+        guarantee on forward compatibility
+        1.  Exception on _Model_: TensorFlow SavedModel has its own
+            compatibility rules, and other model formats are not guaranteed with
+            any compatibility.
+2.  Other packages are encouraged to follow above compliance requirements,
+    especially if they want to graduate into a 1.0 version. This is eventually
+    call of the library owner, with the exception for:
+    1.  Public APIs, Protobuf definitions and RPC interfaces which are used or
+        exposed by either `tfx` package, or its examples/templates shall follow
+        above rules, so tfx can achieve its compliance goal. Some of the
+        underlying libraries will go to 1.0 with TFX, but that is not required
+        by the scope of this doc. Notable exceptions include but not limited to:
+        1.  `struct2tensor` will not go to 1.0 because it is an optional
+            dependency for TFX and only used in recent examples, which are
+            experimental themselves.
+
+See sections below for a more comprehensive list.
+
+##### LifeSpan of Major Version {#lifespan-of-major-version}
+
+LifeSpan of a major version is the difference (in days) between two major
+versions. This is an important factor because all minor versions within the same
+major version are required to be backward compatible, and certain cleanups can
+only be completed in a new major version.
+
+We propose a minimum **12 months lifespan** for TFX 1.0, which shall be the
+minimum for future major versions. When we are close to the end of that, we will
+revisit our tech debts and determine if we want to extend it.
+
+### Compatibility with Key External Libraries {#compatibility-with-key-external-libraries}
+
+The following external libraries are considered key dependencies, i.e we
+definitely need them for key CUJs, and we don’t see us moving away from them in
+the foreseeable future. This section briefly discusses how this affects our
+compatibility guarantees.
+
+It is worth noting that among all these key dependencies, only TensorFlow has
+Artifact compatibility items, and everything else has only API compatibility.
+This is helpful for our effort, because we can do future major version upgrades
+(e.g, from Beam 2 to 3, if/when that happens) in a way hidden from our users, at
+least for all the first party components which we own.
+
+#### TensorFlow {#tensorflow}
+
+Because TensorFlow Transform (TFT) still relies on internal APIs of TensorFlow
+which is not subject to compatibility guarantees, each version of TFX libraries
+is only guaranteed to work with one version of TensorFlow.
+
+Meanwhile, there are users who need to use a different version of TensorFlow in
+their training code, therefore we anticipate a need to mix TFX components which
+link with different TensorFlow versions in the same pipeline. This is possible
+in a containerized deployment, because each component could be configured with
+different container images. In order to process Artifacts produced from
+different TensorFlow versions, we will rely on the guarantees provided by
+TensorFlow on SavedModel
+([source](https://www.tensorflow.org/guide/versions#compatibility_of_savedmodels_graphs_and_checkpoints)),
+most importantly:
+
+*   _In compliance with [semver](https://semver.org/), SavedModels written with
+    one version of TensorFlow can be loaded and evaluated with a later version
+    of TensorFlow with the same major release._
+*   _"Any SavedModel supported in TensorFlow major version N can be loaded and
+    executed with TensorFlow major version N+1"_
+
+Because we do not guarantee forward compatibility on our Artifacts payload, this
+means that if a component (i.e, _Trainer_) needs to use a different version of
+TensorFlow, then all _ancestor_ components which produced SavedModel (i.e,
+`Transform`) must use the same or older version of TFX.
+
+##### TensorFlow Enterprise 1.15 {#tensorflow-enterprise-1-15}
+
+The LTS version of TensorFlow Enterprise version poses a special challenge,
+because the 1.15 distribution is expected to be supported for three years
+starting from Oct 30, 2019. We propose to make the TFX 1.0 minor the only
+supported version for TensorFlow 1.15, and test this combination to the maximum
+extent possible. Any future versions of TFX 1.x will not be supported with TFE
+1.15.
+
+#### Apache Beam {#apache-beam}
+
+Apache Beam mostly follows semantic versioning: breakages are rare and we have
+only seen one occurrence related to type hints once in the entirety of 2020.
+Therefore we propose the following for all TFX 1.x versions:
+
+1.  Each version of TFX is guaranteed to be compatible with latest version of
+    `apache-beam` at release time;
+2.  In addition, `tfx`'s `install_requires` will allow future minor versions of
+    `apache-beam` in the same major version, i.e, `apache-beam>=2.27,<3`.
+3.  When a breaking change to above from `apache-beam` is detected, TFX team
+    will create patch versions to either (a) cherry-pick a fix, or (b) lower the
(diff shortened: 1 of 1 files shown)


rfcs/20210307-tfx-api-artifact-compat.md
```diff
@@ -0,0 +1,583 @@
...
+different executor to an official TFX component. So far we have found two
+concrete use cases:
+
+1.  Google-cloud optimized tuner and trainer;
+2.  Customizing ExampleGen component.
+
+Because this process requires users to understand many details of internal
+implementations of an official TFX component, we plan to deprecate this support
+externally and declare it internal only. We will only support currently known
+use cases listed above.
```
casassg, line 295: What would then be the replacement if we want to customize the execution logic for Trainer (ex: launching a K8s based trainer on a custom pod). Would there be some way to customize this or would it require overwriting the Trainer component (and therefore giving up backwards compatibility)

rfcs/20210307-tfx-api-artifact-compat.md
```diff
@@ -0,0 +1,583 @@
...
+*   KubeflowV2DagRunner
+
+The following Dag runners will remain experimental:
+
+*   AirflowDagRunner
+    *   Honestly speaking, support of this one is on the fence. Due to lack of
+        proper configuration for a distributed backend support, it’s not much
+        more powerful than the LocalDagRunner when used out-of-box. However,
+        some partner teams managed to connect this to a distributed backend.
+        Until that is supported in TFX, we don’t think this one is stable.
```
casassg, line 353: > Distributed backend

Airflow does support quite well distributing workloads via the KubernetesExecutor, Celery or even KEDA. Not sure what this is referring to. 

Note that Airflow usage is quite large at this point in the idustry with many managed options to make it distributed (including Cloud Composer). 

Also it does allow for better customization than the Kubeflow integration as of today (retries, failure callbacks...)

rfcs/20210307-tfx-api-artifact-compat.md
```diff
@@ -0,0 +1,583 @@
...
+
+The following dag runners are slated to be deprecated and therefore won’t
+comply:
+
+*   BeamDagRunner: this will be replaced by the LocaDagRunner
```
casassg, line 358: is the idea to make LocalDagRunner parallelizable for 1.X? (given that Beam is parallizable locally while currently LocalDagRunner it is not).

rfcs/20210307-tfx-api-artifact-compat.md
```diff
@@ -0,0 +1,583 @@
...
+All data model schemas for Sqlite and MySQL backends shall be backward
+compatible with the MLMD schema migration utilities.
+
+### TFX-BSL & TFXIO {#tfx-bsl-&-tfxio}
+
+Only code in the
+_[‘public’](https://github.com/tensorflow/tfx-bsl/tree/master/tfx_bsl/public)_
+module in TFX-BSL is considered public to external users.
+
+#### TFXIO Usage {#tfxio-usage}
```
casassg, line 409: Would be nice to have the TFXIO API be tweakable and stable on API,

We are checking alternative data formats due to tf.examples sometimes not being easy to work with on wide-row datasets or to work with frameworks outside of Tensorflow (for example: XgBoost may not work great w Tf.example). 

Would be nice to have TFXIO be a public interface on TFX where we can inject custom TFXIO classes to the pipeline as needed (similar to DataView).

rfcs/20210307-tfx-api-artifact-compat.md
```diff
@@ -0,0 +1,583 @@
...
+        some partner teams managed to connect this to a distributed backend.
+        Until that is supported in TFX, we don’t think this one is stable.
+
+The following dag runners are slated to be deprecated and therefore won’t
+comply:
+
+*   BeamDagRunner: this will be replaced by the LocaDagRunner
+*   KubeflowDagRunner: this dag runner implementation based on KFP v1 API will
+    be replaced by the V2 dag runner above.
+
```
casassg, line 361: Where does InteractiveContext fit here?

rfcs/20210307-tfx-api-artifact-compat.md
zhitaoli: We don't have such a plan right now, although making such a change should be a net improvement w/o breaking change.

Do you have a concrete use case of a parallelized LocalDagRunner?

@charlesccychen

rfcs/20210307-tfx-api-artifact-compat.md
zhitaoli: @brills to comment on TFXIO API stability.

rfcs/20210307-tfx-api-artifact-compat.md
casassg: Given most of our pipelines just trigger remote jobs, allowing us to trigger several dataflow jobs in parallel for some pre-processing would be nice.

casassg:
Another question: Whats the compatibility between Component and Executor versions. If I define a pipeline in version 1.0.0, will it be possible to have garantees that if the TFX version is 1.X will be able to execute the component nicely? This comes to mind on versioning docker images for Kubeflow or AI Platform Training. Will we need to keep version parity between environment defining the pipeline and environment executing components.

rfcs/20210307-tfx-api-artifact-compat.md
zhitaoli: > Not sure what this is referring to.

This is mostly referring our current implementation, which only uses default PythonOperator to run nodes locally.

If there enough community interest, we are happy to partner with the OSS community to build up the Airflow orchestration and utilize all the benefits you mentioned above, but I don't think we have enough resources and expertise in TFX team to do this ourselves.

casassg: I mean, it works already, you just need to configure Airflow to use KubernetesExecutor. No need to change any part of the integration (which is why it confuses me)

rfcs/20210307-tfx-api-artifact-compat.md
```diff
@@ -0,0 +1,583 @@
...
+#### Apache Beam {#apache-beam}
+
+Apache Beam mostly follows semantic versioning: breakages are rare and we have
+only seen one occurrence related to type hints once in the entirety of 2020.
+Therefore we propose the following for all TFX 1.x versions:
+
+1.  Each version of TFX is guaranteed to be compatible with latest version of
+    `apache-beam` at release time;
+2.  In addition, `tfx`'s `install_requires` will allow future minor versions of
+    `apache-beam` in the same major version, i.e, `apache-beam>=2.27,<3`.
```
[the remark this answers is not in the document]
zhitaoli: TFX tends to stay up-to-date on our beam dependency, so that we can make sure latest optimizations and improvements from beam is used. My experience tells me that older version of beam usually works, but tends to fall into a under-tested scenario.

> cross language upgrades (JVM and Python)

Can you clarify this part a bit? My understanding is that we are requiring a python package `apache-beam` which by large does not force JVM part, and it generally supports most common Python versions, so I'm not exactly sure what you referred to here?

rfcs/20210307-tfx-api-artifact-compat.md
```diff
@@ -0,0 +1,583 @@
...
+
+Because TensorFlow Transform (TFT) still relies on internal APIs of TensorFlow
+which is not subject to compatibility guarantees, each version of TFX libraries
+is only guaranteed to work with one version of TensorFlow.
+
+Meanwhile, there are users who need to use a different version of TensorFlow in
+their training code, therefore we anticipate a need to mix TFX components which
+link with different TensorFlow versions in the same pipeline. This is possible
+in a containerized deployment, because each component could be configured with
+different container images. In order to process Artifacts produced from
```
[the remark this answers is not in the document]
zhitaoli: You are right, this is a future work, but we plan to finish this around TFX 1.0. Updated the tense in the doc.

rfcs/20210307-tfx-api-artifact-compat.md
```diff
@@ -0,0 +1,583 @@
...
+        the version of an Artifact equals to the _&lt;major>.&lt;minor>_ version
+        of its producer component. This version will be further used as
+        consumption hint for these Artifacts, to correctly handle version
+        upgrades.
+*   **No Forward compatibility guarantee on payload**: while the system will not
+    reject passing Artifacts produced from newer versions of TFX to an old
+    version of a component, we do not guarantee this will work. A warning will
+    be generated in this case.
+*   Resolvers and importers would only resolve Artifacts produced with same or
+    older versions of the library, by default.
```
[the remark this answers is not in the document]
zhitaoli: What kind of warning you are looking for? If 1) all candidate artifacts are from newer version of the library and 2) the semantic  of the resolver and importer used here cannot tolerate empty result, we would see a hard error.

It seems like the only warning we could add is something like following in resolver or importer execution:

```python
if skipped_artifacts:
  logging.warn('Skipped %d artifacts because they are produced by future version of the component', len(skipped_artifacts)')
```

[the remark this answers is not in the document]
casassg: Yeah, that would be enough 👍

rfcs/20210307-tfx-api-artifact-compat.md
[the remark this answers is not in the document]
casassg: Beam on 2.28 introduces the concept of cross language transforms. Due to our nature of monorepo (similar to Google's I believe) we only have a single version of Beam for our code, which means that we need to keep both java and Python based Beam dependencies in sync.

> My experience tells me that older version of beam usually works, but tends to fall into a under-tested scenario.

I agree, but setting it as a transtive dependency forces us to upgrade Beam if we want to upgrade TFX, which is largely undesired as it makes it hard.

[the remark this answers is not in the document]
zhitaoli: The primary concern from our side is how many old versions of Beam we have to support. Upgrading Beam versions would allow us to stay on top of latest improvements in Beam.

While a common answer would be "don't upgrade TFX until you are ready to upgrade both TFX and Beam", I understand you also want to use newer features in TFX.

I can see two high level paths from here:

Option 1 (slightly preferred by us): TFX will still use latest Beam, however we can try to open up all our release/packaging scripts and leave enough options for overriding Beam versions, so you can overwrite the version of beam in your own deployment. In this option, you are essentially depending on TFX on source and supporting yourself on the particular combination of TFX + Beam.  We still encourage you to upgrade Beam to as latest as possible. For this to work, I highly encourage you to use a containerized deployment.

Option 2 (not really preferred but with a precedence): we will try our best to make some versions of TFX to work with a (LTS) version of Beam. This is the special arrangement we are targeting with TensorFlow 1.15. Because there is a LTS version of that, we will try to make one version in TFX 1.x (likely 1.0) to work well enough with TensorFlow 1.15. This is mostly making sure there is a very stable combination which sees security backports, and I would not recommend users who wants to stay on top of latest features to use this. To make this work, we would need Beam team to commit to backport security fixes to this LTS version of Beam.

If any of this seems acceptable, I can go back to evaluate its concrete feasibility. WDYT?

rfcs/20210307-tfx-api-artifact-compat.md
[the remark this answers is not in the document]
zhitaoli: Added to the doc.

rfcs/20210307-tfx-api-artifact-compat.md
```diff
@@ -0,0 +1,583 @@
...
+2.  Importing a user provided container image
+    *   this process and APIs will be backward compatible;
+3.  Defining ComponentSpec, ExecutorSpec and subclassing with component class
+    *   This is a very complex user interface, and arguably exposes too much
+        internal implementation details. Moreover, it’s proven that this is
+        quite difficult to do for our users. A possible improvement proposed is
+        to combine the component and executor classes, and have the convention
+        be that users subclass a base component's class to extend or override
+        functionality. Until this is settled, this API will not comply with
+        backward compatibility.
```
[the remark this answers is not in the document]
zhitaoli: @charlesccychen to see what we can try to finalize with current TFX 1.0 timeline.

rfcs/20210307-tfx-api-artifact-compat.md
zhitaoli: If you are customizing a trainer and launching k8s pod, I feel that a Python or container based custom component would not be much harder than fitting a custom executor in place of our Trainer component. And those two experiences should receive more support.

If you must use this path and design your executor to take the same conventions as the CAIP executors, you might receive some warnings saying you are using an unsupported executor class, but we have no plan to remove this functionality so your code would still work post 1.0.

So, if you already have this executor, we could try to make it work. But if it's undeveloped yet, I would recommend you to use Python or container based solution. WDYT?

@1025KB to double check above promise can be preserved.

rfcs/20210307-tfx-api-artifact-compat.md
zhitaoli: Thanks for pointing that out. We are not really aware of this.

rfcs/20210307-tfx-api-artifact-compat.md
zhitaoli: https://www.tensorflow.org/tfx/api_docs/python/tfx/orchestration/experimental/interactive/interactive_context/InteractiveContext

zhitaoli:
> Another question: Whats the compatibility between Component and Executor versions. If I define a pipeline in version 1.0.0, will it be possible to have garantees that if the TFX version is 1.X will be able to execute the component nicely? This comes to mind on versioning docker images for Kubeflow or AI Platform Training. Will we need to keep version parity between environment defining the pipeline and environment executing components.

If I understand it correctly, your example is asking if a TFX pipeline authored with 1.0.0 can use container images from TFX 1.1.0, or vice versa.

To guarantee that example to work, we would need to commit that both TFX IR and Container entrypoint interface are fully stable, both of which was out of our initial scope, because they are still seeing active development.

rfcs/20210307-tfx-api-artifact-compat.md
casassg: So we should not expect InteractiveContext to graduate from experimental on TFX 1.X?

rfcs/20210307-tfx-api-artifact-compat.md
casassg: >So, if you already have this executor, we could try to make it work. But if it's undeveloped yet, I would recommend you to use Python or container based solution. WDYT?

It's currently undeveloped, will take the guidelines as pointed here ty!

rfcs/20210307-tfx-api-artifact-compat.md
casassg: Would love to sync on this and explain a bit.

casassg (commented):
>To guarantee that example to work, we would need to commit that both TFX IR and Container entrypoint interface are fully stable, both of which was out of our initial scope, because they are still seeing active development.

Copy that. May be interesting to call it out in the document.

rfcs/20210307-tfx-api-artifact-compat.md
```diff
@@ -0,0 +1,584 @@
...
+
+1.  Providing a PTransform for converting input source to examples by inheriting
+    from BaseExampleGenExecutor and overwriting the abstract method.
+    *   This is recommended and will be backward compatible.
+2.  Use the above custom Component or custom Executor for use cases like join.
+    *   This is a very complex user interface, and arguably exposes too much
+        internal implementation details (driver:external file detection,
+        span/version/split etc). Moreover, it’s proven that this is quite
+        difficult to do for our users. We plan to mark this process as
+        experimental in 1.0, and not comply with backward compatibility.
```
casassg, line 282: A bit worried on this here for 2 reasons: 

1) **ExampleGen dependency on external systems**
ExampleGen is the main component that depends (implicitly or explicitly) on external systems from TFX. A lot of times you will need some custom code to check on external systems or do some custom metadata preparation before actually running the Executor. This change here, means that the only way to do that is via an experimental API. 

Alternatives:
- Pushing this logic to the scheduler/orchestrator: Feasible but reduces flexibility/portability of TFX pipeline
- Pushing this logic to Executor: Marked as experimental in this RFC as well.

While most users will not use this, advanced users will require this customization.

2) Custom output data formats: As discussed, tf.example may not be the best format for all use cases (ex: recommender systems). As such, you would need to generate custom ExampleGens that output custom formats as needed. This can be done today by customizing the Excecutor on ExampleGen to change output format and using Data views.

Alternatives:
- Push this outside of TFX to external system: Loss of compatibility/portability + Still requires a custom component to import the dataset inside the TFX pipeline.

CC @aoen

rfcs/20210307-tfx-api-artifact-compat.md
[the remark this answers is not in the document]
casassg: I think option 1 could work if we have a way to also run TFX end to end tests for the overwritten version of beam (to make sure its stable). 

Option 2 may be needed down the road, but this can be a separate discussion with Beam team.

rfcs/20210307-tfx-api-artifact-compat.md
zhitaoli: No we are not. We don't want to rush the graduation of experimental features, and this one hasn't picked up enough internal or production uses yet to qualify our graduation standard.

casassg: I see. Would be interesting to know about graduation standard for transparency. We have quite large internal usage of this.

rfcs/20210307-tfx-api-artifact-compat.md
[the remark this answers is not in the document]
zhitaoli: After some more thoughts, I think the `beam_pipeline_args` shouldn't be access from the custom component definition part of API anyway, here's my reasoning:

1. beam_pipeline_args are user input anyway, and a custom component also requires user input;
2. we are in the process of making this configurable to each component, and several non-beam based component like Trainer won't even see this flag after refactoring in https://github.com/tensorflow/tfx/commit/be20d084f2875655e6b94a908206ab8bf6ff278e

WDYT?

zhitaoli:
> > To guarantee that example to work, we would need to commit that both TFX IR and Container entrypoint interface are fully stable, both of which was out of our initial scope, because they are still seeing active development.
> 
> Copy that. May be interesting to call it out in the document.

Done.

rfcs/20210307-tfx-api-artifact-compat.md
zhitaoli: @1025KB

rfcs/20210307-tfx-api-artifact-compat.md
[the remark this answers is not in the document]
casassg: This makes sense. Given https://github.com/tensorflow/tfx/pull/3584 I think it makes sense to not surface it anymore given beam_pipeline_args will be deprecated.

rfcs/20210307-tfx-api-artifact-compat.md
casassg: Proposed rewording:

```
Due to low utilization of the DagRunner, we want to avoid calling it stable. Support will continue to graduate it into production as utilization evolves.
```

Mostly want to avoid calling out distributed backend support. Cloud Composer does support distributed backend  and such, so would avoid saying `Due to lack of proper configuration for a distributed backend support`. I'm okay with the reason being: `Due to low usage and it not being a production usage yet that we dont believe its stable`.

rfcs/20210307-tfx-api-artifact-compat.md
casassg: I think its fine to leave as is. Would be nice to have a public facing ticket to track the work to make it parallelizable.

For prioritization, LocalDagRunner may not be considered a feasible option unless its parallelizable.

rfcs/20210307-tfx-api-artifact-compat.md
casassg: CC @newtonle

rfcs/20210307-tfx-api-artifact-compat.md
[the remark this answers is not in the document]
ConverJens: @zhitaoli @casassg Regarding the beam args, it is crucial that the new logic for defining this on component level also supports fetching values from secrets. Otherwise the serialised component will contain sensitive args (e.g. s3 passwords) in clear text.

I have created a PR that solves this for the current version of TFX: https://github.com/tensorflow/tfx/pull/3455.

rfcs/20210307-tfx-api-artifact-compat.md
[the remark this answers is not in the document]
zhitaoli: Will add following blurbs to the doc:


```
For users who wants to use older versions of `apache-beam` and `tfx` together,
while TFX team cannot directly support them in our released packages, we plan to
refactor our build scripts so that beam dependency can be overridden by
themselves, so they can build Python packages and container images in a
self-service fashion.
```

rfcs/20210307-tfx-api-artifact-compat.md
zhitaoli: Accepted.

rfcs/20210307-tfx-api-artifact-compat.md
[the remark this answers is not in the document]
zhitaoli: > @zhitaoli @casassg Regarding the beam args, it is crucial that the new logic for defining this on component level also supports fetching values from secrets. Otherwise the serialised component will contain sensitive args (e.g. s3 passwords) in clear text.
> 
> I have created a PR that solves this for the current version of TFX: [tensorflow/tfx#3455](https://github.com/tensorflow/tfx/pull/3455).

I think it's technically decoupled from where the `beam_pipeline_args` are configured, pipeline or component.

Left a comment on tensorflow/tfx#3455 too for my concern about that direction.

rfcs/20210307-tfx-api-artifact-compat.md
zhitaoli: Circling back to this: do you think TFXIO stability is a blocker for other API to enter v1 stability? We can do it in a follow up version as long as it does not require we change other parts of the system, which I think it won't?

casassg: sgtm!

rfcs/20210307-tfx-api-artifact-compat.md
```diff
@@ -0,0 +1,600 @@
...
+*   _pipeline_
+*   _run_
+
+_The template_ command group will remain experimental.
+
+### ML-Metadata (MLMD) {#ml-metadata-mlmd}
+
+We are proposing that MLMD project enters the same 1.0 readiness and
+compatibility as TFX. The rest of this section will discuss various aspects of
+MLMD.
```
casassg, line 397: nit: how about context types? Will those be stable (currently they are privately defined - see [this](https://github.com/tensorflow/tfx/blob/master/tfx/orchestration/metadata.py#L65) ). Mostly asking as TFX Addons MLMD Client may need to rely on some those to retrieve them

casassg (approved):
lgtm. Added a small question, but not blocking for release as we can figure out later and make it stable later.

rfcs/20210307-tfx-api-artifact-compat.md
zhitaoli: Good question. Based on my discussion with 
@ruoyu90 this is highly coupled with IR, so we'd view them as the same thing  and probably graduate them in a lock step in the future.

casassg:
🎉

ematejska (approved):

Issue #280, opened 2020-08-19
Broken link for "TF API Owners"


lithuak:
[This "Overview" section](https://github.com/tensorflow/community/blob/master/governance/api-reviews.md#overview) contains the broken link: [TF API Owners](https://github.com/orgs/tensorflow/teams/api-owners)

ematejska:
This link seems works for me (at least now).  Please re-open if that still doesn't work.

adriangb:
It's broken. Doesn't work for me. Maybe it's an internal link?

allenlavoie:
It's possible the visibility for that team is set incorrectly (only people in the TF github org?). It doesn't look like a "secret" team to me; it's listed under visible teams. Opening the link in an incognito window does ask me to sign in, which seems wrong.

adriangb:
If you can't get to it in an incognito window, then it's not a public link. I'm seeing the same thing in incognito.

bhack:
@allenlavoie Probably the visibility concept is different in Github. See https://docs.github.com/en/github/setting-up-and-managing-organizations-and-teams/about-teams#team-visibility

allenlavoie:
Ah, good catch "Visible teams can be viewed and @mentioned by every organization member."; so the exception indicates the rule that non-organization members can't view them. We should probably remove the link and find another way to communicate the information.

@ematejska I don't think anyone else can re-open bugs here.

bhack:
I think also @theadactyl can reopen it

bhack:
I propose a solution with https://github.com/tensorflow/community/pull/379

theadactyl:
cc @ematejska can you take another look?

PR #379, opened 2021-04-02
Remove restricted link

bhack:
Fixes https://github.com/tensorflow/community/issues/280

### DIFF
--- a/governance/api-reviews.md
+++ b/governance/api-reviews.md
@@ -4,7 +4,7 @@
 
 This is an attempt to gather commonly discussed topics when doing API
 reviews. It’ll hopefully be a useful resource to both API owners and people
-proposing API changes.  [TF API Owners](https://github.com/orgs/tensorflow/teams/api-owners) 
+proposing API changes.  TF API Owners team (@tensorflow/api-owners) 
 meet twice weekly to discuss changes. We try to get to PRs on the next meeting, 
 but we don’t always make it all the way through. If your change is particularly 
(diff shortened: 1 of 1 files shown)


adriangb:
I tested this in https://github.com/tensorflow/tensorflow/pull/48125#issuecomment-812543799, it didn't seem to tag a team

bhack:
You don't see the link  generated cause the link target the same team page. My question is: are they notified? Generally a reviewer is a TF member so the team could be mentioned by the reveiwer.

adriangb:
> My question is: are they notified?

I agree, this is the important part. If they are notified but no link is generated that is fine, as long as it is documented (I would suggest in this PR). But I think someone should check if that is the case, eg. ping someone manually and see if they got a notification for https://github.com/tensorflow/tensorflow/pull/48125#issuecomment-812543799.

adriangb:
I'm commenting here to avoid polluting the conversation in tensorflow/tensorflow#48125.

Interestingly, in your comment I can see that it is a link, but it is greyed out and indeed leads me to a 404. In my comment, it is not even a link. At least this is _some_ confirmation that it works, even if it may require a TF member to make the comment.

image

bhack:
I don't think that they are mentionable by non members. I think it is a Github feature to protect unwanted potential group notification.
So I've added a reminder.

theadactyl:
cc @ematejska

ematejska (approved):

Issue #223, opened 2020-03-31
Ask contribution to Tensorflow addons for general scope utils, loss, layers, ops


bhack:
As we have just refreshed the model repo as model garden I would enforce the contributions policies of generale use (or already established in literature utils, losses, layers, ops to be contribute more systematically in Tensorflow/addons instead to be embedded or duplicated in the model repos.
E.g:
https://github.com/tensorflow/addons/issues/361#issuecomment-606856705
https://github.com/tensorflow/addons/issues/1364#issuecomment-602287778
https://github.com/tensorflow/addons/issues/1366
https://github.com/tensorflow/addons/issues/1226#issuecomment-599483920
https://github.com/tensorflow/addons/issues/1226#issuecomment-600204811

 /cc @ewilderj  @facaiy @seanpmorgan 

# Policies to be enforced with a PR
https://github.com/tensorflow/models/blob/master/CONTRIBUTING.md
https://github.com/google-research/google-research/blob/master/CONTRIBUTING.md
https://github.com/google/automl/blob/master/CONTRIBUTING.md

bhack:
See also the approved RFC https://github.com/tensorflow/community/blame/master/rfcs/20190802-model-garden-redesign.md#L60-L61

bhack:
I will add in the target contributions repos the new keras-cv and keras-nlp.

bhack:
E.g. See https://github.com/tensorflow/models/pull/8962#pullrequestreview-457058720

theadactyl:
@bhack still interested in this?

bhack:
In general this is a still valid topic but:

- For TF Addons we could close this as we have:
https://github.com/keras-team/keras-cv/issues/74
https://github.com/keras-team/keras-nlp/issues/11

- More in general for the ecosystem I could add a reference to this ticket in  https://github.com/tensorflow/community/pull/384

PR #384, opened 2021-04-21
CONTRIBUTING.md template

bhack:
This is a Draft CONTRIBUTING.md template.
It is part of an effort to standardize some early access info for Github community health files across the TensorFlow ecosystem repositories.


governance/SIG-contributing-template.md
```diff
@@ -0,0 +1,50 @@
...
+## DEBUG
+How to debug your code
+
+## CI
+Quickly describe your CI
+
+## CODEOWNERS
+Describe how codeownership and maintainership in the repo:
+* How to be a Codeowner
+* How to be a Maintainer
```
yarri-oss, line 45: I would like to suggest we consider adding this as a separate file (CALL_FOR_MAINTAINERS.md?) due to the importance of the role.

bhack: Good, we could put here just a like to an external markdown. 
I don't know I we could open a new thread later about standardizing a little bit the `CALL_FOR_MAINTAINERS` process/roles. 

E.g. Debian has historically a quite regulated [Maintainer process](https://wiki.debian.org/DebianMaintainer). This is just to make a reference example but it is too complex for us.

bhack: /cc I think that also @mihaimaruseac could be interested in the Codeowner/Maintainer status and process.

ematejska:
@bhack Is this ready for review or still a draft?

bhack:
I don't know how to colllect early feedback.
What do you think @theadactyl @yarri-oss?

theadactyl:
Will review.

ematejska:
@theadactyl Can you review?

bhack:
Just a reminder for the reviewers. We need to figure out also how to handle topics mentioned in https://github.com/tensorflow/community/issues/223

I suppose we could add a general point in this template if we could define some general rules on how to handle duplicated/conflicting  contributions.

PR #389, opened 2021-05-20
Add pluggable device profiler support

jzhoulon:
This PR aims at providing TensorFlow profiler support for PluggableDevice


yisitu:
> This RFC will be open for comment until Monday, June 4th , 2021.

Isn't June 4th 2021 a Friday?

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,441 @@
...
+
+## **Motivation**
+
+Performance is a key consideration of successful ML research and production solutions. TensorFlow profiler provides a set of good tools to help user better understand the hardware resource consumption(time and memory) of various TensorFlow operations(ops) as well as perforamnce bottlenecks of TensorFlow models. [PluggableDevice](https://github.com/tensorflow/community/pull/262) provides a mechanism which allows third-party devices to be modularly integrated to current TensorFlow runtime. However, current TensorFlow does not provide a stable application binary infterface(ABI) for registering a new pluggable device specific profiler in a modular way. We propose a C API wrapper of methods in [ProfilerInterface](https://github.com/tensorflow/tensorflow/blob/0a3773ed7b4c1fc60486dddfacbe9a5cbf2b2bdd/tensorflow/core/profiler/lib/profiler_interface.h#L33) as an ABI-stable way to register a custom profiler. The pluggable profiler discovery and initialization are transparent to end users. As long as the profiler plugin libraries follow the design described in this RFC, it can be plugged to TensorFlow proper and register a new profiler into ProfilerFactory.
```
yisitu, line 17:
```suggestion
Performance is a key consideration of successful ML research and production solutions. TensorFlow profiler provides a set of good tools to help user better understand the hardware resource consumption(time and memory) of various TensorFlow operations(ops) as well as performance bottlenecks of TensorFlow models. [PluggableDevice](https://github.com/tensorflow/community/pull/262) provides a mechanism which allows third-party devices to be modularly integrated to current TensorFlow runtime. However, current TensorFlow does not provide a stable application binary interface(ABI) for registering a new pluggable device specific profiler in a modular way. We propose a C API wrapper of methods in [ProfilerInterface](https://github.com/tensorflow/tensorflow/blob/0a3773ed7b4c1fc60486dddfacbe9a5cbf2b2bdd/tensorflow/core/profiler/lib/profiler_interface.h#L33) as an ABI-stable way to register a custom profiler. The pluggable profiler discovery and initialization are transparent to end users. As long as the profiler plugin libraries follow the design described in this RFC, it can be plugged to TensorFlow proper and register a new profiler into ProfilerFactory.
```

yisitu (changes requested):
First round of edits.

penpornk:
If you are interested in attending the design review meeting for this RFC, here are the call details:

Time: Thursday, June 3rd, 2021 from 3-4pm PT

cc: @kulinseth @wchao1115 @reedwm

wchao1115:
yeah, I'll be there.

jzhoulon:
> > This RFC will be open for comment until Monday, June 4th , 2021.
> 
> Isn't June 4th 2021 a Friday?

Thanks for the notify, I have fixed it, and thanks for helping edit the document!

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,440 @@
...
+  // Saves collected profile data into XSpace and serializes it to the buffer.
+  // After this or the overload above are called once, subsequent calls might
+  // return empty data.
+  void (*collect_data_run_metadata)(const TP_Profiler* profiler, uint8_t* buffer, size_t* size_in_bytes, TF_Status* status)
+
+  // Saves collected profile data into run_metadata and serializes it to the buffer.
+  // After this or the overload below are called once, subsequent calls might
+  // return empty data.
+  void (*collect_data_xspace)(const TP_Profiler* profiler, uint8_t* buffer, size_t* size_in_bytes, TF_Status* status);
+};
```
yisitu, line 109:
```suggestion
} TP_ProfilerFns;
```

yisitu (changes requested):

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,440 @@
...
+} TP_Profiler;
+
+#define TP_PROFILER_STRUCT_SIZE TF_OFFSET_OF_END(TP_Profiler, type)
+
+typedef struct TP_ProfilerFns {
+  size_t struct_size;
+
+  void* ext; // reserved for future use.
+
+  void (*start)(const TP_Profiler* profiler, TF_Status* status); // Starts profiling
```
yisitu, line 96:
```suggestion
  // Starts profiling
  void (*start)(const TP_Profiler* profiler, TF_Status* status);
```

yisitu (changes requested):

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,440 @@
...
+  class PluggableTracer : public ProfilerInterface {
+   public:
+     explicit PluggableTracer(std::vector<PluginTracerInterface> plugin_interfaces)
+             : plugin_interfaces_(plugin_interfaces) {}
+
+     ~PluggableTracer() override {}
+
+     Status Start() override {
+       for (auto& profiler_interface : plugin_interfaces_) {
+         profiler_interface.DoStart().IgnoreError();
```
yisitu, line 307: (Here and below) Errors should not be ignored, and instead should be Update() and returned.

https://abseil.io/docs/cpp/guides/status#keeping-track-of-the-first-error-encountered

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,440 @@
...
+}
+
+void profiler_collect_data_xspace(const TP_Profiler* profiler, uint8_t* buffer, size_t* size_in_bytes, TF_Status* status) {
+  Xspace xspace = get_my_xspace(); // Plugin generates Xspace based on collected profiler data.
+  *size_in_bytes = xspace.ByteSizeLong(); // get the size of Xspace
+  if (buffer == nullptr) {
+    return; // TensorFlow will first get the size of Xspace, then allocate the big enough buffer and pass it to plugin for retrieving Xspace.
+  }
+  xspace.SerializeToArray(buffer, xspace.ByteSizeLong());
+}
```
yisitu, line 383: 1) Need to check the return value of `SerializeToArray()`
2) Need to pass the user's buffer size to SerializeToArray().
3) Need to set the out parameter `status`.


```suggestion
void profiler_collect_data_xspace(const TP_Profiler* profiler, uint8_t* buffer, size_t* size_in_bytes, TF_Status* status) {
  Xspace xspace = get_my_xspace(); // Plugin generates Xspace based on collected profiler data.
  size_t buffer_size_in_bytes = * size_in_bytes;
  *size_in_bytes = xspace.ByteSizeLong(); // get the size of Xspace
  if (buffer == nullptr) {
    return; // TensorFlow will first get the size of Xspace, then allocate the big enough buffer and pass it to plugin for retrieving Xspace.
  }
  bool success = xspace.SerializeToArray(buffer, buffer_size_in_bytes);
 // TODO: set status to FAILED_PRECONDITION if success is false, OK otherwise.
}
```

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,440 @@
+# Modular TensorFlow profiler C API
```
yisitu, line 1: @deven-amd
Please have a look at this RFC. I think it would be cleaner to have all the GPU profilers consolidated behind a common interface.

yisitu (changes requested):

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,440 @@
...
+  int32_t major_version;
+  int32_t minor_version;
+  int32_t patch_version;
+
+  TP_Profiler* profiler; // output, set by plugin
+  TP_ProfilerFns* profiler_fns; // output set by plugin
+  // Clean up fields inside TP_Profiler that were allocated
+  // by the plugin. `profiler` itself should not be deleted here.
+  void (*destroy_profiler)(TP_Profiler* profiler); // out, set by plugin
+  void (*destroy_profiler_fns)(TP_ProfilerFns* profiler_fns);
```
yisitu, line 127:
```suggestion
  [in/out] Memory owned by core but attributes within are populated by the plugin.
  TP_Profiler* profiler;
  [in/out] Memory owned by core but attributes within are populated by the plugin.
  TP_ProfilerFns* profiler_fns;
  // [out] Pointer to plugin's `TP_Profiler` clean up function. 
  // Cleans up fields inside `TP_Profiler` that were allocated
  // by the plugin. `profiler` itself must not be deleted by the plugin.
  void (*destroy_profiler)(TP_Profiler* profiler);
  // [out] Pointer to plugin's `TP_ProfilerFns` clean up function. 
  // Cleans up fields inside `TP_ProfilerFns` that were allocated
  // by the plugin. `profiler_fns` itself must not be deleted by the plugin.
  void (*destroy_profiler_fns)(TP_ProfilerFns* profiler_fns);
```

yisitu (changes requested):

rfcs/20210513-pluggable-profiler-for-tensorflow.md
jzhoulon: Thanks for the comments, the consideration here is whether one plugin profiler failed will affect other plugin profiler, I see ProfilerSession, it seems it ignored all the profiler status. 
```
 61 
 62 Status ProfilerSession::CollectData(profiler::XSpace* space) {
 63   mutex_lock l(mutex_);
 64   TF_RETURN_IF_ERROR(status_);
 65 #if !defined(IS_MOBILE_PLATFORM)
 66   LOG(INFO) << "Profiler session collecting data.";
 67   for (auto& profiler : profilers_) {
 68     profiler->Stop().IgnoreError();
 69   }
 70 
 71   for (auto& profiler : profilers_) {
 72     profiler->CollectData(space).IgnoreError();
 73   }
 74 
 75   if (active_) {
 76     // Allow another session to start.
 77     profiler::ReleaseProfilerLock();
 78     active_ = false;
 79   }
 80 
 81   PostProcessSingleHostXSpace(space, start_time_ns_);
 82 #endif
 83 
 84   return Status::OK();
 85 }
```

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,440 @@
...
+<div align=center>
+<img src=20210513-pluggable-profiler-for-tensorflow/Xspace.png>
+</div>
+
+- RunMetadata:
+<div align=center>
+<img src=20210513-pluggable-profiler-for-tensorflow/RunMetadata.png>
+</div>
+
+To achieve the goal, this RFC extends the TensorFlow profiler class hierarchy to add a new profiler named `PluggableProfiler` which is built on top of a set of C APIs, all plugin writers who want to integrate their own device profilers to  current TensorFlow runtime only need to implement Profiler C APIs(shown as diagram of Architecture overview).
```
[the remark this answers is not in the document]
jzhoulon: done

rfcs/20210513-pluggable-profiler-for-tensorflow.md
yisitu: That is a bug in ProfilerSession. The correct behavior should be to `Update()` status for each profiler and not return early upon failure. This way we track the first error encountered and resume calling to other plugin profilers.

rfcs/20210513-pluggable-profiler-for-tensorflow.md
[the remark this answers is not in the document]
yisitu: Thank you.

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,440 @@
...
+    // DeviceType::GPU: only CPU/GPU will be profiled.
+    // DeviceType::TPU: only CPU/TPU will be profiled.
+    // DeviceType::PLUGGABLE_DEVICE: all pluggable devices with profiler enabled will be profiled. 
+    DeviceType device_type = 6;
+  ```
+  Due to `device_type` here is enum, we can't differentiate between multiple pluggable profilers, so we define a common device type `PLUGGABLE_DEVICE` for them, if `ProfileOptions` is configured with `PLUGGABLE_DEVICE` type, then all the registered pluggable profilers will be enabled.
+
+* **Plugin Profiler Initialization**
+
+  Core TensorFlow will load `TF_InitProfiler` from plugin's dynamic library installed under "…python_dir.../site-packages/tensorflow-plugins" and pass the handle to `InitPluginProfiler` to do initialization. Here we define a `PluginInterfaceFatory` to store all the registered pluggable profilers, all of pluggable profilers in this factory will be enabled if `device_type` in `ProfileOptions` is configured as `PLUGGABLE_DEVICE`.
```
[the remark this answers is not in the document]
jzhoulon: done, TFInitProfilerFn is expected to define in profiler_internal.h, which is not a public API

yisitu:
I would also like for us to discuss the design alternative of having each vendor inherit C++ class `ProfilerInterface`. Each vendor will own lower part of the stack; beginning from the C++ `ProfilerInterface` subclass, and bridge to C API (not standardized across vendors) and shared library binary.

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,440 @@
...
+
+#### Protobuf class
+Profiler uses `RunMetadata` and `XSpace` to store the performance data collected by backends. With these two data, TensorFlow's profiler tools can generate various views of performance, such as timeline, memory consumption, performance of every TensorFlow op and set of summaries. `RunMedata` and `XSpace` are C++ objects generated by protobuf toolchain with a predefined structure in [config.proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/config.proto#L721) and [xplane.proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/profiler/protobuf/xplane.proto#L9).
+On the plugin side, plugin will collect performance data with its own profiler infrastructures. When TensorFlow invokes `CollectData()`, plugin transforms the performance data to `RunMetadata` and `XSpace` and then serializes `RunMetadata` and `XSpace` to the buffer provided by TensorFlow. To successfully serialize the object, plugin writers should keep a copy of `config.proto` and `xspace.proto`, and make it exactly the same as that in the TensorFlow side.
+
+### Detailed API
```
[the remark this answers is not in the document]
jzhoulon: done,

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,440 @@
...
+  // return empty data.
+  void (*collect_data_run_metadata)(const TP_Profiler* profiler, uint8_t* buffer, size_t* size_in_bytes, TF_Status* status)
+
+  // Saves collected profile data into XSpace and serializes it to the buffer.
+  // If either this or `collect_data_run_metadata` have been called, subsequent calls might
+  // return empty data.
+  void (*collect_data_xspace)(const TP_Profiler* profiler, uint8_t* buffer, size_t* size_in_bytes, TF_Status* status);
+} TP_ProfilerFns;
+
+#define TF_PROFILER_FNS_STRUCT_SIZE TF_OFFSET_OF_END(TP_ProfilerFns, collect_data_xspace)
```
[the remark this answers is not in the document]
jzhoulon: done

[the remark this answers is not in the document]
jzhoulon: done

rfcs/20210513-pluggable-profiler-for-tensorflow.md
jzhoulon: Done

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,440 @@
...
+<div align=center>
+<img src=20210513-pluggable-profiler-for-tensorflow/Architecture.png>
+</div>
+
+### Versioning Strategy and Stability
+* **Profiler C API**
+    Version strategy of Profiler C API follows Semantic Versioning 2.0.0 ([semver](http://semver.org/)). Each release version has a format `MAJOR.MINOR.PATCH`, as outlined in [TensorFlow version compatibility](https://www.tensorflow.org/guide/versions#semantic_versioning_20). Struct size is used to track compatibility. More details can be found in [StreamExecutor C API Versioning Strategy RFC](https://github.com/tensorflow/community/blob/master/rfcs/20200612-stream-executor-c-api/C_API_versioning_strategy.md)
+* **XPlane and RunMetadata**
+   The compatibility of `XPlane` and `RunMetadata` between plugin and TensorFlow follows the same compatibility [rules](https://developers.google.com/protocol-buffers/docs/cpptutorial?hl=en#extending-a-protocol-buffer) and [guarantees](https://developers.google.com/protocol-buffers/docs/proto3?hl=en#updating) as protobuf library.
+
```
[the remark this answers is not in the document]
jzhoulon: Done

jzhoulon:
> I would also like for us to discuss the design alternative of having each vendor inherit C++ class `ProfilerInterface`. Each vendor will own lower part of the stack; beginning from the C++ `ProfilerInterface` subclass, and bridge to C API (not standardized across vendors) and shared library binary.

Hi, @yisitu  I don't quite understand this part, can you share more details so that we can evaluate it before the meeting. 
1. each vendor inherit C++ class `ProfilerInterface
    How vendor inherit C++ class `ProfilerInterface`, does vendors need to copy core/profiler/lib/profiler_interface.h to their plugin repo?
2. bridge to C API (not standardized across vendors) 
    How to bridge to the C API and why the C API is not standarized?
Thanks.

penpornk:
### Design Review Meeting Notes
@jzhoulon went through recently resolved comments then started a discussion on the [design alternative](https://github.com/tensorflow/community/pull/389#issuecomment-848427864).

@jzhoulon: Could you please clarify the design alternative?
@jbaiocchi: TF Profiler already has a plug-in system. This design adds another plug-in on top.
  * Forcing every vendor to conform to the same C API doesn’t seem advisable.
  * The alternative is to let each vendor own a C++ subclass of [ProfilerInterface](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/profiler/lib/profiler_interface.h;l=33;drc=122655424ec4e5842ad8981a70e4b7fa13c6103d) in core TensorFlow and a C API for the subclass to communicate with its corresponding profiler plug-in.
  * We don’t need to standardize the C API. Just the C++ part.

@jbaiocchi: What about a C++ boundary?
@aselle: Then people will have to recompile everything
  * C++ header-only file.
  * Specific compiler.

@jzhoulon: That wouldn’t work for us. Users should be able to just install plug-ins (alongside the default tensorflow package) and everything works.
@yisitu: C++ ABI boundary is too messy. We should focus on the current C API design, or the alternative design which adds C++ subclasses with C API gateways.
@jbaiocchi: Sounds good.

@jbaiocchi: The `PluginInterfaceFactory` stores a vector of interfaces. Why is one interface not sufficient?
@jbaiocchi: Who is responsible for testing if I want my own ops?

@aselle: Does the mechanism allow multiple profilers?
@jbaiocchi: The expectation is that there is one profiler per device.
@aselle: Having one profiler per .so file is a limitation. However, a workaround is to have two .so files and `dlsym` the entry points. 
@yisitu: I think that can work.
@aselle: Metaprofiler.

@jbaiocchi: Having two levels of plug-ins within TF doesn’t convince me.
@yisitu: Do we want to own the ABI?
@jbaiocchi: The question is whether the data can be interpreted correctly by all profiler tools.
@aselle: How does TF Profiler work right now? Through Proto?
@jbaiocchi: Proto -> MetaProto that can be interpreted by tools
* Look into properties of events by names whether they are meaningful to the tools.
* Conform to XSpace schema.
* @yisitu: Can be thought of as a nested map in proto where you can attach metadata.

@aselle: Binary protocol is well-defined.

@jbaiocchi: Is there a fundamental reason for having multiple interfaces? Two-level registration worries me.
* Plug-in interface is not a `ProfilerInterface`. Is there a reason it shouldn’t be?
* Should the registration go to the main registry instead of the pluggable profiler registry?
* Should just go through necessary interfaces, not all the available ones.

@aselle: Flatter hierarchy would be good, although the 2-level hierarchy shouldn’t introduce any extra limitations either.
@jbaiocchi: I’d rather have a flat hierarchy.
@aselle: Can use different constructor strings.
@jbaiocchi: Reuse what we already have.
* Factory that knows how to instantiate different .so files.
* Add an enum to the lease.

@aselle: Do we have a way to initialize different things right now? The current profiler factory is not like that.
@yiqianglee: A device wouldn’t know between `profiler1` vs `profiler2`.
@aselle: That’s a limitation of static registration. Can do name mapping with `user1`, `user2`, ...

@aselle: So pluggable profiler can be a subclass of `ProfilerInterface`
* Get rid of the for loop.
* Single DSL.
* Need some redesign work to fix [ProfileOptions](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/profiler/profiler_options.proto;l=6;drc=a838a5f5a14c4858eeed01e883e12c2f92aeb58e). 
* Probably using strings.

@jianhuili: Right now PluggableDevices use two strings, `device_type` and `device_alias`. (Design/naming subject to change.)
@aselle: Platform name?
@reedwm: Physical device name?

@aselle: `XPlane` protobuf semantics need to be documented more thoroughly.
@yiqiangli: We require plug-ins to store the exact copy of the `XPlane` protobuf.

@aselle: For collection calls, there are two function pointers. Can we separate them into 4 functions?
@jbaiocchi: We actually don’t want the `RunMetadata` one now.
@yisitu: We made it one function to minimize the surface area.
@aselle: It’s still 2 functions in one: give me the size, I’ll give you the buffer.

@aselle: What is the difference between using `sizeof` and `TF_OFFSET_OF_END` for struct size?
@yisitu: `TF_OFFSET_OF_END` gives the size of the structure without padding. It can be used to track the offset of the last variable in the struct.

@aselle: There are two use cases for `ext`, one is reserved for future use, the other is for storing user data. 
@yisitu: We don’t believe it’s being used for user data.
@aselle: We should call it something more descriptive rather than `ext`

@jzhoulon: Many device event summary pattern-matches strings, e.g., in [ClassifyGpuEvent](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/core/profiler/utils/event_span.cc;l=178;drc=b690269e02b96b36fbd559e8af7bf13a47d2828a).
@jbaiocchi: Problem of XSpace not telling you what should be in there for the tools to work.
@yiqiangli: Do you plan to make it more general (not just pattern-matching)?
@jbaiocchi: Devices should expose the strings and we shouldn’t use hard coding.

@yisitu: Someone should write up a pseudocode for the new design, then we can have another design review meeting.

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,429 @@
...
+typedef struct TF_ProfilerRegistrationParams {
+  size_t struct_size;
+  void* ext; // reserved for future use
+
+  // TensorFlow Profiler C API version.
+  int32_t major_version;
+  int32_t minor_version;
+  int32_t patch_version;
+
+  [in/out] Memory owned by core but attributes within are populated by the plugin.
```
jbaiocchi, line 126: Needs to be a comment: //

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,429 @@
...
+  TP_Profiler* profiler;
+  [in/out] Memory owned by core but attributes within are populated by the plugin.
```
jbaiocchi, line 128: Needs to be a comment: //

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,429 @@
...
+      GPU = 2;
+      TPU = 3;
+      PLUGGABLE_DEVICE = 4;
+    }
+      // Device type to profile/trace: (version >= 1)
+    // DeviceType::UNSPECIFIED: All registered device profiler will be enabled.
+    // DeviceType::CPU: only CPU will be profiled.
+    // DeviceType::GPU: only CPU/GPU will be profiled.
+    // DeviceType::TPU: only CPU/TPU will be profiled.
+    // DeviceType::PLUGGABLE_DEVICE: all pluggable devices with profilers enabled will be profiled.
```
jbaiocchi, line 177: only CPU and pluggable devices with profilers enabled will be profiled

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,429 @@
...
+  			   TP_ProfilerFns profiler_fns,
+  			   void (*destroy_profiler_fns)(TP_ProfilerFns*))
+    	: profiler_(std::move(profiler)),
+            destroy_profiler_(destroy_profiler),
+            profiler_fns_(std::move(profiler_fns)),
+            destroy_profiler_fns_(destroy_profiler_fns) {}
+  
+    ~PluggableProfilerFactory() {
+      destroy_profiler_(&profiler_);
+      destroy_profiler_fns_(&profiler_fns_);
```
jbaiocchi, line 265: Needs closing brace: }

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,429 @@
...
+  Status InitPluginProfiler(TFInitProfilerFn init_fn) {
+    TF_ProfilerRegistrationParams params{TF_PROFILER_REGISTRATION_PARAMS_STRUCT_SIZE};
+    TP_Profiler profiler{TP_PROFILER_STRUCT_SIZE};
+    TP_ProfilerFns profiler_fns{TP_PROFILER_FNS_STRUCT_SIZE};
+    params.major_version = TP_MAJOR;
+    params.minor_version = TP_MINOR;
+    params.patch_version = TP_PATCH;
+    params.profiler = &profiler;
+    params.profiler_fns = &profiler_fns;
+    OwnedTFStatus c_status(TF_NewStatus());
```
jbaiocchi, line 291: std::unique_ptr<TF_Status, TFStatusDeleter> c_status(TF_NewStatus());

for consistency with the rest of the code

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,429 @@
...
+    TF_RETURN_IF_ERROR(tensorflow::StatusFromTF_Status(c_status.get()));
+    TF_RETURN_IF_ERROR(ValidateTPProfilerRegistrationParams(params));
+    TF_RETURN_IF_ERROR(ValidateTPProfiler(profiler));
+    TF_RETURN_IF_ERROR(ValidateTPProfilerFns(profiler_fns));
+  
+    PluggableProfilerFactory factory(std::move(profiler),
+  				   params.destroy_profiler,
+  				   std::move(profiler_fns),
+  				   params.destroy_profiler_fns);
+    )
```
jbaiocchi, line 302: Orphan closing parenthesis, remove.

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,429 @@
...
+  
+    tensorflow::profiler::ProfilerFactory::RegisterProfilerFactory(
+      [factory = std::move(factory)](const ProfileOptions& options) {
+        factory.CreatePluggableProfiler(options);
```
jbaiocchi, line 306: missing 'return'

jbaiocchi (changes requested):
Please fix typos.

rfcs/20210513-pluggable-profiler-for-tensorflow.md
jzhoulon: Done

rfcs/20210513-pluggable-profiler-for-tensorflow.md
jzhoulon: Done

rfcs/20210513-pluggable-profiler-for-tensorflow.md
jzhoulon: Done

rfcs/20210513-pluggable-profiler-for-tensorflow.md
jzhoulon: Done

rfcs/20210513-pluggable-profiler-for-tensorflow.md
jzhoulon: Done

rfcs/20210513-pluggable-profiler-for-tensorflow.md
jzhoulon: Done

rfcs/20210513-pluggable-profiler-for-tensorflow.md
jzhoulon: Done

penpornk:
The RFC has been updated. There will be a second design review meeting. Here are the call details:

Time: Thursday, July 22nd, 2021 from 3-4pm PT

Meeting participants are expected to read the updated RFC ahead of time as the meeting will focus on the remaining issues/questions.

cc: @kulinseth @wchao1115 @reedwm

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,429 @@
...
+4. Core Tensorflow will create a `PluggableProfiler` during [ProfilerSession](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/profiler/lib/profiler_session.cc#L109) setup.
+
+#### Protobuf class
+Profiler uses `XSpace` to store the performance data collected by backends. With this type of data structure, TensorFlow's profiler tools can generate various views of performance, such as timeline, memory consumption, performance of every TensorFlow op and set of summaries. `XSpace` is C++ object generated by protobuf toolchain with a predefined structure in [xplane.proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/profiler/protobuf/xplane.proto#L9).
+On the plugin side, plugin will collect performance data with its own profiler infrastructures. When TensorFlow invokes `CollectData()`, the plugin transforms the performance data to `XSpace` and then serializes `XSpace` to the buffer provided by TensorFlow. To successfully serialize the object, plugin writers should keep a copy of `xspace.proto`, and make it exactly the same as that in the TensorFlow side.
```
yisitu, line 79:
```suggestion
On the plugin side, plugin will collect performance data with its own profiler infrastructures. When TensorFlow invokes `CollectData()`, the plugin transforms the performance data to `XSpace` and then serializes `XSpace` to the buffer provided by TensorFlow. To successfully serialize the object, plugin writers should keep a copy of `xspace.proto`, and make it exactly the same as that on the TensorFlow side.
```

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,429 @@
...
+* If the profiled program does not use the device, the profiler produces no output on CollectData.
+e.g., returns 0 as the required buffer size for the serialized XSpace.
+* Restartability: It must be possible to Start/Stop the profiler many times. (Since a single TP_Profiler object is created per library).
+* No (perceivable) overhead: Profiling should not change the observed performance of the profiled job.
+e.g. the step time of a TF training job should be similar with profiling on or off.
+* No OOM: Profiling should not require significant additional memory resources.
+Also, profiling should not cause the process to be killed due to running out-of-memory (OOM).
+* No leaks: any resources (memory) acquired for handling a profiling request should be released by the end of the request.
+Repeated profiling requests should not increase resource utilization over time.
+* No memory corruption: profiling should not corrupt memory due to dangling pointers.
```
yisitu, line 390:
```suggestion
* No memory corruption: profiling should not corrupt memory. e.g. due to dangling pointers.
```

yisitu (changes requested):
Minor typos + formatting + clearing entire profiler registration struct.

rfcs/20210513-pluggable-profiler-for-tensorflow.md
```diff
@@ -0,0 +1,429 @@
...
+      : profiler_fns_(profiler_fns), profiler_(profiler) {}
+    
+    TP_ProfilerFns* profiler_fns_;
+    TP_Profiler* profiler_;
+  }
+  ```
+
+* **PluggableProfiler Initialization and Registration**
+
+  Core TensorFlow will load `TF_InitProfiler` from plugin's dynamic library installed under "…python_dir.../site-packages/tensorflow-plugins" and pass the address of `TF_InitProfiler` symbol to `InitPluginProfiler` to do initialization and registration. TensorFlow retrieves `TF_ProfilerRegistrationParams` from the plugin and does the compatibility checks. If passed, TensorFlow will register the PluggableProfiler creation function to the [ProfilerFactroy](https://github.com/tensorflow/tensorflow/blob/aa855d21ac89e6649ec782ef4efd4e126b37d79d/tensorflow/core/profiler/lib/profiler_factory.cc#L39).
```
yisitu, line 249:
```suggestion
  Core TensorFlow will load `TF_InitProfiler` from plugin's dynamic library installed under "…python_dir.../site-packages/tensorflow-plugins" and pass the address of `TF_InitProfiler` symbol to `InitPluginProfiler` to do initialization and registration. TensorFlow retrieves `TF_ProfilerRegistrationParams` from the plugin and does the compatibility checks. If passed, TensorFlow will register the PluggableProfiler creation function to the [ProfilerFactory](https://github.com/tensorflow/tensorflow/blob/aa855d21ac89e6649ec782ef4efd4e126b37d79d/tensorflow/core/profiler/lib/profiler_factory.cc#L39).
```

yisitu (changes requested):
Fixed typo.

rfcs/20210513-pluggable-profiler-for-tensorflow.md
reza-amd: Hi @yisitu, I am working closely with Deven in AMD. Please also CC me for further discussions for the TF profiler. Thanks.

yisitu: Sure!

yisitu (approved):
As spoken in the meeting, RFC looks good to me.

penpornk:
## 2nd Design Review Meeting Notes

@jzhoulon: Changes from last time:
* `PluggableProfiler` is now a subclass of `tensorflow::profiler::ProfilerInterface`.
* Only `XSpace` will be passed around between TensorFlow and plug-in. (No `RunMetadata` anymore.)

### Start()/Stop() Behaviors
@aselle: Can `Stop()` be called multiple times without a `Start()`?
@jbaiocchi: TensorFlow only has 1 global client to ensure that `Start()/Stop()` are called exactly once per instance.
@yisitu: After a `Start()` is called, there could be a hardware lock while the profiling session is active. If `Start()` is called twice, the plug-in needs to fail elegantly.
@jbaiocchi: From TF’s perspective, `Start()/Stop()` will be called only once. 
Plug-ins can expect TF to be well-behaved. However, profiler libraries themselves should be robust, in case it’s not plugged into TF but something else that doesn’t have this guarantee. 
@aselle: Let’s add “TF will call 1 `Start()` followed by 1 `Stop()`” to the RFC.
@yisitu: Let’s also add that plug-ins should fail when getting 2 `Start()`s?
@jbaiocchi: Some hardware might support multiplexing, e.g., multiple profiler sessions at the same time. 
The plug-ins can just return a status warning that a `Start()` has already been called instead.
@yisitu: Agreed.
@jbaiocchi: We should document the contract from the TF side, e.g., what TF is promising to the libraries.

@reza-amd: Does `Stop()` need to be blocking?
@jzhoulon: `Stop()` is non-blocking.
@jbaiocchi: 
* We have two functions, `Stop()` and `CollectData()`, exactly for this. – We should also document this.
    * `Stop()` interacts with the hardware.
    * `CollectData()` converts data from native profiler format to TF format. It’s not allowed to interact with the hardware.
* In an RPC mode where we send profiling requests to workers, the control thread of the profiler can be independent of threads that are doing the work. In that case, the control thread can be blocking.
* Anyway, instantiating a new profiler session while converting the existing profiling data should work.

@yisitu: API can fit different types of use cases, e.g., `Stop()` can either flush or not flush the buffer. Interaction with hardware is intentionally not mentioned in the RFC such that the API is not too restrictive.
Maybe we need an `IsReady()` call?
@jbaiocchi: That can be done behind the scenes.
@yisitu: We don’t need to add anything right now. We can extend it afterwards if needed.

### Data Structure
@Jianhui-Li: One last unresolved topic from the last review was about the data structure. Just wanted to confirm that it’s not part of this RFC.
@yisitu / @jbaiocchi: Yes, it should be a follow-up RFC – define standard ML metrics / construct data collection.
@Jianhui-Li: Will clarify this in the RFC.
@jbaiocchi: Right now, `TraceViewer` is the only tool that we promise will work.
Plug-ins own all the contents you put in `XPlane`.
@Jianhui-Li: Do you have a plan to sort that out?
@jbaiocchi: When they did it for AMD GPU, they looked at the API for Nvidia and created an equivalent one. We understand that this is not necessarily the case for every device.
@yisitu: We will look at what other tools users are interested in apart from `TraceViewer`. `TraceViewer` is a good first step. 
@Jianhui-Li: Sounds good.

### Action Items
@jbaiocchi: Two action items:
* Outline the behaviors that plug-ins can expect from TensorFlow.
* Document what people need to do to ensure TraceViewer will work.

### Code Location and Testing
@reza-amd: Where would the code be in?
@yisitu: The interface code will be in the TensorFlow repository. Plug-in code would be in your own repositories.
@reza-amd: Who will run the unit tests?
@jbaiocchi: We won’t have all the hardware to test for all plug-ins, so the plug-in developers will have to run the tests themselves. We can do a simple mock test.
@reza-amd: You mentioned multiple-pipeline profiling sessions. What are the scenarios we should test for? Is there a test plan?
@jbaiocchi: We can create a mock library with mock tests. You can use this as an example of what we expect the real plug-ins to conform to.
@yisitu: You can also contribute tests to the TensorFlow repository.

@penpornk / @yisitu: What device are you targeting, @reza-amd?
@reza-amd: AMD GPUs. We are looking to improve the profiler support.

Pasio47DBG:
26cb891

Pasio47DBG:
RFC:Accepted cla:yes

PR #391, opened 2021-06-18
Remove cpp style guide

aaudiber:
TF now (as of #60fcc4cb35748b9b201e091e0c4edd9709a17f0a) supports StatusOr, in line with the Google C++ style guide. The exception no longer needs to be documented.


aaudiber:
@ematejska PTAL

PR #395, opened 2021-07-31
A "Named-Dimensions" TypeScript Library for TensorFlowJS (initial draft)

iislucas:
This RFC will be open for comment until Friday, August 30th, 2021.

cc @iislucas @pyu10055

| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)|
| **Author(s)** | Lucas Dixon (ldixon@google.com) |
| **Sponsor**   | Ping Yu (piyu@google.com) (piyu@tensorflow.org) |
| **Updated**   |  2021-07-31     |

## Objective

Support the naming of dimensions in TensorFlow JS to provide an abstraction that makes it easier for developers and researchers to write and read ML algorithms (and support some automated optimisation):

 * Make it harder to make subtle mistakes (have type-checking get intelligible type-error messages).
 * Leveraging TypeScript IDE support to get better auto-completion.
 * Make it easier to support writing an algorithm and then generalise it to support batches.
 * Have code more closely resemble the pseudocode and algorithms in papers.
 * Allow machine optimization of the ordering of indexes, transpositions and permutations.

### DIFF
--- a/rfcs/20210731-tfjs-named-tensors.md
+++ b/rfcs/20210731-tfjs-named-tensors.md
@@ -1,0 +1,368 @@
+# A "Named-Dimensions" TypeScript Library for TensorFlowJS
+
+| Status        | Proposed       |
+:-------------- |:---------------------------------------------------- |
+| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)|
+| **Author(s)** | Lucas Dixon (ldixon@google.com) |
+| **Sponsor**   | Ping Yu (piyu@google.com)       |
+| **Updated**   | 2021-07-31                      |
+
+## Objective
+
+Support the naming of dimensions in TensorFlow JS to provide an abstraction that makes it easier for developers and researchers to write and read ML algorithms (and support some automated optimisation):
+
+ * Make it harder to make subtle mistakes (have type-checking get intelligible type-error messages).
+ * Leveraging TypeScript IDE support to get better auto-completion.
+ * Make it easier to support writing an algorithm and then generalise it to support batches.
+ * Have code more closely resemble the pseudocode and algorithms in papers.
+ * Allow machine optimization of the ordering of indexes, transpositions and permutations.
+
+## Motivation
+
+ML programming is currently much like assembly programming: you have to remember what each index in a tensor means, and refer to dimensions by index numbers, and when they change due to the operations you perform, you have to remember the new index position. Higher level abstractions could make this better and various ideas have been proposed around naming the dimensions in a Tensor e.g. http://nlp.seas.harvard.edu/NamedTensor
+
+The impact is likely largest for people who want to write ML algorithms, and for people using TFJS for education, e.g. to illustrate and visualize ML algorithms. However, it may also improve the experience for others using TFJS by providing better tool support, better auto-completion, and improved  optimization for evaluation.
+
+Here's an example TFJS implementation of single attention-head:
+
+```ts
+// keys.shape == [seqLen, inputRep]
+const keys = tf.matMul(xs, keyM.read())
+
+// queries.shape == [seqLen, inputRep]
+const queries = tf.matMul(xs, queryM.read())
+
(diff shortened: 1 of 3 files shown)


rfcs/20210731-tfjs-named-tensors.md
```diff
@@ -0,0 +1,557 @@
...
+
+The proposed Named-Dimension library allows the same thing to be written as:
+
+```ts
+const queryM: GTensor<'inputRep', 'kqRep'> = ...
+const keyM: GTensor<'inputRep', 'kqRep'> = ...
+const valueM: GTensor<'inputRep', 'kqRep'> = ...
+const xs: GTensor<'inputRep', 'seqLen'> = ...
+
+const inputKeys = xs.inputRep.dot(keyM.inputRep).seqLen.rename('keySeqLen');
```
robertlugg, line 92: In this example you first name it 'keys' (line 76) but here name it 'inputKeys'.  I would suggest keep everything consistent between the two (before and after) code chunks, except for what the RFC differences consist of.  Otherwise you are biasing the "readability" of the two code chunks.

rfcs/20210731-tfjs-named-tensors.md
```diff
@@ -0,0 +1,557 @@
...
+// Note: that rename returns a GTensor, and typechecking happens at edit time,
+// so 'xs.inputRep.dot(keyM.kqRep)' would give a red underline and nice error :)
+const inputQueries = xs.inputRep.dot(queryM.inputRep);
+const attention = inputKeys.kqRep.dot(inputQueries.kqRep);
```
robertlugg, line 96: Perhaps I don't understand the code, but 'attention' here looks to be much different logic than in the previous two (before) examples.  It should be as close to the same in every way except for the RFC changes.

rfcs/20210731-tfjs-named-tensors.md
```diff
@@ -0,0 +1,557 @@
...
+g1.dim.tokenRep.dot(g2.dim.foo)
+             // Type Error: "tokenRep" is different from "foo".
+```
+
+Dimensions can be renamed also to provide a new dimension object with the
+correct name, e.g. if `g2` didn't have an `tokenRep` dimension, but we wanted to
+dot product with the `foo` dimension, we could do:
+
+```ts
+g1.dim.tokenRep.dot(g2.dim.foo.rename('tokenRep'));
```
robertlugg, line 167: This seems clunky and confusing.  Maybe you all have thought this through more and I understand what you are trying to do, but this line seems to be adding complexity while other aspects, I agree, are improving things.  I just don't see how this is improving readability or reducing complexity.  I realize that's subjective.  Perhaps a more concrete example (don't use 'foo' as a dimension).

iislucas:
I'm having a strangely hard time finding these comments in the GitHub UI,
maybe github is having issues, but thanks for the feedback!  I clarified a
little why I suggest using explicit renaming, and improved the example as
per your suggestion :)

iislucas:
Thanks! I fixed the issue that the attention computation wasn't the same -
I removed the normalization as it doesn't really make the example clearer.

iislucas:
Good idea, fixed the names to be the same in both examples. I thought I had
done that already, so easy to become blind to your own RFC! :P

pyu10055:
@ematejska The design review is concluded with no unsolved questions, and here is the notes.
Design review notes:
Participants:
Ping, Lucas, Masoud, Mark, Mehdi

Lucas Dixon gives an overview of the problem.
The design is to track the name of the dimensions instead of the size of the dimensions. The penalty to pay is that it will require renaming when you need to match two dimensions with different names.
Questions:
Will dimension names collide with gTensor methods?
Answer: the dimension names are exposed only on the dimension object not the gTensor.

Dimension type can be string/symbol/number, should it just be string type to reduce user error?
Answer: agreed

What is the performance implication if tfjs layers use this API? 
Still under investigation.
Is GTensor a new implementation or a wrap of the existing tensor object?
It is a wrap of the existing tensor object.

PR #398, opened 2021-09-09
RFC: Mechanism to indicate that tflite model is structurally pruned.

wwwind:
<html><body>
<h1>Mechanism to indicate that tflite model is structurally pruned.</h1>

Status | Draft
-- | --
RFC # | 398
Author(s) | Elena Zhelezina (elena.zhelezina@arm.com)
Sponsor | David Rim (davidrim@google.com)

<h2>Objective</h2>
<p>When we have the structurally pruned tflite model, then the only way for hardware to identify this to get benefits is to scan all weights and do checks. This increases the inference time and should be done every time, when the model is loaded. The goal of this RFC is to add special flag to tflite file to mark such models.</p>
</body>
</html>

### DIFF
--- a/rfcs/20210908-structural-sparsity-flag.md
+++ b/rfcs/20210908-structural-sparsity-flag.md
@@ -38,9 +38,9 @@ table StructuralSparsityParameters {
 During the model conversion, we need to indicate whether tensors should be checked.
 For this purpose we introduce a new flag `STRUCTURAL_SPARSITY` to the TensorFlow Lite convertor.
 
-```cpp
+```python
 converter = tf.lite.TFLiteConverter.from_keras_model(model)
-converter.optimizations = {tf.lite.Optimize.STRUCTURAL_SPARSITY}
+converter.optimizations = [tf.lite.Optimize.STRUCTURAL_SPARSITY]
 tflite_model = converter.convert()
 ```
 
--- a/rfcs/20210908-structural-sparsity-flag.md
+++ b/rfcs/20210908-structural-sparsity-flag.md
@@ -1,0 +1,104 @@
+# Title of RFC
+
+| Status        | (Proposed)       |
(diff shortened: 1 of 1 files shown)


rino20:
It saves model loading time but increases model conversion time. 
I have two questions about this proposal:
1. How to map between the flag and tensors(weight) 
2. What additional check will be done to the indicated tensors?

rfcs/20210908-structural-sparsity-flag.md
```diff
@@ -0,0 +1,104 @@
...
+## Design Proposal
+
+We propose to add an optional structure `StructuralSparsityParameters` to the structure [Tensor] (https://github.com/tensorflow/tensorflow/blame/master/tensorflow/lite/schema/schema.fbs#L195) that provides the parameters of structural pruning. If it is present, then the tensor is structurally pruned.
+
+```
+table StructuralSparsityParameters {
+  // For a tensor that is structurally pruned we save the type of the pruning.
+  // (m, n) means that in a block of n elements, m elements are zeros.
+  // If this field is NULL, then the tensor is not sparse.
+  sparsity_type_m_by_n:[int];
```
daverim, line 34: If I'm reading this correctly this is true/false? MxN are not stored in the model, but set at initialization?

Is the current flag for sparsity not enough?

ematejska:
Has this been or is this ready for community review?

daverim (commented):
I think we need to review this further with checks to make sure this is universally useful, otherwise, this can be handled by delegates that support MxN sparsity (by using a persistent variable, for example) I'm re-assigning this for further review

daverim:
@rino20, can you take over as sponsor?

daverim:
> 

Sorry for all the comments, no it's not ready for community review yet.

PR #424, opened 2022-08-25
RFC: FunctionType for tf.function

faizan-m:


faizan-m:
@ematejska do you think it is ok to merge this PR now?

PR #434, opened 2022-10-25
RFC: API to synchronize devices

reedwm:
| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)|
| **Author(s)** | Reed Wanderman-Milne (reedwm@google.com), Jonathan Dekhtiar (jdekhtiar@nvidia.com) |
| **Sponsor**   | Rohan Jain (rohanj@google.com)                       |
| **Updated**   | 2022-10-25                                           |

## Objective

This document proposes a simple API to synchronize TensorFlow devices: `tf.sync_devices()`. This is important in accurately measuring execution time in TensorFlow GPU benchmarks, especially in microbenchmarks.

/CC @DEKHTIARJonathan @rohan100jain

### DIFF
--- a/rfcs/20221025-sync-devices.md
+++ b/rfcs/20221025-sync-devices.md
@@ -4,8 +4,8 @@
 :-------------- |:---------------------------------------------------- |
 | **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)|
 | **Author(s)** | Reed Wanderman-Milne (reedwm@google.com), Jonathan Dekhtiar (jdekhtiar@nvidia.com) |
-| **Sponsor**   |                                                      |
-| **Updated**   | 2022-10-24                                           |
+| **Sponsor**   | Rohan Jain (rohanj@google.com)                       |
+| **Updated**   | 2022-10-25                                           |
 
 ## Objective
 
--- a/rfcs/20221024-sync-devices.md
+++ b/rfcs/20221024-sync-devices.md
@@ -1,0 +1,91 @@
+# API to synchronize devices
+
+| Status        | Proposed                                             |
+:-------------- |:---------------------------------------------------- |
+| **RFC #**     | [NNN](https://github.com/tensorflow/community/pull/NNN) (update when you have community PR #)|
+| **Author(s)** | Reed Wanderman-Milne (reedwm@google.com), Jonathan Dekhtiar (jdekhtiar@nvidia.com) |
+| **Sponsor**   |                                                      |
+| **Updated**   | 2022-10-24                                           |
+
+## Objective
+
+This document proposes a simple API to synchronize TensorFlow devices: `tf.sync_devices()`. This is important in accurately measuring execution time in TensorFlow GPU benchmarks, especially in microbenchmarks.
+
+## Motivation
+
+TensorFlow runs GPU ops asynchronously. This means when a user calls an op, the op will return to the user before the GPU actually finishes computing the output. Unfortunately, this causes issues when measuring performance during a benchmark. For example, the following program tries to measure how long a matrix multiplication takes, but significantly underestimates the time taken because the matmul is still running asynchronously on the GPU even after `tf.linalg.matmul` returns.
+
+```python
+start = time.time()
+y = tf.linalg.matmul(x, x)
+print(f'Time taken: {time.time() - start}')
+```
+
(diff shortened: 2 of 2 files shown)


rfcs/20221025-sync-devices.md
```diff
@@ -0,0 +1,91 @@
...
+| Status        | Proposed                                             |
+:-------------- |:---------------------------------------------------- |
+| **RFC #**     | [434](https://github.com/tensorflow/community/pull/434) |
+| **Author(s)** | Reed Wanderman-Milne (reedwm@google.com), Jonathan Dekhtiar (jdekhtiar@nvidia.com) |
+| **Sponsor**   | Rohan Jain (rohanj@google.com)                       |
+| **Updated**   | 2022-10-25                                           |
+
+## Objective
+
+This document proposes a simple API to synchronize TensorFlow devices: `tf.sync_devices()`. This is important in accurately measuring execution time in TensorFlow GPU benchmarks, especially in microbenchmarks.
```
sun51, line 12: Can we put into this namespace. tf.config.experimental.sync_devices()?

rfcs/20221025-sync-devices.md
```diff
@@ -0,0 +1,91 @@
...
+
+TensorFlow runs GPU ops asynchronously. This means when a user calls an op, the op will return to the user before the GPU actually finishes computing the output. Unfortunately, this causes issues when measuring performance during a benchmark. For example, the following program tries to measure how long a matrix multiplication takes, but significantly underestimates the time taken because the matmul is still running asynchronously on the GPU even after `tf.linalg.matmul` returns.
+
+```python
+start = time.time()
+y = tf.linalg.matmul(x, x)
+print(f'Time taken: {time.time() - start}')
+```
+
+This can be fixed by calling `y.numpy()` which forces the Python thread to wait until the matmul finishes, but this also adds a device-to-host transfer. The benchmark only wants to measure the matmul time, not the device transfer time.
```
sun51, line 24: what are you trying to measure exactly?  because when we call tf.matmul(), even with sync, it also includes all the Python overhead, C++ launch overhead, and then GPU execution time.

rfcs/20221025-sync-devices.md
DEKHTIARJonathan: @sun51 It is not a config call. It could be call at every iterations:

```python
for step in range(NUM_STEPS):
   data = ds_iter.next()
   start_t = time.time()
   rslt = model(data)
   tf.sync_devices()
   print(f"Time: {time.time() - start_t}")
```

I don't think `config` is the appropriate namespace for such an API.

rfcs/20221025-sync-devices.md
DEKHTIARJonathan: The idea is to measure step time (including all the potential overheads) fron start to finish.
Having the most accurate picture from start to finish of a single "step" or "compute unit". Without having to force a `.numpy()` that would memcpyDtoH.

See: https://github.com/tensorflow/community/blob/1fbc2877e154973cbc37d0405e94cb18852e67cd/rfcs/20221025-sync-devices.md#motivation

rfcs/20221025-sync-devices.md
sun51: It is actually changing the setting of runtime.  Meanwhile,  I am still not quite clear about the semantics of this API.  In the above case, model() will include many ops,  it is called  before the tf.sync_device() is even called. How does this work?   Once you sync device, can you set to normal config(async) later?

DEKHTIARJonathan: I think you there is a misunderstanding on how the API works.
> It is actually changing the configuration of runtime.

No, the API introduces a blocking call that awaits for the OPs scheduled on every devices to be cleared. There is no "configuration" being updated or modified. It's just an API call that awaits for the GPUs to be done. Hence this has nothing to do in the `config` namespace IMHO.

sun51: chatted with reed, put into tf.test.XXX namespace is something we both think reasonable.

reedwm: Sorry for the delay, I edited the RFC to put under tf.test.sync_devices.

sun51:
LGTM, thanks

sun51 (approved):

DEKHTIARJonathan:
@reedwm can you please link the commit that adds the feature or the PR/CL here ?

Thanks

PR #436, opened 2022-11-17
Stochastic rounding

elfieguo:
[RFC] Enable Stochastic Rounding in Tensorflow


rfcs/20221116-enable-stochastic-rounding.md
```diff
@@ -0,0 +1,335 @@
...
+
+```
+tf.stochastic_cast(input, dtype, random_seed=None, random_algo=’Philox’)
+```
+
+<table>
+ <tr><td><strong>Args</strong></td><td><strong>Definition</strong></td></tr>                                      
+ <tr><td>input</td><td>The original tensor to be casted</td></tr>                              
+ <tr><td>dtype</td><td>Desired type after casting</td></tr>       
+ <tr><td>random_seed</td><td>Random seed if users need to control the randomness. If None, the system will generate a seed.</td></tr>
```
reedwm, line 134: If the user has a set a seed with [`tf.random.set_seed`](https://www.tensorflow.org/api_docs/python/tf/random/set_seed), I think we should use that if `random_seed` is None. Typically ops that take a `seed` argument defaulting to None, like [`tf.random.normal`](https://www.tensorflow.org/api_docs/python/tf/random/normal), use the global seed from `tf.random.set_seed` if no seed is passed to the function.

Not all random ops use the global seed. There are three classes of random ops:

* Legacy random ops like `tf.random.normal`. These ops takes a seed defaulting to None. `tf.stochastic_cast` also follows this pattern. Legacy random ops will generate deterministic random numbers if either the seed argument or the global seed is set.
* [`tf.random.Generator`][rand-gen]: This doesn't use the global seed, but instead is created with the `from_seed` method, which requires passing in a seed, or the `from_non_deterministic_state` method, which nondeterministically creates a seed
* Stateless random ops like [`tf.random.stateless_uniform`](https://www.tensorflow.org/api_docs/python/tf/random/stateless_uniform). These ops require the user to pass a seed, and deterministically return the same values each time given the same inputs and seed.

Since `tf.stochastic_cast` is most similar to the legacy random ops, it's best to use the global seed from `tf.random.set_seed` to be consistent with them. The user expects that if they call `tf.random.set_seed`, then non-stateless non-generator random ops will use the global seed if no seed is passed.

This is important for determinism, which is enabled by calling [`tf.config.experimental.enable_op_determinism`](https://www.tensorflow.org/api_docs/python/tf/config/experimental/enable_op_determinism) and setting the global seed. Once the user enables determinism, they expect all ops to be deterministic.

CC @wangpengmit

[rand-gen]: https://www.tensorflow.org/api_docs/python/tf/random/Generator

rfcs/20221116-enable-stochastic-rounding.md
```diff
@@ -0,0 +1,335 @@
...
+low-precision fixed-point numbers or lower-bit floats to reduce the
+computational cost of neural network training or inference.
+
+For example, below is a code snippet that represents a part of process to
+quantize a tensor from floating-points to int8, which is then used for matrix
+multiplications:
+
+```
+inputs = tf.clip_by_value(inputs, -128.0, 127.0)
+inputs = tf.math.round(inputs) inputs = tf.cast(inputs, dtype=tf.int8)
```
reedwm, line 95: This line should probably be split into two.

rfcs/20221116-enable-stochastic-rounding.md
wangpengmit: The 3 classes Reed listed are correct. I view `tf.stochastic_cast` as a new RNG op. For a new RNG op, we only recommend implementing it as the 3rd class (i.e. totally stateless). After that, the 2nd class can be (optionally) easily implemented as a thin wrapped on top of the 3rd class ([example](https://github.com/tensorflow/tensorflow/blob/20b6531762ca1949d25ecf858a0bb3de906df896/tensorflow/python/ops/stateful_random_ops.py#L719)). The 1st class (potentially along with `tf.random.set_seed` and `tf.compat.v1.get_seed`) is going to be deprecated because those ops store the RNG state in the op kernel, which interacts badly with other TF facilities (e.g. `tf.function`).

A detail is that to implement the 3rd class, we currently recommend the new op taking in a `key` and a `counter` ([example](https://github.com/tensorflow/tensorflow/blob/20b6531762ca1949d25ecf858a0bb3de906df896/tensorflow/core/ops/stateless_random_ops_v2.cc#L90)), instead of a `seed` ([example](https://github.com/tensorflow/tensorflow/blob/20b6531762ca1949d25ecf858a0bb3de906df896/tensorflow/core/ops/stateless_random_ops.cc#L57)), to make supporting the 2nd class easier. The Python API can still take in a `seed` ([example](https://github.com/tensorflow/tensorflow/blob/20b6531762ca1949d25ecf858a0bb3de906df896/tensorflow/python/ops/stateless_random_ops.py#L503)).

wangpengmit (changes requested):

rfcs/20221116-enable-stochastic-rounding.md
```diff
@@ -0,0 +1,335 @@
...
+experimental random number generation algorithms to be passed in. Under such
+circumstances, encapsulation might complicate future developments. However, up
+till now we haven’t received any user requests regarding random number
+generation algorithms or custom generators.
+
+### Alternatives Considered
+
+Similar to the proposal, but this option encapsulates random number generation
+inside of the API call and asks users to provide random seed and random number
+generation algorithms.
```
poulsbo, line 165: This sounds like a description of the proposal above, not the alternative below. Should it not say something like "this option requires the caller to generate a random number in the range [0, 1)".

poulsbo:
Does the proposed API allow the caller to pass in a single seed that governs the RNG for multiple operations?

For example, does this work as the user probably expects?

```
cast_a = tf.stochastic_cast(input_a, dtype=tf.bf8, random_seed=my_seed)
cast_b = tf.stochastic_cast(input_b, dtype=tf.bf8)
```

Will `my_seed` be used for all random number generation across both calls to `tf.stochastic_cast`?

wangpengmit:
> Does the proposed API allow the caller to pass in a single seed that governs the RNG for multiple operations?

I recommend making `random_seed` (I recommend renaming it to `seed`) not optional, similar to e.g. `tf.random.stateless_uniform`. Then we won't have this ambiguity.

rfcs/20221116-enable-stochastic-rounding.md
elfieguo: Thanks both! To summarize:
We endorse following tf.stateless_uniform and tf.random.Generator's user usage and implementation over legacy tf.random.normal's, as the latter is incorporated badly in the new TF stack.

A.I.s:
1) Random seed to be required in the API level 
2) Use key + counter instead of single seed in the kernel

rfcs/20221116-enable-stochastic-rounding.md
elfieguo: Thanks for catching! revised.

rfcs/20221116-enable-stochastic-rounding.md
elfieguo: Thanks for catching! revised

elfieguo:
> Does the proposed API allow the caller to pass in a single seed that governs the RNG for multiple operations?
> 
> For example, does this work as the user probably expects?
> 
> ```
> cast_a = tf.stochastic_cast(input_a, dtype=tf.bf8, random_seed=my_seed)
> cast_b = tf.stochastic_cast(input_b, dtype=tf.bf8)
> ```
> 
> Will `my_seed` be used for all random number generation across both calls to `tf.stochastic_cast`?

As Peng suggested, flagging random seed as required to 1) disambiguates the usage; 2) keep usage consistent with tf.random.stateless_uniform

rfcs/20221116-enable-stochastic-rounding.md
```diff
@@ -0,0 +1,329 @@
...
+Another way that sticks to the current rounding API design:
+
+```
+tf.stochastic_round(input, to_precision, seed=None, alg=’Philox’)
+```
+
+<table>
+ <tr><td><strong>Args</strong></td><td><strong>Definition</strong></td></tr>
+ <tr><td>to_precision</td><td>Desired precision after rounding</td></tr>
+ <tr><td>seed</td><td>Required seed for the RNG.</td></tr>
```
sun51, line 200: it has a default value 'None' so what happens if users do not set it? what is the guideline on how to choose a seed

rfcs/20221116-enable-stochastic-rounding.md
```diff
@@ -0,0 +1,329 @@
...
+There is a disadvantage of this implementation that users need to materialize
+large tensors before calling this API. Once they are
+materialized, calling this API will require storing and passing the large tensor
+around, causing OOM issues. When facing this OOM issue, users will need to potentially break down
+the tensor in order to bypass the memory limit.
+
+Another way that sticks to the current rounding API design:
+
+```
+tf.stochastic_round(input, to_precision, seed=None, alg=’Philox’)
```
sun51, line 194: will this be implemented by a op kernel, or totally in python

elfieguo: Current plan is to implement with op kernels, unless suggested otherwise.

elfieguo:
Approved by the leads. Below are the notes from the design review: 

1. From Reed: We can consider adding a `stochastic_cast` method to `tf.random.Generator`, in addition to the top-level `tf.stochastic_cast` function. For example, in the longer term, we can add a `Generator.stochastic_cast` method. That way the user can call `generator.stochastic_cast(x, tf.float8_e5m2)` instead of `tf.stochastic_cast(x, tf.float8_e5m2, seed=generator.uniform_full_int([2]))`. For now we can just stick with `tf.stochastic_cast`. 

2. From Yu: Fixed seed might still introduce rounding bias even though the whole point of stochastic rounding is to reduce bias. This can be eliminated by varying seeds every time calling the API. 

3. From Antonio: Given a fixed seed and fixed location within the tensor, there will be a bias at that location dependent on the value of `pRNG(seed, location)`.  This can be easily seen if `pRNG(s1, i1) = 1`, for example, which will always cause the value at tensor location `l1` to be rounded up.
Without prior knowledge of the generated values, the expected bias is still zero.  And in operations that mix the values (reductions, multiplications), as long as the tensor values themselves are somewhat independent and identically distributed (or at least have similar means), the mixing of terms will again cause an expected bias of zero.

3. From Chiachen: A fixed seed might not be of a big problem as in sequential matmuls, rounding errors across operands may cancel off bias, accumulated error might be insignificant. 

4. From Peng: Should default random number algorithm to `auto_select` because Philox is slow on TPU and Threefry is slow on CPU and GPU. Also, Peng has is working in progress to add uint16 support to random number generator API. 

5. From Antonio:  In the future, if users request a custom random number generator for stochastic rounding, they might experiment their algorithms with a simplified implementation in primitive ops. If they are proven to work better than current algorithms, it's not hard to support additional algorithms on the current infra. 

6. From Reed: We should emphasize in the doc string that it's encouraged to use different seeds (potentially generated from random generators) instead of fixed ones.

theadactyl (approved):

PR #439, opened 2023-01-11
Introduce Fingerprint class and add experimental namespace.

monicadsong:
The Fingerprint class and read_fingerprint method will first be implemented as `tf.saved_model.experimental.Fingerprint` and `tf.saved_model.experimental.read_fingerprint`.


monicadsong:
Gentle ping, @theadactyl!

PR #446, opened 2023-05-16
Added DynamicEmbedding RFC

divyashreepathihalli:
Added DynamicEmbedding RFC


rfcs/20230515-DynamicEmbedding.md
```diff
@@ -0,0 +1,251 @@
...
+   ])
+   train_labels = np.array([0, 1, 2])
+   vocab = tf.constant(['a', 'b', 'c', 'd', 'e'])
+   eviction_policy = 'LFU'
+   # Define the model
+   model = keras.models.Sequential([
+       DynamicEmbedding(
+           input_dim=5,
+           output_dim=2,
+           input_length=5,
```
rhdong, line 205: Can we support the inputs with dynamic shapes?

rfcs/20230515-DynamicEmbedding.md
```diff
@@ -0,0 +1,251 @@
...
+### Extended goals
+    * Works with synchronous distributed training
+
+## Motivation
+Recommendation systems and search ranking are crucial in powering the largest
+revenue streams, such as PCTR/PCVR and video recommendation. However, as
+recommendation models have become more complicated, there are three distinct
+challenges that need to be addressed. These include 
+
+the difficulty in separating popular and less-popular items or adapting to
```
rcrowe-google, line 31: Is this line break intended?

rfcs/20230515-DynamicEmbedding.md
rhdong: Excellent, thank you for your answer! I would like to know what the `input_dim` means. From my understanding, `input_dim` should be less or equal to the `vocabulary size`, which is fixed when training going on, is it right?

rfcs/20230515-DynamicEmbedding.md
```diff
@@ -0,0 +1,251 @@
...
+    * Old vocab keys will have the same embedding vector 
+    * New vocab keys will have newly initialized embedding vector
+This updated vocabulary is used for lookup in the DynamicLookup layer
+until the callback event is triggered again after the time interval.
+
+![DynamicLookup](0230515-DynamicEmbedding/DynamicLookup.png)
+
+The image below illustrates the workflow when the parameter server
+strategy is used. PSS supports asynchronous training. Each worker
+will have a copy of the vocabulary, which will be consistent across
```
rhdong, line 122: If it means each worker will hold a full set of vocabulary that maps the `vocab` to `index`, and the real embedding vectors stored in some PSs with dense format(for example: the `tf.Variable`)?

rfcs/20230515-DynamicEmbedding.md
divyashreepathihalli: input_dim should be vocabulary size

rfcs/20230515-DynamicEmbedding.md
divyashreepathihalli: That is correct. Each worker should have a copy of the vocabulary( vocab->index mapping). The embedding variable will be split in distributed servers.

rfcs/20230515-DynamicEmbedding.md
divyashreepathihalli: no, updated.

rfcs/20230515-DynamicEmbedding.md
rhdong: Hi @divyashreepathihalli, thank you for your comment!  If we have a full set copy of the `key-index` mapping on each worker, there should be some upper limitations on the `vocabulary size`. To my best knowledge, some `vocabulary size` in some industrial scenarios can be **tens or hundreds of billions**, which causes the memory consumption on GPU/TPU to be significantly large and expensive. The practical method is storing the `key-value` with the format of abstract hashtable in a distributed way like TFRA. Hope it's helpful. Thanks!

rfcs/20230515-DynamicEmbedding.md
rhdong: > @rhdong, I would like to clarify that for the layer initialization `inp_dim` is vocabulary size(tried to keep it consistent with the [embedding layer](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Embedding#:~:text=.Embedding(-,input_dim%2C,-output_dim%2C%0A%C2%A0%20%C2%A0%20embeddings_initializer) - . The input to the layer can be of any dynamic shape.

Hi @divyashreepathihalli, thank you for your clarification. I understand now, but I believe the `input_dim` setting would limit the features encoding space. In the dynamic embedding context, `input_dim` should be `std::numeric_limits<uint64_t>::max`.  I would try to explain it in one google doc.

MoFHeka: > @rhdong, I would like to clarify that for the layer initialization `inp_dim` is vocabulary size(tried to keep it consistent with the [embedding layer](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Embedding#:~:text=.Embedding(-,input_dim%2C,-output_dim%2C%0A%C2%A0%20%C2%A0%20embeddings_initializer) - . The input to the layer can be of any dynamic shape.

I may think @divyashreepathihalli  may misunderstand the meaning of dynamic shape embedding. For example, there is a training feature input that is both large scale and sparse, such as USER_ID. If we apply the vocabulary method to USER_ID, it will only map USER_ID to the dimension of vocabulary size, which is a compression of the information dimension. Since the vocabulary size is fixed, this is still a static embedding. Dynamic embedding means that all inputs can be mapped non-conflicting through a hash function. The size of the dynamic embedding is not fixed because the USER_ID grows with the growth of the business.

divyashreepathihalli: @rhdong  @MoFHeka thank you for the clarification. I tried to read up further. If I understand correctly you are looking for a dynamic vocabulary size and a dynamic embedding matrix as well, correct? One that would keep growing?

As of now our scope of work will be limited to maintaining a fixed size vocabulary and fixed embedding size, updating the vocabulary based on inputs received by the layer and eviction policies. The embedding values will be remapped whenever the vocabulary is updated based on input patterns (most frequently seen input, TTL, etc). If the input key is not in the vocab it will be mapped to a default value, however we keep track of these keys and add it to the vocab when the updates are done in the callback(new keys are added in the vocab by kicking out old keys based on eviction policies specified)

rhdong: @divyashreepathihalli It's my pleasure. Considering the practical scenario of dynamic embedding we reached out to, the hashtable-based dynamic vocabulary size would be a fundamental requirement. I guess one of the PROs of your current design is that there is no need to modify the `tf.optimizer`; that makes sense, but in addition to the considerations we discussed above, I'm also a little worried it will introduce the data consistency issue caused by decoupling the `embedding indexing` and `embedding looking up`, especially in eviction involved. As you know, adding atomic or lock mechanisms to ID and embedding is challenging when they are operated in two separate OPs.

Mr-Nineteen:
Dynamic embedding is a very important feature for us.

When training the sorting model that supports scenarios such as search, recommendation, and advertisement, we encountered  the following problems:
1. For the feature selection of the sorting model in the e-commerce search and promotion scenario, the current industry mostly adopts the idea of large-scale discrete IDs. ID features (product IDs, user IDs, brand IDs, etc.) are large and sparse, and the native TF framework is not applicable.
2. The TensorFlow variable has a fixed size and cannot dynamically increase the ID without restarting training.

With this feature, the main reasons:
1. Support dynamic scale-out of dynamic embedding features at the TB level.

rfcs/20230515-DynamicEmbedding.md
```diff
@@ -0,0 +1,250 @@
...
+front - TorchRec with a native embedding distribution and a dynamic
+embedding solution & HugeCTR (Merlin) with a highly-performant
+embedding caching strategy. This makes it essential for TensorFlow to
+introduce a native solution to stay competitive in the market.
+
+## Design Proposal
+In this design approach, the DynamicEmbedding layer is composed of two
+layers: the DynamicLookup layer and the Embedding layer. The
+DynamicLookup layer is responsible for the following tasks:
+    * Maintaining a vocabulary table using an eviction policy that is
```
Lifann, line 78: We are currently using parameter size in volume about 1E13 bytes, and also in many other companies. Will it be very expansive to maintain vocabulary for large parameter?

rfcs/20230515-DynamicEmbedding.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Design Proposal
+In this design approach, the DynamicEmbedding layer is composed of two
+layers: the DynamicLookup layer and the Embedding layer. The
+DynamicLookup layer is responsible for the following tasks:
+    * Maintaining a vocabulary table using an eviction policy that is 
+    updated based on input pattern.
+    * Performing vocabulary lookup for the given input and returning
+    integer indexes.
+    * The index is then passed to the Embedding layer, which looks
```
Lifann, line 82: In many case, we don't know how many keys in a feature exactly, since the property of videos/images/commodity/video-games are always in change. Preset a vocab/index range may lead to waste in storage or feature conflicts.

rfcs/20230515-DynamicEmbedding.md
```diff
@@ -0,0 +1,250 @@
...
+
+## Objective
+The objective of this proposal is to introduce the DynamicEmbedding layer to
+the Keras ecosystem, providing a native solution for handling
+colossal-scale problems in recommendation systems. The proposed solution 
+facilitates automatic vocabulary building and updates, and dynamic embedding
+updates corresponding to evolving input patterns and vocabulary changes.
+### Goal
+    * Works across accelerators (GPU / TPU)
+    * Works with Parameter server strategy (asynchroous distributed training)
```
pjannaty, line 18: Typo: asynchronus

PR #453, opened 2023-07-24
RFC: Unbound SavedModel

BlaziusMaximus:
This RFC will be open for comment until Friday, August 11th, 2023.
cc @k-w-w @petrychenko

# Unbound SavedModel

| Status        | Implemented                                                                                  |
:-------------- |:-------------------------------------------------------------------------------------------- |
| **RFC #**     | [453](https://github.com/tensorflow/community/pull/453)  |
| **Author(s)** | Kathy Wu (<kathywu@google.com>), Adam Cogdell (<adamcogdell@google.com>)                     |
| **Sponsor**   | Ivan Petrychenko (<petrychenko@google.com>)                                                  |
| **Updated**   | 2023-07-24                                                                                   |

## Objective

We are proposing a new format for `SavedModel` and a generic proto-splitting library that resolves the 2GB proto issue. The purpose of this RFC is to publicize the design of this new format, which has been implemented (see [tensorflow/tools/proto_splitter](https://github.com/tensorflow/tensorflow/tree/master/tensorflow/tools/proto_splitter)), but is open to comments and changes from the open source community.

### DIFF
--- a/rfcs/20230720-unbound-saved-model.md
+++ b/rfcs/20230720-unbound-saved-model.md
@@ -198,7 +198,7 @@ The outputs of `Splitter.Split()` (`ChunkMetadata` and `ChunkedMessage`) have a 
 
 We decided to explore different serialization formats, even outside of proto. There haven't been any major changes to `SavedModel` in ages - lots of libraries are built on top of the `saved_model.pb` or a `GraphDef` file, and are parsing that proto directly. Any change we make to the format now is almost equally painful, and since we've decided to do this, we explored all options out there.
 
-The conclusion was that re-using proto is the best, as it gave us an easy solution to compatibility so that we can focus on other areas (e.g. MLA integration to interface with other frameworks). However, even splitting the proto comes with compatibility problems: What happens when we need to split the proto again for constants? This breaks compatibility *again*. Also in the future, are we anticipating more cross-framework models (JAX2TF is using a workaround by storing StableHLO as an attribute in a custom op)? Why not invest this time in MLIR, which supports multiple dialects and will never hit a file size limit?
+The conclusion was that re-using proto is the best, as it gave us an easy solution to compatibility so that we can focus on other areas. However, even splitting the proto comes with compatibility problems: What happens when we need to split the proto again for constants? This breaks compatibility *again*. Also in the future, are we anticipating more cross-framework models (JAX2TF is using a workaround by storing StableHLO as an attribute in a custom op)? Why not invest this time in MLIR, which supports multiple dialects and will never hit a file size limit?
 
 #### Proto Splitting: Introduce REF fields to the current schema
 
@@ -255,8 +255,6 @@ The problem is that many existing converters and libraries parse the proto direc
 #### Alternate Formats
 
 We explored written formats for both the image dump and function graphs (which are represented as Proto data structures in-memory). The format for the image does not necessarily need to be Proto-compatible, but the serialization of the function graph should be.
-
-In the future, we should consider using other HLO or using native data structures to represent functions and ops in-memory, rather than a proto-backed data structure. Proto recommends not using protobuf when it is not being serialized.
 
 Specific features we are looking for:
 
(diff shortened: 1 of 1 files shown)


othakkar:
@BlaziusMaximus Thank you for the proposal. Based on my understanding, the proposed implementation solves the issue of **exporting** a `SavedModel` that currently fails due to exceeding the 2GB protobuf serialization limit. 

Does the proposed solution also address the protobuf's limitation that is encountered while **importing** a `GraphDef` or while **freezing** (converting variables to constants that are of size > 2GB) a `SavedModel`? (e.g. https://github.com/tensorflow/tensorflow/issues/60571#issuecomment-1546545861) If not, are there any plans for the same? Thanks!

k-w-w:
@othakkar Thanks for your comment! The GraphDef freezing code is not planned to be updated as part of this project.

However, you may be able to use some of the added infrastructure to address this issue. If I remember correctly, the implementation of `convert_to_constants_v2`, it goes through `Graph.as_graph_def` at some point which fails if the graph is > 2GB, because it first calls into the C API's Graph to GraphDef function, then serializes it to string (then deserializes the string in Python).  This string serialization is typically what hits the 2GB limit.

@BlaziusMaximus added a new argument to `GraphDef._as_graph_def` which directly uses pybind11 proto and skips the string serialization. The default is currently set to `False` to be safe (performance-wise and behavior-wise), but you could try setting the default to `True` to always use pybind11.

```
  def _as_graph_def(
      self, from_version=None, add_shapes=False, use_pybind11_proto=False):
```
https://github.com/tensorflow/tensorflow/blob/863451d7d0875aac843e65f485cdf7dcbbd72c0a/tensorflow/python/framework/ops.py#L2300C7-L2301

othakkar:
@k-w-w thank you for your response!
The reference implementation (in [`as_graph_def(...)`](https://github.com/tensorflow/tensorflow/blob/863451d7d0875aac843e65f485cdf7dcbbd72c0a/tensorflow/python/framework/ops.py#L2300C7-L2301)) that uses `pybind11` is used to convert a `Graph` into a `GraphDef`, whereas the input to the `import_graph_def(...)` function itself is a `GraphDef` proto that is further serialized and is used to be imported into the default graph.

Would you or @BlaziusMaximus be able to elaborate on how to go about updating this code-path to use `pybind11` proto api? My goal is to freeze a `GraphDef` by avoiding the 2GB protobuf limit issue. Please excuse my lack of understanding of the `pybind11` proto api :)

othakkar:
@BlaziusMaximus thanks for the explanation.
I've been exploring how to update the [`import_graph_def()`](https://github.com/tensorflow/tensorflow/blob/v2.13.0/tensorflow/python/framework/importer.py#L353-L411) code-path to use pybind11_protobuf and I could use your help with the following:  Similar to how pybind11_protobuf allows us to pass protos directly from C++ to Python, is there a way to pass a `GraphDef` proto from python to C++ without performing serialization? This would be needed to invoke the [TF_GraphImportGraphDefWithResults](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/c_api.cc#L1801) from pywrap session in C++.

On another note, I'm wondering how do I use the `chunks` and apply some sort of transformation on it? Can you give an end-to-end example where one splits a very large (>2GB) `GraphDef` (in-memory), applies some sort of transformation to the `chunks` and merges the modified `chunks` into a new `GraphDef` proto? I saw this [example](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/proto_splitter/g3doc/in-depth-guide.md#putting-it-all-together) provided by you, but was confused about how to apply any transformation on the `chunks`.

BlaziusMaximus:
@othakkar I'm currently working on an answer to your first question, and have created a separate issue for it (https://github.com/tensorflow/tensorflow/issues/61587).

Unfortunately, I don't have a coded-up example to give for your second question. However, one example of how to use in-memory chunks is if two processes need to send chunked protos over a network, which would require the sender to Split() and the receiver to Merge().

But if you do need to apply a custom transformation, that would require a lot of parsing logic similar to what's done in the Merger, where you'd use a riegeli::Reader to parse through the chunks, using the ChunkMetadata as a guide and making changes along the way. See the [in-depth guide](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/proto_splitter/g3doc/in-depth-guide.md) for more info, or let me know if you have any specific questions. Glad to see that you're interested in the api!

theadactyl (approved):

